PR #1, opened 2017-02-28
Update for buffer API changes

jasongin:


kkoopa:
The data argument of buffers should arguably be `void * `, using `char *` is a legacy oversight on Node's part. Using `char *` makes typecasting required for the user.

Even better would of course be to make them templated on the type of data.

jasongin:
I made the Buffer class templated on the data type. For testing I updated my node-canvas NAPI port to use it, and it did indeed eliminate a lot of casting.

jasongin:
Note there are some `reinterpret_cast<>` calls in the implementation that can go away after or change to `static_cast<>` after https://github.com/nodejs/abi-stable-node/issues/135 is fixed.

napi-inl.h
```diff
@@ -975,41 +979,72 @@ inline Function::CallbackData::CallbackData(FunctionCallback cb, void* data)
...
 }
 
-inline char* Buffer::Data() const {
-  char* data;
-  napi_status status = napi_buffer_data(_env, _value, &data);
-  if (status != napi_ok) throw Error::New(Env());
-  return data;
+template <typename T>
+inline size_t Buffer<T>::Length() const {
+  if (_data == nullptr) {
```
boingoing: This same initialization code is duplicated in Buffer<T>::Length and Buffer<T>::Data. Would it make sense to put it into a helper like a private method Buffer<T>::EnsureCachedInfo or something to make the intent more clear and avoid having to change two places in the future? Also, might want to copy/paste this block again for another method later and inlining a small helper should translate to exactly the same code.

napi-inl.h
```diff
@@ -979,55 +979,70 @@ inline Function::CallbackData::CallbackData(FunctionCallback cb, void* data)
...
+// Buffer<T> class
 ////////////////////////////////////////////////////////////////////////////////
 
-inline Buffer Buffer::New(Napi::Env env, size_t length) {
+template <typename T>
+inline Buffer<T> Buffer<T>::New(Napi::Env env, size_t length) {
   napi_value value;
   char* data;
-  napi_status status = napi_create_buffer(env, length, &data, &value);
+  napi_status status = napi_create_buffer(env, length * sizeof(T), &data, &value);
```
kkoopa: space after keyword, goes for the rest of the file
`sizeof(T) -> sizeof (T)`

napi-inl.h
```diff
@@ -979,55 +979,70 @@ inline Function::CallbackData::CallbackData(FunctionCallback cb, void* data)
...
-inline size_t Buffer::Length() const {
+template <typename T>
+inline size_t Buffer<T>::Length() const {
   if (_data == nullptr) {
+    size_t byteLength;
     napi_status status = napi_get_buffer_info(
-      _env, _value, const_cast<char**>(&_data), const_cast<size_t*>(&_length));
+      _env, _value, reinterpret_cast<char**>(const_cast<T**>(&_data)), &byteLength);
     if (status != napi_ok) throw Error::New(Env());
+    *const_cast<size_t*>(&_length) = byteLength / sizeof(T);
```
kkoopa: Mark the field `_length` `mutable` instead.

napi-inl.h
```diff
@@ -979,55 +979,70 @@ inline Function::CallbackData::CallbackData(FunctionCallback cb, void* data)
...
-inline char* Buffer::Data() const {
+template <typename T>
+inline T* Buffer<T>::Data() const {
   if (_data == nullptr) {
+    size_t byteLength;
     napi_status status = napi_get_buffer_info(
-      _env, _value, const_cast<char**>(&_data), const_cast<size_t*>(&_length));
+      _env, _value, reinterpret_cast<char**>(const_cast<T**>(&_data)), &byteLength);
     if (status != napi_ok) throw Error::New(Env());
+    *const_cast<size_t*>(&_length) = byteLength / sizeof(T);
```
kkoopa: Mark the field `_length` `mutable` instead.

napi-inl.h
jasongin: Yes, I refactored this.

napi-inl.h
jasongin: Fixed.

napi-inl.h
jasongin: Fixed.

napi-inl.h
jasongin: Fixed.

PR #2, opened 2017-02-28
Enable include files via package dependency

jasongin:
This enables other packages to get the N-API headers via a package dependency, the same way that NAN works. The header files would be referenced like this in `binding.gyp`: 
```json
  'include_dirs': ["<!(node -p \"require('node-api').include\")"],
```

The same mechanism can be used in the future to enable referencing other things from the package like a static lib or maybe a gyp dependency, when we add the source code for back-compat.

Currently I've set a package name of **node-api**, with the expectation that we'll be able to [acquire that package name](https://github.com/laurie71/node-api/issues/2).

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,7 @@
+const include = __dirname;
+const isNodeApiBuiltin = process.versions.modules >= 52;
+
+module.exports = {
+   include: include,
+   isNodeApiBuiltin: isNodeApiBuiltin,
+};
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -0,0 +1,7 @@
+const include = __dirname;
+const isNodeApiBuiltin = process.versions.modules >= 52;
```
boingoing: Is there a story for versioning N-API or do we always just provide whichever version is provided by the napi header/cc?

jasongin: For versions of node without N-API built-in, the N-API version will be the same as the node-api package version. For versions of node with N-API built-in, the N-API version will be the same as the node version. It's tricky because the former is determined at build time, while the latter is determined at run-time. This is an area we know we need to further define and document the behavior and best practices for native module developers.

PR #3, opened 2017-02-28
AsyncWorker, ObjectReference, FunctionReference classes

jasongin:
- Add an `AsyncWorker` class that is an equivalent of the one defined in `node_api_helpers.h` and `nan.h`, but done in the style of the other NAPI C++ wrappers.
 - Create new `ObjectReference` and `FunctionReference` classes that make it easy to get/set properties and call functions via a reference with automatic handle scoping. These are both used by `AsyncWorker`, but are also likely to be useful in other scenarios.
 - Change many method parameters to accept NAPI primitives instead of NAPI classes. Since the classes are implicitly convertible to the primitives it doesn't affect passing class instances, but it makes it easier to interop with the NAPI C APIs that only use the primitives.
 - Move the index getters and setters from `Array` to `Object`, because objects support indexed properties. This more closely matches the V8 object model anyway.

After this, we should be able to delete `node_api_helpers.h` and update any uses of the helper classes there to use the classes from here instead.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1151,7 +1151,7 @@ inline Reference<T> Reference<T>::New(const T& value, int initialRefcount) {
   if (val == nullptr) {
     return Reference<T>(env, nullptr);
   }
-  
+
   napi_ref ref;
   napi_status status = napi_create_reference(env, value, initialRefcount, &ref);
   if (status != napi_ok) throw Error::New(Napi::Env(env));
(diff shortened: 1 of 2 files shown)


napi-inl.h
```diff
@@ -1779,6 +1779,103 @@ inline Value EscapableHandleScope::Escape(Value escapee) {
...
+    HandleOKCallback();
+  }
+  else {
+    HandleErrorCallback();
+  }
+}
+
+inline Object AsyncWorker::Persistent() {
+  return _persistent.Value();
+}
```
kkoopa: I find the name slightly confusing. I would expect a method called `Persistent` to return a `Persistent<Object> &`.

napi-inl.h
```diff
@@ -1779,6 +1779,103 @@ inline Value EscapableHandleScope::Escape(Value escapee) {
...
+
+inline void AsyncWorker::HandleOKCallback() {
```
kkoopa: Would it not be better to drop the Handle-prefix on these method names? The `Callback` suffix also feels slightly redundant. How about something like `OnOk`, `OnError`, `OnExecute`, which would fit the other names?

jasongin: This is intended to be similar to the NAN class with the same name [and same `Handle`-prefixed methods](https://github.com/nodejs/nan/blob/master/nan.h#L1575), to make porting from NAN to NAPI straightforward. I do prefer the `On` prefix myself. So I'd be OK with that if we think continuity from NAN is less important.

napi-inl.h
jasongin: This is a replacement for the `SaveToPersistent` and `GetFromPersistent` overloaded methods on `Nan::AsyncWorker`. However I've been thinking this might not be the best design. The problem with the single method for retrieving the `Object` from the persistent reference is that it requires the caller to have a `HandleScope` to ensure the `Object` handle gets cleaned up efficiently. The other methods were verbose but had the advantage of managing the `HandleScope` automatically.

kkoopa: Just call them `Set` and `Get`. Think of `AsyncWorker` as an extension of `Object`, although it is a composition.

napi-inl.h
jasongin: Yes, I'll update it.

sampsongao:
If we don't want to take care of the node-0.10 support in our PR, then the napi_async_XXX api could be reverted back to direct uv calls.

napi-inl.h
```diff
@@ -1779,6 +1779,103 @@ inline Value EscapableHandleScope::Escape(Value escapee) {
...
+inline AsyncWorker::operator napi_work() const {
+  return _work;
+}
+
+inline Env AsyncWorker::Env() const {
+  return Napi::Env(_env);
+}
+
+inline void AsyncWorker::Queue() {
+  napi_async_set_data(_work, static_cast<void*>(this));
```
kkoopa: The typecast should be unnecessary, since `void *` is the top type of pointers.

mhdawson: I still think that encapsulation of the uv calls here can make sense. Otherwise if uv changes then you won't get the abi stability.  I know we won't wrap all of uv, but since this very commonly used I think it makes sense here.

napi-inl.h
mhdawson: I think I like having the Set and Get as opposed to just returning the persistentl. I'm assuming they were in NaN because they were used frequently enough that having the code once on AsyncWorker made sense.

jasongin:
Added a big commit with a bunch of refactoring and improvements. I tested this version by updating the NAPI leveldown port to use this instead of `node_api_helpers.h`. @sampsongao you'll probably want to use this (and the [buffer update](https://github.com/nodejs/abi-stable-node-wrapper/pull/1)) for updating nanomsg.

jasongin:
@digitalinfinity For some reason I can't add you as a reviewer in this repo. But you might take a look if you have time.

boingoing:
Looks good to me. 👍

jasongin:
OK, I'm not switching from napi to libuv async calls right now. If we decide to remove those APIs from NAPI then I'll update this code accordingly.

PR #4, opened 2017-03-16
Enable initializer_list for function args

jasongin:
The C++ APIs for calling functions and defining properties required use of `std::vector`. While that's most flexible, a vector always allocates on the heap, which is costly and usually should not be necessary. This change adds alternative overloads everywhere `std::vector` is used, to allow `std::initializer_list` instead (that is very cheap, and sits on the stack). That makes the most common use of the APIs, where the arguments are specified using initializer-list syntax inline in the function call, very efficient. This change did yield a measurable performance improvement: approximately 5% improvement in the most extreme canvas benchmarks.

I also experimented with variadic template function arguments as an alternative to `std::initializer_list`. They might have offered a more natural syntax for function calls, with similar efficiency. But it just got too complex (especially restricting the types of the variable arguments), and could not disambiguate between the *this* argument and normal arguments, so it wasn't going to be worth it.

Also in this change, I'm extending the compatibility with the NAPI primitives for function args and some other places, particularly `::New()` methods. Because the NAPI C++ classes are implicitly convertible to the primitives, and the reverse conversion is not implicit, most of the methods need to accept primitive arguments (`napi_env`, `napi_value`, etc.) but return classes(`Napi::Value`, etc.) in the return value. That makes it possible to use most of the APIs using either the primitives or the classes without any explicit conversion. (Note nearly all conversions get optimized out at compile time.)

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -74,7 +74,7 @@ inline PropertyName::PropertyName(napi_env env, napi_propertyname name) : _env(e
 
 inline PropertyName::PropertyName(napi_env env, const char* utf8name) : _env(env) {
   napi_status status = napi_property_name(env, utf8name, &_name);
-  if (status != napi_ok) throw Error::New(Env());
+  if (status != napi_ok) throw Error::New(_env);
 }
 
 inline PropertyName::PropertyName(napi_env env, const std::string& utf8name)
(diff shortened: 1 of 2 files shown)


jasongin:
@digitalinfinity

napi-inl.h
```diff
@@ -1267,7 +1257,7 @@ inline Reference<T>::operator napi_ref() const {
 
 template <typename T>
 inline bool Reference<T>::operator ==(const Reference<T> &other) const {
-  HandleScope scope = HandleScope(Env());
+  HandleScope scope = Handlescope(_env);
```
boingoing: Is this a typo? HandleScope vs Handlescope?

jasongin: Huh, good catch. That shouldn't compile... but that's a funny thing about inline functions, they don't get compiled if nothing uses them.

jasongin: Fixed with another commit.

PR #5, opened 2017-03-17
Updates for N-API changes

jasongin:
- Remove the `PropertyName` class (wrapper around `napi_propertyname` type) and all uses of it.
 - Add overloads for `Has()`, `Get()`, `Set()` that take `napi_value` parameters as an alternative to strings.
 - Call the new `named_property` APIs where applicable.
 - Add overloads for `Set()` that take `Napi::Value` parameters as an alternative to `napi_value`. I don't entirely understand why this is necessary, but it eliminates some compiler errors about ambiguous overload resolution.
 - Make the `External` class templated on the type of external data that it holds.
 - Support finalize hints

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -11,25 +11,37 @@
 
 namespace Napi {
 
-// Adapt the NAPI_MODULE registration function:
-//  - Wrap the arguments in NAPI wrappers.
-//  - Catch any NAPI errors that might be thrown.
-#define NODE_API_MODULE(modname, regfunc)                                \
-  void __napi_ ## regfunc(napi_env env,                                  \
-                          napi_value exports,                            \
-                          napi_value module,                             \
-                          void* priv) {                                  \
-    try {                                                                \
-      regfunc(Napi::Env(env),                                            \
-              Napi::Object(env, exports),                                \
-              Napi::Object(env, module));                                \
-    }                                                                    \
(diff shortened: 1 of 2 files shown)


napi-inl.h
```diff
@@ -11,19 +11,24 @@
...
+                          napi_value exports,                            \
+                          napi_value module,                             \
+                          void* priv) {                                  \
+    try {                                                                \
+      regfunc(Napi::Env(env),                                            \
+              Napi::Object(env, exports),                                \
+              Napi::Object(env, module));                                \
+    }                                                                    \
+    catch (const Napi::Error&) {                                         \
+      assert(false); /* Uncaught error in native module registration. */ \
```
digitalinfinity: Should this print out a warning in retail code? or just abort perhaps?

digitalinfinity:
:shipit:

napi-inl.h
jasongin: Good question. Now that I think about it, the correct thing to do here would be to throw a JavaScript exception. (If there is not already a pending exception.) [That is exactly what the callback wrapper does](https://github.com/nodejs/node-api/blob/master/napi-inl.h#L976-L981), and it should work just as well here. I think I coded this `assert()` before I designed the error-to-exception mechanism.

I'll update this PR.

PR #7, opened 2017-03-18
Enable back-compat with pre-N-API node

jasongin:
Adding logic to `index.js` and a `node_api.gyp` to enable conditional static-linking to N-API. See the updated `README.md` for more info. And we still need to figure out a good way to detect at runtime whether the current node engine has N-API built-in or not.

The `node_api*` files here are copied directly from https://github.com/nodejs/abi-stable-node/pull/167; see that PR for more details. Technically there is no reason those files need to be identical, but it makes maintenance much easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,34 @@
 # Node.js API (N-API) Package
 
 This package contains header-only C++ wrapper classes for the ABI-stable
-Node.js API (N-API).
+Node.js API (N-API), along with library code that enables
+backward-compatibility with use with older versions of Node.js that do
+not have N-API built-in.
 
-To use these headers in a native module:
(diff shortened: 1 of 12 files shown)


jasongin:
@digitalinfinity and I discussed offline the potential use of a .gypi file to simplify the build configuration for native modules that depend on the node-api package. I will submit that later as a separate PR.

jasongin:
Oh, that's not going to work as nicely as I hoped. Apparently it's not possible to use a command expansion to get the path to an included file. ([GYP reference](https://gyp.gsrc.io/docs/InputFormatReference.md#Including-Other-Files)) That's a problem because the path to the .gypi is not fixed.

PR #10, opened 2017-04-03
Updates for N-API async changes

jasongin:
Updates for pending changes in async APIs at https://github.com/nodejs/abi-stable-node/pull/204, along with other recent changes.
 - Rename `napi_work` to `napi_async_work`
 - Async APIs all take an `env` parameter and return a `napi_status`
 - New `napi_cancel_async_work()` API
 - Track (a persistent reference to) an `Error` object for async work errors, rather than just an error message.
 - Error handling APIs all return a `napi_status`
 - Reference counts are unsigned

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1229,7 +1229,7 @@ inline RangeError::RangeError(napi_env env, napi_value value) : Error(env, value
 ////////////////////////////////////////////////////////////////////////////////
 
 template <typename T>
-inline Reference<T> Reference<T>::New(const T& value, int initialRefcount) {
+inline Reference<T> Reference<T>::New(const T& value, uint32_t initialRefcount) {
   napi_env env = value.Env();
   napi_value val = value;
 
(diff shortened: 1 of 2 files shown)


napi-inl.h
```diff
@@ -1316,15 +1322,15 @@ inline T Reference<T>::Value() const {
 
 template <typename T>
 inline int Reference<T>::Ref() {
```
kkoopa: Either the type is wrong or `result` needs a typecast before returning.

napi-inl.h
```diff
@@ -1316,15 +1322,15 @@ inline T Reference<T>::Value() const {
...
-  int result;
+  uint32_t result;
   napi_status status = napi_reference_ref(_env, _ref, &result);
   if (status != napi_ok) throw Error::New(_env);
   return result;
 }
 
 template <typename T>
 inline int Reference<T>::Unref() {
```
kkoopa: Either the type is wrong or `result` needs a typecast before returning.

napi-inl.h
jasongin: Fixed here and a few other places.

jasongin:
@boingoing What is the plan for getting the async APIs into node?

Anyway, I don't see any point in holding this PR until those changes land, because the current code in this repo references the old N-API async functions that never landed in node. So updating this repo won't make things any worse.

boingoing:
>@boingoing What is the plan for getting the async APIs into node?

Prepping the PR for these changes as well as the one which changes ```napi_callback``` to return a ```napi_value```. Both should be live today.

Agreed, no reason to hold on the changes in this PR here.

PR #19, opened 2017-04-11
Make Function::New more usable, support obj[…] for writing and add a small test

addaleax:
Hi all! :wave: I’ve decided to do a thing:

* 52c1aa9 makes it easier to create new functions (avoid compiler warnings about ambiguity when creating functions)
* 051eb13 adds support for `obj[key] = value` notation
* 4f2fe82 adds a small test for the 2 features mentioned above

Let me know what you think!

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -898,47 +898,66 @@ inline const T* TypedArray_<T,A>::Data() const {
 // Function class
 ////////////////////////////////////////////////////////////////////////////////
 
+// Helpers to handle functions exposed from C++.
+namespace details {
+
+template <typename Callable, typename Return>
+struct CallbackData {
+  static inline
+  napi_value Wrapper(napi_env env, napi_callback_info info) {
+    try {
+      CallbackInfo callbackInfo(env, info);
+      CallbackData* callbackData =
+        static_cast<CallbackData*>(callbackInfo.Data());
+      return callbackData->callback(callbackInfo);
+    }
+    NAPI_RETHROW_JS_ERROR(env)
+  }
+
+  Callable callback;
+};
+
+template <typename Callable>
+struct CallbackData<Callable, void> {
+  static inline
+  napi_value Wrapper(napi_env env, napi_callback_info info) {
+    try {
+      CallbackInfo callbackInfo(env, info);
+      CallbackData* callbackData =
+        static_cast<CallbackData*>(callbackInfo.Data());
+      callbackData->callback(callbackInfo);
+      return nullptr;
+    }
+    NAPI_RETHROW_JS_ERROR(env)
+  }
+
+  Callable callback;
+};
+
+}  // namespace details
+
+template <typename Callable>
 inline Function Function::New(napi_env env,
-                              VoidFunctionCallback cb,
-                              const char* utf8name,
-                              void* data) {
+                              Callable cb,
+                              const char* utf8name) {
+  typedef decltype(cb(CallbackInfo(nullptr, nullptr))) ReturnType;
+  typedef details::CallbackData<Callable, ReturnType> CbData;
   // TODO: Delete when the function is destroyed
-  VoidFunctionCallbackData* callbackData = new VoidFunctionCallbackData({ cb, data });
+  auto callbackData = new CbData({ cb });
 
   // TODO: set the function name
   napi_value value;
   napi_status status = napi_create_function(
-    env, utf8name, VoidFunctionCallbackWrapper, callbackData, &value);
+    env, utf8name, CbData::Wrapper, callbackData, &value);
   if (status != napi_ok) throw Error::New(env);
   return Function(env, value);
 }
 
+template <typename Callable>
 inline Function Function::New(napi_env env,
-                              FunctionCallback cb,
-                              const char* utf8name,
-                              void* data) {
-  // TODO: Delete when the function is destroyed
-  FunctionCallbackData* callbackData = new FunctionCallbackData({cb, data});
-
-  napi_value value;
-  napi_status status = napi_create_function(
-    env, utf8name, FunctionCallbackWrapper, callbackData, &value);
-  if (status != napi_ok) throw Error::New(env);
-  return Function(env, value);
-}
-
-inline Function Function::New(napi_env env,
-                              VoidFunctionCallback cb,
-                              const std::string& utf8name,
-                              void* data) {
-  return New(env, cb, utf8name.c_str(), data);
-}
-
-inline Function Function::New(napi_env env,
-                              FunctionCallback cb,
-                              const std::string& utf8name,
-                              void* data) {
-  return New(env, cb, utf8name.c_str(), data);
+                              Callable cb,
+                              const std::string& utf8name) {
+  return New(env, cb, utf8name.c_str());
 }
 
 inline Function::Function() : Object() {
@@ -1015,27 +1034,6 @@ inline Object Function::New(const std::vector<napi_value>& args) const {
     _env, _value, args.size(), args.data(), &result);
   if (status != napi_ok) throw Error::New(_env);
   return Object(_env, result);
-}
-
-inline napi_value Function::VoidFunctionCallbackWrapper(napi_env env, napi_callback_info info) {
-  try {
-    CallbackInfo callbackInfo(env, info);
-    VoidFunctionCallbackData* callbackData =
-      reinterpret_cast<VoidFunctionCallbackData*>(callbackInfo.Data());
-    callbackData->callback(callbackInfo);
-    return nullptr;
-  }
-  NAPI_RETHROW_JS_ERROR(env)
-}
-
-inline napi_value Function::FunctionCallbackWrapper(napi_env env, napi_callback_info info) {
-  try {
-    CallbackInfo callbackInfo(env, info);
-    FunctionCallbackData* callbackData =
-      reinterpret_cast<FunctionCallbackData*>(callbackInfo.Data());
-    return callbackData->callback(callbackInfo);
-  }
-  NAPI_RETHROW_JS_ERROR(env)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -425,6 +425,21 @@ inline std::u16string String::Utf16Value() const {
 // Object class
 ////////////////////////////////////////////////////////////////////////////////
 
+template <typename Key>
+inline PropertyLValue<Key>::operator Value() const {
+  return _object.Get(_key);
+}
+
+template <typename Key> template <typename ValueType>
+inline PropertyLValue<Key>& PropertyLValue<Key>::operator =(ValueType value) {
+  _object.Set(_key, value);
+  return *this;
+}
+
+template <typename Key>
+inline PropertyLValue<Key>::PropertyLValue(Object object, Key key)
+  : _object(object), _key(key) {}
+
 inline Object Object::New(napi_env env) {
   napi_value value;
   napi_status status = napi_create_object(env, &value);
@@ -436,6 +451,18 @@ inline Object::Object() : Value() {
 }
 
 inline Object::Object(napi_env env, napi_value value) : Value(env, value) {
+}
+
+inline PropertyLValue<std::string> Object::operator [](const char* name) {
+  return PropertyLValue<std::string>(*this, name);
+}
+
+inline PropertyLValue<std::string> Object::operator [](const std::string& name) {
+  return PropertyLValue<std::string>(*this, name);
+}
+
+inline PropertyLValue<uint32_t> Object::operator [](uint32_t index) {
+  return PropertyLValue<uint32_t>(*this, index);
 }
 
 inline Value Object::operator [](const char* name) const {
--- a/napi.h
+++ b/napi.h
@@ -49,12 +49,6 @@ namespace Napi {
   typedef TypedArray_<float, napi_float32_array> Float32Array;
   typedef TypedArray_<double, napi_float64_array> Float64Array;
 
-  // Functions exposed to JavaScript must conform to one these callback signatures.
-  // These are std::function<> typedefs instead of function pointers to enable lambdas.
-  // (See ObjectWrap<T> for callbacks used when wrapping entire classes.)
-  typedef std::function<void(const CallbackInfo& info)> VoidFunctionCallback;
-  typedef std::function<Value(const CallbackInfo& info)> FunctionCallback;
-
   // A N-API C++ module's registration callback (init) function has this sinature.
   typedef void ModuleRegisterCallback(Env env, Object exports, Object module);
 
@@ -323,22 +317,16 @@ namespace Napi {
 
   class Function : public Object {
   public:
+    // Callable must implement operator() accepting a const CallbackInfo&
+    // and return either void or Value.
+    template <typename Callable>
     static Function New(napi_env env,
-                        VoidFunctionCallback cb,
-                        const char* utf8name = nullptr,
-                        void* data = nullptr);
+                        Callable cb,
+                        const char* utf8name = nullptr);
+    template <typename Callable>
     static Function New(napi_env env,
-                        FunctionCallback cb,
-                        const char* utf8name = nullptr,
-                        void* data = nullptr);
-    static Function New(napi_env env,
-                        VoidFunctionCallback cb,
-                        const std::string& utf8name,
-                        void* data = nullptr);
-    static Function New(napi_env env,
-                        FunctionCallback cb,
-                        const std::string& utf8name,
-                        void* data = nullptr);
+                        Callable cb,
+                        const std::string& utf8name);
 
     Function();
     Function(napi_env env, napi_value value);
@@ -357,18 +345,6 @@ namespace Napi {
 
     Object New(const std::initializer_list<napi_value>& args) const;
     Object New(const std::vector<napi_value>& args) const;
-
-  private:
-    static napi_value VoidFunctionCallbackWrapper(napi_env env, napi_callback_info info);
-    static napi_value FunctionCallbackWrapper(napi_env env, napi_callback_info info);
-
-    template <typename TCallback>
-    struct CallbackData {
-      TCallback callback;
-      void* data;
-    };
-    typedef CallbackData<VoidFunctionCallback> VoidFunctionCallbackData;
-    typedef CallbackData<FunctionCallback> FunctionCallbackData;
   };
 
   template <typename T>
--- a/napi.h
+++ b/napi.h
@@ -36,6 +36,7 @@ namespace Napi {
   class PropertyDescriptor;
   class CallbackInfo;
   template <typename T> class Reference;
+  template <typename Key> class PropertyLValue;
 
   class TypedArray;
   template <typename T, napi_typedarray_type A> class TypedArray_;
@@ -175,6 +176,10 @@ namespace Napi {
     Object();
     Object(napi_env env, napi_value value);
 
+    PropertyLValue<std::string> operator [](const char* name);
+    PropertyLValue<std::string> operator [](const std::string& name);
+    PropertyLValue<uint32_t> operator [](uint32_t index);
+
     Value operator [](const char* name) const;
     Value operator [](const std::string& name) const;
     Value operator [](uint32_t index) const;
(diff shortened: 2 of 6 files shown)


addaleax:
Also, is there a good existing way to test the code here?

jasongin:
> is there a good existing way to test the code here?

What I've done is build my and test [my fork of node-canavas](https://github.com/jasongin/node-canvas/tree/napi) that uses this library. It doesn't cover absolutely everything but it does cover a lot. However that's not super easy since canvas has some prerequisites, and then you need to manually update the node-api dependency to pick up the changes.

That was never meant to be the only way to test this code in the long term though. We definitely should add tests for this library that provide more complete coverage and are easier to run. So far, most of the effort for the N-API project has been on getting the lower level APIs integrated into node. I think we will start to put more effort into this package in the coming months, including tests and documentation.

addaleax:
> I think we will start to put more effort into this package in the coming months, including tests and documentation.

That’s cool to hear – I’m glad to help wherever I can. :) Are there any fixed rules on contributing to this repo? Also, should I maybe join the WG? :smile:

jasongin:
Nice, at first glance these look like some great improvements! I don't have time right now to do a thorough review, I'll try to get to that later tonight.

jasongin:
For this repo, the only requirement right now is one or more reviews from people familiar with the code, I guess that would be: @mhdawson, @boingoing , @kkoopa, and myself. We obviously haven't yet set up anything like lint or tests or CI. As mentioned above, I do try to make sure at least the node-canvas port works, but with breaking changes it's not expected to be always in sync.

If you're interested, @aruneshchandra  organizes weekly hangout meetings, currently Thursdays at 10:30am Pacific Time where we track progress, make plans and schedules, and discuss any significant issues.

addaleax:
> If you're interested, @aruneshchandra organizes weekly hangout meetings, currently Thursdays at 10:30am Pacific Time where we track progress, make plans and schedules, and discuss any significant issues.

Yeah, thanks for letting me know. :) This week won’t work for me but I’ll try to join you sometime.

aruneshchandra:
@addaleax feel free to jump in our weekly sync up whenever you have time. Here's the hangout link
 https://plus.google.com/u/0/events/c0eevtrlajniu7h8cjrdk0f56c8?authkey=COH04YCalJS8Ug

Scheduled sync up time

weekly-standup-time

//cc: @mhdawson for adding @addaleax to @nodejs/n-api

mhdawson:
@aruneshchandra, added @addaleax

mhdawson:
I do want to get some CI testing for the modules we've ported, some of which would test this as well, just have not had time to get to that yet.

napi-inl.h
```diff
@@ -392,6 +425,38 @@ inline std::u16string String::Utf16Value() const {
...
+  friend class Napi::Object;
+};
+
+template <typename Key>
+inline PropertyLValue<Key>::operator Value() const {
+  return _object.Get(_key);
+}
+
+template <typename Key>
+inline PropertyLValue<Key>& PropertyLValue<Key>::operator =(Value value) {
```
kkoopa: Either this definition or the declaration above has a space wrong.

napi-inl.h
```diff
@@ -865,47 +942,65 @@ inline const T* TypedArray_<T,A>::Data() const {
...
+  Callable callback;
+};
+
+}  // namespace helper
+
+template<typename Callable>
 inline Function Function::New(napi_env env,
-                              FunctionCallback cb,
-                              const char* utf8name,
-                              void* data) {
```
kkoopa: What became of the data argument?

napi-inl.h
```diff
@@ -1791,9 +1834,8 @@ inline ClassPropertyDescriptor<T> ObjectWrap<T>::InstanceMethod(
     InstanceMethodCallback method,
     napi_property_attributes attributes,
     void* data) {
-  CallbackData* callbackData = new CallbackData({}); // TODO: Delete when the class is destroyed
-  callbackData->instanceMethodCallback = method;
-  callbackData->data = data;
+  // TODO: Delete when the class is destroyed
+  InstanceMethodCallbackData* callbackData = new InstanceMethodCallbackData({ method, data });
```
kkoopa: Why not `auto`? Applies to other locations as well.

napi-inl.h
```diff
@@ -1866,190 +1907,140 @@ inline void ObjectWrap<T>::ConstructorCallbackWrapper(
...
 
 template <typename T>
-inline void ObjectWrap<T>::StaticVoidMethodCallbackWrapper(
+inline napi_value ObjectWrap<T>::StaticVoidMethodCallbackWrapper(
     napi_env env,
     napi_callback_info info) {
   try {
     CallbackInfo callbackInfo(env, info);
-    CallbackData* callbackData = reinterpret_cast<CallbackData*>(callbackInfo.Data());
+    StaticVoidMethodCallbackData* callbackData =
```
kkoopa: `auto`

napi-inl.h
```diff
@@ -336,14 +347,36 @@ inline double Number::DoubleValue() const {
...
-inline String String::New(napi_env env, const char* val, int length) {
+inline String String::New(napi_env env, const std::string& val) {
+  return String::New(env, val.c_str(), val.size());
+}
+
+inline String String::New(napi_env env, const std::u16string& val) {
+  return String::New(env, val.c_str(), val.size());
+}
+
+inline String String::New(napi_env env, const char* val) {
```
jasongin: The `napi_value` created here is not temporary. It gets wrapped in the `String` object that is returned from this function.

Also, this code actually is part of a commit that is from my PR https://github.com/nodejs/node-api/pull/18, so it might be better to review that over there, and look at the other commits building on it separately.

napi-inl.h
jasongin: _Why_ `auto`? I don't usually prefer it, personally. Is there a Node.js style guide that suggests when to use it?

napi-inl.h
```diff
@@ -425,6 +425,38 @@ inline std::u16string String::Utf16Value() const {
 // Object class
 ////////////////////////////////////////////////////////////////////////////////
 
+namespace helper {
```
jasongin: I've usually seen namespaces like this named `details`.

napi.h
```diff
@@ -168,13 +168,21 @@ namespace Napi {
     std::u16string Utf16Value() const;
   };
 
+  namespace helper {
+    template <typename T> class PropertyLValue;
```
jasongin: `typename Key`

napi-inl.h
```diff
@@ -425,6 +425,38 @@ inline std::u16string String::Utf16Value() const {
...
+
+template <typename Key>
+class PropertyLValue {
+public:
+  operator Value() const;
+  PropertyLValue& operator=(Value value);
```
jasongin: I think this should have overloads for other types that can be assigned to a value, like `Object::Set()`.

napi-inl.h
jasongin: Though I'm not sure this class really needs to be "hidden" in a namespace like this, since it is sort of part of the public API.

napi-inl.h
```diff
@@ -425,6 +425,38 @@ inline std::u16string String::Utf16Value() const {
 // Object class
 ////////////////////////////////////////////////////////////////////////////////
 
+namespace helper {
+
+template <typename Key>
+class PropertyLValue {
```
jasongin: I might just name this `PropertyValue` and drop the `L`. But I don't feel strongly about it.

napi-inl.h
```diff
@@ -898,47 +898,65 @@ inline const T* TypedArray_<T,A>::Data() const {
...
 
+template <typename Callable>
+struct CallbackData<Callable, void> {
+  static inline
+  napi_value Wrapper(napi_env env, napi_callback_info info) {
+    try {
+      CallbackInfo callbackInfo(env, info);
+      CallbackData* callbackData =
+        static_cast<CallbackData*>(callbackInfo.Data());
+      callbackData->callback(callbackInfo);
```
jasongin: I think this has to return `nullptr` here.

napi-inl.h
```diff
@@ -898,47 +898,65 @@ inline const T* TypedArray_<T,A>::Data() const {
...
 ////////////////////////////////////////////////////////////////////////////////
 
-inline Function Function::New(napi_env env,
-                              VoidFunctionCallback cb,
-                              const char* utf8name,
-                              void* data) {
-  // TODO: Delete when the function is destroyed
-  VoidFunctionCallbackData* callbackData = new VoidFunctionCallbackData({ cb, data });
+// Helpers to handle functions exposed from C++.
+namespace helper {
```
jasongin: Is a nested namespace required? Or could the `CallbackData` structs remain private inside the `Function` class?

kkoopa:
Do not know of a node.js style guide on auto, but in general auto should be used everywhere. Ask Herb Sutter. I mentioned it because I saw auto in some other places in this PR, maybe from other commits.

napi-inl.h
addaleax: @kkoopa Sorry, which space are you talking about?

napi-inl.h
addaleax: > What became of the data argument?

It was unused. I don’t mind it being there, but that should maybe be reserved for a version that actually expects callbacks with a `data` argument.

napi-inl.h
addaleax: This is also from #18, so let’s discuss it there. Personally, I like to use it when it de-clutters the code at points where types are obvious anyway (e.g. I think it’s perfectly okay for the return value of `new Foobar()`), but prefer to avoid it when using `auto` would mean I have to look up the type.

I don’t think we have a fixed style guide; iirc even Nan and Node core don’t follow the same rules.

napi-inl.h
addaleax: I moved this one to the `Napi` namespace and into the public header. (And also renamed the namespace to `details`.)

napi.h
addaleax: Dropped this part anyway because of moving the class into the public header

napi-inl.h
addaleax: Makes sense, `template`’d :smile:

napi-inl.h
addaleax: I’d prefer to keep a name that reflects that this is something that’s meant to be *assignable*. Calling it something along the lines of `PropertyReference` also crossed my mind (e.g. comparable to `std::bitset<N>::reference`), that works for me too.

napi-inl.h
addaleax: done!

napi-inl.h
addaleax: Yeah, the explicit template specialization doesn’t work outside of namespace scope 😩

napi-inl.h
kkoopa: The space between keyword `operator` and `=`. In the declaration it is `operator=`, here it says `operator =`

napi-inl.h
kkoopa: It is for the arbitrary data you can attach to a callback and get from `CallbackInfo::Data`.

napi-inl.h
addaleax: @kkoopa Ah, thanks, updated (with `operator =`, which seems to be the prevalent style for declarations/definitions here).

napi-inl.h
kkoopa: Ah, it has to use an `Object`, cf. NAN. Something like this (in horrible pseudocode)
```c++
Function::New(..., Value v) {
CallbackInfo(Object{v});
...
}

RealCallback(CallbackInfo info) {
 auto data = info.Data().real_data;
  info.Data().call(real_data);
}
```c++

Comment (addaleax) [inline napi-inl.h]: >  `auto data = info.Data().real_data;`
>  `info.Data().call(real_data);`

Sorry, what is `real_data` in that context?

And also: I have no problem re-adding a generic data value here, whether that’s a `Value` or a pointer if you would prefer that. I just decided to leave it away for now because it didn’t work until now, and allowing `std::function`s more or less circumvents the problem solved by that extra opaque anyway.

Comment (kkoopa) [inline napi-inl.h]: `real_data` is the data argument supplied to function::New, stored on an `Object` (as hidden value or such for V8), which is passed as the `data` argument to `v8::Function::New` or equivalent. Then this gets unwrapped before calling the callback the user sees.
However, if this was not working previously, I guess it is better to leave this as-is and address `data` in a separate PR. Maybe there was a plan for it.

For future reference, perhaps the actual code I was thinking of makes this comprehensible
```c++
Factory<v8::Function>::return_t
Factory<v8::Function>::New( FunctionCallback callback
                          , v8::Local<v8::Value> data) {
  v8::Isolate *isolate = v8::Isolate::GetCurrent();
  v8::EscapableHandleScope scope(isolate);
  v8::Local<v8::ObjectTemplate> tpl = v8::ObjectTemplate::New(isolate);
  tpl->SetInternalFieldCount(imp::kFunctionFieldCount);
  v8::Local<v8::Object> obj = NewInstance(tpl).ToLocalChecked();

  obj->SetInternalField(
      imp::kFunctionIndex
    , v8::External::New(isolate, reinterpret_cast<void *>(callback)));

  v8::Local<v8::Value> val = v8::Local<v8::Value>::New(isolate, data);

  if (!val.IsEmpty()) {
    obj->SetInternalField(imp::kDataIndex, val);
  }

  return scope.Escape(v8::Function::New( isolate
                          , imp::FunctionCallbackWrapper
                          , obj));
}
...
static
void FunctionCallbackWrapper(const v8::FunctionCallbackInfo<v8::Value> &info) {
  v8::Local<v8::Object> obj = info.Data().As<v8::Object>();
  FunctionCallback callback = reinterpret_cast<FunctionCallback>(
      reinterpret_cast<intptr_t>(
          obj->GetInternalField(kFunctionIndex).As<v8::External>()->Value()));
  FunctionCallbackInfo<v8::Value>
      cbinfo(info, obj->GetInternalField(kDataIndex));
  callback(cbinfo);
}
```

jasongin: > However, if this was not working previously, I guess it is better to leave this as-is and address data in a separate PR.

It does look like there was a bug in handling the user data pointer previously. The intention was for the `void* data` field of `CallbackData` to be made available to user code via `CallbackInfo::Data()`, but that didn't happen.

addaleax: Okay, I think we’re all very much on the same page. Again, I’d be happy to re-add a `data` pointer/field, but maybe not for this PR?

napi-inl.h
jasongin: I like `PropertyReference` too.

jasongin: Well, maybe not, that might be too confusing when there are `ObjectReference` and `FunctionReference` classes that use a different meaning of "reference".

napi-inl.h
jasongin: Sure, I'm OK with fixing that separately.

napi-inl.h
addaleax: ’xactly. :smile: Happy to do whatever you prefer, but as long as it’s not intended for public usage it won’t matter much anyway.

jasongin:
> in general auto should be used everywhere. Ask Herb Sutter.

Please, no. In my opinion, Herb Sutter's reasoning severely underweights the importance of readability / reviewability. As someone who reviews a lot of code (usually not in an IDE that can help with type info)
I find over-use of auto costs extra time and mental effort for the reviewer. Maybe small amounts in each instance, but it can add up.

I'm _okay_ with it being used in places like where you suggested here, when the type is obvious, and in other limited cases like iterators. But usually I don't see much benefit: extra characters are cheap, and if you have a really long templated type you're using often you probably should make a typedef anyway.

kkoopa:
> Please, no. In my opinion, Herb Sutter's reasoning severely underweights the importance of readability / > reviewability. As someone who reviews a lot of code (usually not in an IDE that can help with type info)
>I find over-use of auto costs extra time and mental effort for the reviewer. Maybe small amounts in each
> instance, but it can add up.

 I think he makes a fair point about that being coding against implementation and not against interfaces.

> The third and most common argument concerns readability: “My code gets unreadable quickly when I don’t know what exact type my variable is without hunting around to see what that function or expression returns, so I can’t just use auto all the time.” There is truth to this, including losing the ability to search for occurrences of specific types when using the non-typed syntax auto x = expr; in 4(a) below, so this appears at first to be a strong argument. And it’s true that any feature can be overused. However, > I think this argument is actually weaker than it first seems for four reasons, two minor and two major.
>
> The two minor counterarguments are:
>
> The “can’t use auto” part isn’t actually true, because as we just saw above you can be explicit about your type and still use auto, with good benefit.
> The argument doesn’t apply when you’re using an IDE, because you can always tell the exact type, for example by hovering over the variable. Granted, this mitigation goes away when you leave the IDE, such as if you print the code.
> But we should focus on the two major counterarguments:
>
> It reflects a bias to code against implementations, not interfaces. Overcommitting to explicit types makes code less generic and more interdependent, and therefore more brittle and limited. It runs counter to the excellent reasons to “write code against interfaces, not implementations” we saw in Question 2.
> We (meaning you) already ignore actual types all the time…
```c++
template<class Container, class Value>
void append_unique( Container& c, const Value& v ) {
    if( find(begin(c), end(c), v) == end(c) )
        c.emplace_back(v); 
    assert( !c.empty() );
}
```
https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/

However, I do not care strongly enough to argue about this. In the end, it is the person actually writing the code who decides what said code is. (Code review cannot unconditionally force someone to do anything, there is always the option of retracting the merge request.)

jasongin:
Yes, I read that post from Herb Sutter. While I accept his reasons in favor of using `auto`, I think in practice for most codebases and teams they are not that significant compared to the importance of readability, which he dismisses by suggesting that IDEs can mitigate it somewhat (but we're all reviewing code on GitHub here, not in an IDE), and trying to convince us we really shouldn't care about the type. But as a code-reviewer I very often _do_ care about what the type is, because I want to go find the header comments or implementation or search for other uses of that type, to better understand the code and potentially spot problems.

Anyway, I don't mean to be argumentative here, just sharing my perspective. I will absolutely yield to consensus, if there is one.

napi-inl.h
bnoordhuis: Anecdotal: I use `auto` pretty heavily and I consider it idiomatic for the node code base (but I introduced most uses so of course I would say that.)

> iirc even Nan and Node core don’t follow the same rules.

Nan is still C++03 (I think) because of v0.10 and v0.12 compatibility.

kkoopa:
Oh, don't get me wrong, I love debating minutiae as much as the next pedant. ;)
However, this is probably not the right place.

As I see it, if you are going to do most of the reviewing of the code here (seems likely as you wrote most of it) and you feel a certain style makes that easier, then by all means let it be so.

kkoopa:
I did not see anything else to remark upon, so it should be good to go.

jasongin:
Oh, @addaleax I will work on fixing [the issue with the `void*` callback data](https://github.com/nodejs/node-api/pull/19#discussion_r111223760), unless you've already started on it.

jasongin:
@addaleax I merged https://github.com/nodejs/node-api/pull/18 so you'll need to rebase. It should be easy though since there weren't any changes from what you had.

addaleax:
@jasongin Go for it. :) If you want to hear my thoughts:
- I’d consider a template-typed solution like what V8’s `WeakCallbackInfo` does (basically use `T*` rather than `void*`, that’s just more C++-y)
- I would not drop the current `Function::New` signatures, they are already pretty usable with e.g. lambdas. I think you could actually build the extra-`data` version on top of it, if you like.

addaleax:
> I merged #18 so you'll need to rebase. It should be easy though since there weren't any changes from what you had.

done!

jasongin:
> I would not drop the current Function::New signatures, they are already pretty usable with e.g. lambdas. I think you could actually build the extra-data version on top of it, if you like.

Yes, in a C++ API we don't really _need_ to offer a data pointer, since as you say the same effect can be achieved using lambda captures. Still I guess it might be convenient for devs less comfortable using lambdas?

addaleax:
> Still I guess it might be convenient for devs less comfortable using lambdas?

Yup, I agree it’s a good idea to have an API like that. Wasn’t trying to imply anything else :)

kkoopa:
Would not something like this cover both:
```c++
template<typename T>
Function<T> Function::New(napi_env env, FunctionCallback cb, const char* utf8name, T* data = nullptr);
Function<void> Function::New(napi_env env, FunctionCallback cb, const char* utf8name, void* data = nullptr);
```

addaleax:
@kkoopa That depends … I guess I thought it would be more straightforward to pass `data` as an extra argument to `cb`, which would require different signatures for `cb` depending on the variant. But I do now realize I didn’t say that out loud. :smile:

I guess @jasongin will come up with something reasonable and we can discuss the details there.

kkoopa:
Oh, I meant the templating to be on the FunctionCallback.

addaleax:
Can/should I merge this?

PR #21, opened 2017-04-13
Add migration tool for conversion from NAN to N-API

sampsongao:


jasongin:
I used a very similar approach here: https://github.com/jasongin/node-api/blob/convert/convert-to-napi.js

We should combine the best from both.

README.md
```diff
@@ -35,6 +35,22 @@ To use N-API in a native module:
 #include "napi.h"
 ```
 
+To make the migration to node-api easier, we have also provide a script to do these steps above for you. To use the conversion script:
```
mhdawson: "we have also provide" -> "we have provided"
"to do these steps above for you" -> "to help complete the steps listed above"

README.md
```diff
@@ -35,6 +35,22 @@ To use N-API in a native module:
...
+```
+  2. Install node-api module
+```
+npm install node-api
+```
+  3. Run node-api conversion script
+```
+node ./node_modules/node-api/tools/conversion.js ./
+```
+  4. This script is to make the migration easier and it is possible to miss out some conversion. So next step is trying to build your module and fix what the script is missing.
```
mhdawson: How about this instead:

This script converts as much of your code to using the node-api wrapper as possible, however, the conversion will likely be incomplete.  Once converted, the next step is
trying to build your module and fixing the parts that the script could not convert for you.  We currently expect the script to be able to complete about 80% of the required conversions.

mhdawson:
I think the intro (not being changed by this PR) in the README.md could also use some updates but I'll PR those suggestions one this one lands.

README.md
sampsongao: I think we shouldn't include the "80%" here, since we don't have method to justify it.

mhdawson: we should give them some sort of idea of how much will be covered. Something less precise is fine but I think we want them to have a feeling for if most of the work will be done or just a small part.

sampsongao: How about we say "We tried our best to cover all common replacement patterns. This script should be able to handle majority of the migration work."

README.md
```diff
@@ -35,6 +35,22 @@ To use N-API in a native module:
 #include "napi.h"
 ```
 
+To make the migration to node-api easier, we have provided a script to help complete the steps listed above. To use the conversion script:
```
jasongin: Add a `## Conversion tool` header here to separate this from the previous section.

README.md
```diff
@@ -35,6 +35,22 @@ To use N-API in a native module:
...
+npm install node-api
+```
+  3. Run node-api conversion script
+```
+node ./node_modules/node-api/tools/conversion.js ./
+```
+  4. While this script makes conversion easier, it still cannot fully convert the module.  The next step is to try to build the module and complete the remaining conversions necessary to allow it to compile and pass all of the module's tests.
+
+
 At build time, the N-API back-compat library code will be used only when the targeted node version *does not* have N-API built-in.
```
jasongin: Move this line up before the conversion tool section, because it is related to the .gyp build logic above.

sampsongao:
Landed https://github.com/nodejs/node-addon-api/commit/16fc9a5db27490d3b2522995143580a6ded590d6

PR #24, opened 2017-04-17
Set up unit tests

jasongin:
- Set up a pattern for organizing a .cc and .js test files
   roughly corresponding to each class to be tested
 - Add unit tests for the Function and Error classes
 - Update test binding.gyp file to build on Windows (fix quotes)
 - Update test binding.gyp and README to enable exceptions with MSVC
 - Fix type of CallbackInfo::This

### DIFF
--- a/test/error.js
+++ b/test/error.js
@@ -1,0 +1,40 @@
+'use strict';
+
+assert.throws(() => binding.error.throwError('test'), err => {
+   return err instanceof Error && err.message === 'test';
+});
+
+assert.throws(() => binding.error.throwTypeError('test'), err => {
+   return err instanceof TypeError && err.message === 'test';
+});
+
+assert.throws(() => binding.error.throwRangeError('test'), err => {
+   return err instanceof RangeError && err.message === 'test';
+});
+
+assert.throws(
+  () => binding.error.doNotCatch(
+    () => {
+      throw new TypeError('test');
+    }),
+  err => {
+    return err instanceof TypeError && err.message === 'test' && !err.caught;
+  });
+
+assert.throws(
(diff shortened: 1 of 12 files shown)


test/error.js
```diff
@@ -0,0 +1,40 @@
...
+      throw new TypeError('test');
+    }),
+  err => {
+    return err instanceof TypeError && err.message === 'test' && err.caught;
+  });
+
+let err = binding.error.catchError(
+   () => { throw new TypeError('test'); });
+assert(err instanceof TypeError);
+assert.equal(err.message, 'test');
```
addaleax: Just fyi, this is how we’d write this in Node core right now:
```js
{
  const err = binding.error.catchError(
     () => { throw new TypeError('test'); });
  assert(err instanceof TypeError);
  assert.strictEqual(err.message, 'test');
}
```

(`const` wherever possible, block scope + always using strict equality asserts)

test/index.js
```diff
@@ -1,5 +1,18 @@
 'use strict';
-const binding = require('./build/Release/binding.node');
-const assert = require('assert');
+global.buildType = process.config.target_defaults.default_configuration;
+global.binding = require(`./build/${buildType}/binding.node`);
+global.assert = require('assert');
```
addaleax: This means that somebody looking at the test files themselves might need to look up where those “magic” globals come from … I’d be very okay with just copying these lines into the test files themselves :)

That would also have the advantage that they are are runnable as standalone node scripts, which might be nice when the test suite is bigger and takes longer to run.

jasongin: Yes, I didn't really like this use of global either. Ability to run the test files individually is a good point.

PR #25, opened 2017-04-17
Update code to support N-API on older versions of node

jasongin:
**Note** there are two commits in this PR:

1. Copy the latest node_api* source files from the node master branch, to pick up all the changes from the past few weeks. _All these changes were reviewed when they went into node-master, so you don't need to re-review._

2. Re-enable building as an external native module and back-compat with Node v4. 

Recent changes introduced dependencies on internal node APIs:
 - `Environment::GetCurrent()`
 - `FatalError()`
 - `arraysize()`

To enable building this code as an external module, the internal APIs must be avoided. Some of these changes might be controversial, since there was some debate about them before. But previously we didn't consider that it was better to avoid internal APIs. (While we could use different code for different versions of node, that would make maintenance more difficult.)

Also replace use of `v8::Maybe::ToChecked()` (available only in Node >= 7) with `FromJust()`, which does exactly the same thing and works in Node v4.

We should consider applying the second commit here back to the code in the node master branch.


src/node_api.cc
```diff
@@ -570,8 +570,7 @@ class SetterCallbackWrapper
 
   /*virtual*/
   void SetReturnValue(napi_value value) override {
-    node::FatalError("napi_set_return_value",
-      "Cannot return a value from a setter callback.");
+    assert(false);  // Cannot return a value from a setter callback
```
gabrielschulhof: Isn't `assert()` a no-op on a release build of node? Then this will not accomplish what we want when built into node. If it did, you could use it for `CHECK_ENV()` as well.

gabrielschulhof: What about simply forward-declaring `node::FatalError()` if internals are unavailable? Would
```C++
#if not (defined(NODE_WANT_INTERNALS) && NODE_WANT_INTERNALS)
namespace node {
  void FatalError(const char *location, const char *message);
}
#endif
```
work?

addaleax: > Isn't `assert()` a no-op on a release build of node?

What matters is whether this code is compiled in debug mode or not, not what node was compiled as. But yes, in release mode this will be a no-op.

(Also, can you `#include <assert.h>` if you’re going with this?)

> What about simply forward-declaring `node::FatalError()` if internals are unavailable? 

Please don’t make code rely on things that Node has explicitly marked as internals, you’ll be sad when they go away (also, it wouldn’t work because node doesn’t `dllexport` them on Windows). If you *really* want access to `node::FatalError()`, a better option would be to open a PR that moves it to the public headers.

src/node_api.cc
```diff
@@ -2684,9 +2684,11 @@ napi_status napi_queue_async_work(napi_env env, napi_async_work work) {
...
   CHECK_ARG(env, work);
 
-  // Consider: Encapsulate the uv_loop_t into an opaque pointer parameter
-  uv_loop_t* event_loop =
-    node::Environment::GetCurrent(env->isolate)->event_loop();
+  // Consider: Encapsulate the uv_loop_t into an opaque pointer parameter.
+  // Currently the environment event loop is the same as the UV default loop.
+  // Someday (if node ever supports multiple isolates), it may be better to get
+  // the loop from node::Environment::GetCurrent(env->isolate)->event_loop();
+  uv_loop_t* event_loop = uv_default_loop();
```
addaleax: I’m not sure, this might not work when there are native modules that use a different loop, or Node embedders that use multiple loops… but `nan` does it too so it can’t be so terribly bad I guess?

src/node_api.cc
gabrielschulhof: https://github.com/gabrielschulhof/node-api/commit/8e5fe6ee3c0db3ccf8a6e5e6a371be2f1ac0cb69

addaleax: Yea, that’s why I said it won’t work on Windows. ;) On POSIX it will work right now but it’s most definitely not something we should do, copying all the relevant code from node.cc would be a better option than using internal symbols from the node binary.

gabrielschulhof: Oh, I see ... the `dllexport` thing. Right.

gabrielschulhof: I actually had the code copied earlier, when I got the "bright" idea to forward-declare. Lemme test that on both Linux, Windows, and OSX.

gabrielschulhof:
When building against node-api on Windows every N-API symbol remains undefined. Does gyp properly generate the rules to statically link node-api into its dependent?

gabrielschulhof:
On Windows you need to add `"<(PRODUCT_DIR)/node-api.lib"` to your "libraries" list, but only if N-API is external.

gabrielschulhof:
OK, that can be accomplished with
```gyp
"conditions": [
	[ "'<!(node -p \"require('node-api').isNodeApiBuiltin\")'=='false'", {
		"libraries+": [
			"<(PRODUCT_DIR)/node-api.lib"
		]
} ]
```
Yet, and assuming that file really does get linked in, the unresolved symbols persist. Perhaps the definition of NAPI_EXTERN is improper.

gabrielschulhof:
@jasongin can you help me out please? https://gist.github.com/gabrielschulhof/2aa53679e517ad9985052d71b7f4c11e works fine on Linux but I can't figure out why all N-API symbols are unresolved on Windows.

If it complains of symbols of the form `__imp_napi_*` does that mean they should be declared with `dllimport`?

gabrielschulhof:
Nope. I'm really grasping at straws here :)

src/node_api.cc
jasongin: Yes, I was aware `assert()` is a no-op in a release build. I don't think we need a fatal error here. If someone tries to set a return value from a property setter, it's not so bad to just ignore it in a release build.

addaleax: @jasongin I think this might be something that really surprises people when they set out to debug an issue in their addon and suddenly have to deal with more, unexpected, crashes that weren’t there in release mode…

src/node_api.cc
jasongin: Right, `nan` does this now. Other native modules are free to use non-default loops, if they call the libuv APIs directly, and that won't cause any problems with this.

Eventually, we might wrap more of the libuv APIs in N-API, and that would support non-default loops.

src/node_api.cc
jasongin: It shouldn't surprise anyone that `assert()` statements do extra checks in debug mode. V8 has lots of things like that (`DCHECK`).

jasongin:
@gabrielschulhof I haven't had problems building on Windows. The .gyp dependency mentioned in the README should cause the .lib file to be linked automatically:

```gyp
'dependencies': ["<!(node -p \"require('node-api').gyp\")"],
```

The APIs are static-linked, so there should be no need for `dllimport`.

gabrielschulhof:
I ended up compiling in N-API:
```gyp
"conditions": [
  [ "'<!(node -p \"require('node-api').isNodeApiBuiltin\")'=='false'", {
    "defines+": [ "EXTERNAL_NAPI" ],
    "include_dirs+": ["<!(node -p \"require('node-api').include\")"],
    "sources+": [
      "<!(node -p \"var path = require('path'); path.join(path.dirname(require.resolve('node-api')), 'src', 'node_api.cc')\")"
    ]
  } ]
]
```

jasongin:
I think all that should be replaced with the .gyp dependency.

gabrielschulhof:
@jasongin ... and the .gyp dependency works - on Linux, but not on Windows.

The fact is that by the nature of gyp you have to put so many calls to `<!(node -p "require(\"node-api\").<something>")` into your own project's gyp file that the dependency doesn't really provide that much.

src/node_api.cc
jasongin: > (Also, can you #include <assert.h> if you’re going with this?)

This is at the top of `napi-inl.h`:

```C++
#include <cassert>
```

gabrielschulhof:
Having it all in one condition like that is actually nice, because all calls to node-api are localized.

src/node_api.cc
addaleax: Ah, sorry, missed that. ;)

jasongin:
> the .gyp dependency works - on Linux, but not on Windows

It works for me on Windows. The tests at https://github.com/nodejs/node-api/pull/24 build that way.

gabrielschulhof:
Yeah, I dunno why it doesn't work for me. Must be some combination of VS/node.js/npm/node-gyp.

gabrielschulhof:
I have the toolchain from npm -g install windows-build-tools, so that's, like, the free version. Maybe that has some limitations.

src/node_api.cc
addaleax: But wait, that isn’t included from `node_api.cc` – it should probably really have a `cassert` include of its own?

jasongin:
I think it's more likely a node-gyp issue. What version of `node-gyp` are you using?

src/node_api.cc
jasongin: Oh, right!

gabrielschulhof:
3.5.0

src/node_api.cc
jasongin: Note if you implement a setter function in JavaScript (via `Object.defineProperty()`), and you return some value from the setter function, the value is simply ignored. That suggests it should be ignored here also.

gabrielschulhof:
Oh, my approach doesn't work with `npm install -g`, because `"sources"` must not contain any absolute paths, it seems. So, for me it's just a lot easier to ship a copy of `node_api.*` with iotivity-node and
```gyp
[ "'<!(node -p \"!!process.version.match(/^v8/)\")'=='false'", {
  "defines+": [ "EXTERNAL_NAPI" ],
  "include_dirs+": [ "src/n-api" ],
  "sources+": [ "src/n-api/node_api.cc" ]
} ]
```

jasongin:
Oh, for Windows I had to flip some quotes in the test `binding.gyp` for the dependency to work. That fix is in https://github.com/nodejs/node-api/pull/24. Try copying from there, or building with both that PR and this one combined.

gabrielschulhof:
Hmmm ... lemme try that.

gabrielschulhof:
Back to the unresolveds in Windows, with:

```gyp
{
	"targets": [ {
		"target_name": "hello",
		"sources": [ "hello.cc" ],
		"conditions": [
			[ "'<!(node -p \"require('node-api').isNodeApiBuiltin\")'=='false'", {
				"include_dirs+": ["<!(node -p \"require('node-api').include\")"],
				"dependencies+": ["<!(node -p \"require('node-api').gyp\")"],
			} ]
		]
	} ]
}
```

jasongin:
Try removing the condition. You shouldn't need to put those lines inside a conditional block anyway. The include path is needed either way, and the `require('node-api').gyp` is designed to return an empty dependency if the node API is builtin.

jasongin:
Also, I'm unfamiliar with `"dependencies+"` in `node-gyp`. Does that really do what you expect?

gabrielschulhof:
AFAIK a '+' suffix means "add to an existing list if present".

gabrielschulhof:
It's now down to
```gyp
{
	"targets": [ {
		"target_name": "hello",
		"sources": [ "hello.cc" ],
		"include_dirs": ["<!(node -p \"require('node-api').include\")"],
		"dependencies": ["<!(node -p \"require('node-api').gyp\")"]
	} ]
}
```
Still no joy.

gabrielschulhof:
... and I can see it building `node_api.cc`, and there's a `node_api.lib` of size 952776 in build\Release\hello, but still no joy.

gabrielschulhof:
Anyhoo - looks like we're headed for non-fatal errors here, but still, this is what it would look like to copy the (condensed version of) `node::FatalError()` into `node_api.cc`:

https://github.com/gabrielschulhof/node-api/commit/76f3a759864086021472ddc518f378003552a9b7

jasongin:
The .gyp dependency should get translated into a `<ProjectReference>` element in the generated `.vcxproj` file, that references `node-api.vcxproj`. Do you see something like this?

```xml
    <ProjectReference Include="path\to\node-api.vcxproj">
      <Project>{48DE6A98-BCC5-AED1-43A1-3BA1AC79F59F}</Project>
      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
    </ProjectReference>
```

gabrielschulhof:
It's there:
```xml
    <ProjectReference Include="..\node_modules\node-api\src\node-api.vcxproj">
      <Project>{C6B3FC4A-6125-FF58-8F61-7C7914BA71AA}</Project>
      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
    </ProjectReference>
```

jasongin:
Try this:

```cmd
cd build
"C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe" /t:Rebuild /p:Configuration=Release /p:Platform=x64 > build.log
```

Then put the `build.log` up on a gist and I'll compare it to what I see. In mine, I can clearly see the `node-api.lib` file being generated and then linked into the final target.

gabrielschulhof:
It's too bad `set V 1; npm install` has no effect in terms of making things verbose like it has on Linux ...

addaleax:
@gabrielschulhof long shot because I didn’t read all the context and am not a Windows person, but does `npm install --verbose` do what you need?

gabrielschulhof:
https://gist.github.com/gabrielschulhof/7967e960d00370474e7247452635e2c7

gabrielschulhof:
@addaleax actually, yes! Unfortunately, it doesn't help :/

jasongin:
Oh, I think I know how to fix your build: in `hello.cc`, change `#include "node_api.h"` to `#include "napi.h"`.

The reason is you need the [special `napi.h` that defines `EXTERNAL_NAPI`](https://github.com/nodejs/node-api/blob/master/src/napi.h) to get the non-dllimport linkage correct.

This might not be the best way to set things up. I'm open to suggestions, but that was the best solution I could come up with to support building against both built-in N-API and external N-API, without modifying the `node_api*` files.

src/node_api.cc
kkoopa: `nan` supports pick-and-mix. If the default loop is too restrictive, you can fall back to libuv. If the point here is to abstract away libuv, there needs to be a way of specifying a non-default loop without having to call the libuv API directly.

jasongin: As I said just above, these APIs are _not_ intended to entirely abstract away libuv, at least not at this time.

jasongin:
@gabrielschulhof gave verbal approval regarding the removal of `node::FatalError()`

With that and the other approvals, I'm merging this.

I will prepare a PR to node master with the changes from the second commit here, so that the sources stay in sync.

And since there are other changes in the pipeline there, we will of course need to update the sources here again, but then the changes will be relatively small.

PR #30, opened 2017-04-27
Test and fix AsyncWorker

jasongin:
- Remove `MakeCallback()` overloads that defaulted to undefined receiver, because `node::MakeCallback()` requires an object.
 - Use the async worker persistent object as the receiver of a worker callback
 - Persist async errors as strings, because an `Error` object cannot be created outside of a JS context
 - Remove overridable `AsyncWorker::WorkComplete()` because it wasn't useful and caused confusion. `OnOK()` and/or `OnError()` should be (optionally) overridden instead.
 - Add tests to validate basic success and error scenarios for `AsyncWorker`

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2684,16 +2684,7 @@ inline void AsyncWorker::OnWorkComplete(
         self->OnError(Error::New(self->_env, self->_error));
       }
     } catch (const Error& e) {
-      // A JS or N-API exception was thrown, and propagated as a C++ exception.
-      // But throwing a JS exception here would have no effect because there
-      // is no JS on the callstack.
-
-      // TODO: Print the exception details and abort, just like Node.js normally
-      // does for unhandled exceptions. But there is no N-API function for that.
-      // For now just assert, so at least the exception message is discoverable
-      // in a debug build.
-      const char* message = e.what();
-      assert(false);
+      e.ThrowAsJavaScriptException();
     }
   }
   delete self;
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1,4 +1,4 @@
-﻿#ifndef SRC_NAPI_INL_H_
+#ifndef SRC_NAPI_INL_H_
 #define SRC_NAPI_INL_H_
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -38,9 +38,7 @@ inline void RegisterModule(napi_env env,
                        Napi::Object(env, module));
   }
   catch (const Error& e) {
-    if (!Napi::Env(env).IsExceptionPending()) {
-      e.ThrowAsJavaScriptException();
-    }
+    e.ThrowAsJavaScriptException();
   }
 }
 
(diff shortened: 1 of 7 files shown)


jasongin:
@digitialinfinity please review, this is somewhat based on your feedback.

napi.h
```diff
@@ -850,18 +846,16 @@ namespace Napi {
     void Queue();
     void Cancel();
 
-    virtual void Execute() = 0;
-    virtual void WorkComplete();
-
     ObjectReference& Persistent();
 
   protected:
     explicit AsyncWorker(const Function& callback);
```
digitalinfinity: Were you still planning on having an overload that allowed the user to pass in the this pointer for use with the callbacks?

digitalinfinity:
>     T* instance = Unwrap(callbackInfo.This());

While you're touching this file, can you also fix this to convert this to an object since that's what Unwrap expects?

---
Refers to: napi-inl.h:2001 in 9a0062a. [](commit_id = 9a0062a482e2acbc5547f3e4b89275a2226dc3a1, deletion_comment = False)

napi.h
jasongin: I had been thinking that another object reference would be wasteful, but now I realize it could make sense to use the receiver object as the `_persistent` object reference, if one was supplied, instead of creating a new object. I'll update this PR.

napi.h
```diff
@@ -850,18 +846,16 @@ namespace Napi {
...
 
+    virtual void Execute() = 0;
     virtual void OnOK();
     virtual void OnError(Error e);
 
-    void SetError(Error error);
+    void SetError(std::string error);
```
addaleax: This should be `const std::string&`

napi-inl.h
```diff
@@ -2193,47 +2175,38 @@ inline void AsyncWorker::Cancel() {
...
 
 inline void AsyncWorker::OnExecute(napi_env env, void* this_pointer) {
   AsyncWorker* self = static_cast<AsyncWorker*>(this_pointer);
-  try {
-    self->Execute();
-  }
-  catch (const Error& e) {
-    self->SetError(e);
-  }
+  self->Execute();
```
addaleax: We could still store `error.what()` for any `std::exception` we run into … anything else is going to hard crash the process anyway, afaict?

jasongin:
I was working on updating this PR, but after adding more tests I ran into a bug https://github.com/nodejs/node-api/issues/31. I'm going to fix that first, then come back to this.

jasongin:
I rebased on top of the [error fix](https://github.com/nodejs/node-api/pull/32), and updated to address review feedback so far. So when reviewing here, **please only look at the second commit**.

I encountered an interesting problem: how to deal with unhandled JS exceptions during an async callback. See the TODO comment in `AsyncWorker::OnWorkComplete()`. Apparently this is something that Nan (or `node::MakeCallback()`) doesn't really address, but ideally this wrapper should do something better to provide consistent error-handling behavior. I think it will require a new function added to N-API. If others agree, I'll open a bug to track that.

jasongin:
Related to the TODO mentioned above, I'm working on a change in the C APIs that will call `node::FatalException()` (which is a public API, [unlike `node::FatalError`](https://github.com/nodejs/abi-stable-node/issues/232)) if a JavaScript exception escapes from a `napi_async_complete_callback`. Then `AsyncWorker::OnWorkComplete()` should just rethrow any `Napi::Error` as a JavaScript exception (just like other function callbacks already do via the `NAPI_RETHROW_JS_ERROR` macro).

jasongin:
Related node PR: https://github.com/nodejs/node/pull/12838

PR #32, opened 2017-05-01
Make Error a persistent reference

jasongin:
A `Napi::Error` instance must hold a persistent reference to the JS error object so that it can escape from handle scopes when thrown as a C++ exception.

Fixes: https://github.com/nodejs/node-api/issues/31

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1381,6 +1381,8 @@ inline void Buffer<T>::EnsureInfo() const {
 ////////////////////////////////////////////////////////////////////////////////
 
 inline Error Error::New(napi_env env) {
+  HandleScope scope(env);
+
   napi_status status;
   napi_value error = nullptr;
   if (Napi::Env(env).IsExceptionPending()) {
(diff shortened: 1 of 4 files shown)


napi-inl.h
```diff
@@ -1432,16 +1434,49 @@ inline Error Error::New(napi_env env, const std::string& message) {
...
 }
 
-inline Error::Error(napi_env env, napi_value value) : Object(env, value) {
+inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullptr) {
+  if (value != nullptr) {
+    napi_status status = napi_create_reference(env, value, 1, &_ref);
+
+    // Avoid infinite recursion in the failure case.
+    // Don't try to construct & throw another Error instance.
+    assert(status == napi_ok);
```
boingoing: Should we cause a fatal error here to kill the process?

jasongin: Probably. See also a few other places in this `Error` class where error-handling code can't throw errors.

We'd need a `napi_fatal_error()` API or similar.

jasongin: I'll leave this the way it is for now. Fatal error discussion is at https://github.com/nodejs/abi-stable-node/issues/232

PR #34, opened 2017-05-03
Set up doc-comments and doxygen

jasongin:
- Add doc-comments to a few of the classes in `napi.h`: `Value`, `String`, `Error`
 - Add a doxygen config file

To generate documents:
1. Download and install doxygen from http://www.stack.nl/~dimitri/doxygen/download.html
2. Run the "doxywizard" (GUI) or doxygen.exe (command-line) using the `Doxyfile` configuration file added in this change.

Doxygen supports multiple different doc-comment styles and text formatting methods. Here, I have chosen:
 - Triple-slash style doc-comments, because they make it easier to add end-of-line comments for method parameters, and for simple property-style methods that are common in these APIs.
 - Markdown for text formatting, because we're all familiar with markdown and it's nicer to read compared to the `@command` or `\command` formatting instructions that are also supported. We can still mix in any of [Doxygen's supported commands](http://www.stack.nl/~dimitri/doxygen/manual/commands.html).

Doxygen also supports Javadoc style comments (`/** ... */`) but those would not be as nice here IMO.


jasongin:
While someone has created a [doxygen npm package](https://github.com/EruantalonJS/node-doxygen), it apparently isn't widely used. I tried it out and it doesn't work very well: there's no way to download the doxygen binaries only when needed (so it's very slow when running repeatedly), and it doesn't show progress while downloading or running.

jasongin:
Related: https://github.com/nodejs/node-api/issues/6 Write documentation for C++ APIs

napi.h
```diff
@@ -72,51 +80,63 @@ namespace Napi {
...
+    bool IsArrayBuffer() const; ///< Tests if a value is a JavaScript array buffer.
+    bool IsTypedArray() const;  ///< Tests if a value is a JavaScript typed array.
+    bool IsObject() const;      ///< Tests if a value is a JavaScript object.
+    bool IsFunction() const;    ///< Tests if a value is a JavaScript function.
+    bool IsBuffer() const;      ///< Tests if a value is a Node buffer.
+
+    /// Converts to another type of `Napi::Value`, when the actual type is known or assumed.
+    ///
+    /// This conversion does NOT coerce the type. Calling any methods inappropriate for the actual
+    /// value type will cause `Napi::Error`s to be thrown.
```
boingoing: Just nitpicky, should we write this as such?

Calling any methods inappropriate for the actual value type will throw `Napi::Error`.

napi.h
```diff
@@ -151,28 +171,61 @@ namespace Napi {
...
+    /// Creates a new String value from a UTF-16 encoded C++ string.
+    static String New(
+      napi_env env,               ///< N-API environment
+      const std::u16string& value ///< UTF-16 encoded C++ string
+    );
+
+    /// Creates a new String value from a UTF-8 encoded C string.
+    static String New(
+      napi_env env,     ///< N-API environment
+      const char* value ///< UTF-8 encoded C string
```
boingoing: C string seems to imply "null-terminated" but should we explicitly mention that here since there is an overload which takes a length parameter?

napi.h
```diff
@@ -151,28 +171,61 @@ namespace Napi {
...
+    static String New(
+      napi_env env,         ///< N-API environment
+      const char16_t* value ///< UTF-16 encoded C string
+    );
+
+    /// Creates a new String value from a UTF-8 encoded C string with specified length.
+    static String New(
+      napi_env env,      ///< N-API environment
+      const char* value, ///< UTF-8 encoded C string
+      size_t length      ///< length of the string in bytes
```
boingoing: length is not capitalized, seems like it should be to be consistent with other comments.

napi.h
```diff
@@ -536,12 +590,11 @@ namespace Napi {
...
-   * Holds a counted reference to a value; initially a weak reference unless otherwise specified.
-   * May be changed to/from a strong reference by adjusting the refcount. The referenced value
-   * is not immediately destroyed when the reference count is zero; it merely then eligible for
-   * GC if there are no other references to the value.
-   */
+  /// Holds a counted reference to a value; initially a weak reference unless otherwise specified,
+  /// may be changed to/from a strong reference by adjusting the refcount.
+  ///
+  /// The referenced value is not immediately destroyed when the reference count is zero; it is
+  /// merely then eligible for if there are no other references to the value.
```
boingoing: Eligible for destruction?

napi.h
```diff
@@ -789,36 +841,31 @@ namespace Napi {
...
+  ///             InstanceMethod("doSomething", &DoSomething),
+  ///           });
+  ///           target.Set("Example", constructor);
+  ///         }
+  ///
+  ///         Example(const Napi::CallbackInfo& info); // Constructor
+  ///         Napi::Value GetSomething(const Napi::CallbackInfo& info);
+  ///         void SetSomething(const Napi::CallbackInfo& info, const Napi::Value& value);
+  ///         Napi::Value DoSomething(const Napi::CallbackInfo& info);
+  ///       }
```
boingoing: Nitpicky: This brace is not aligned with the class statement - looks like it's 2 spaces further right.

jasongin:
Thanks @boingoing. I pushed an update that fixed all the issues you commented on.

jasongin:
I was thinking an easy way to publish the generated API docs would be to use GitHub Pages. Based on [the GitHub help](https://help.github.com/articles/user-organization-and-project-pages/#project-pages), I created a `gh-pages` branch in this repo and pushed a commit with all the generated docs.

@mhdawson Can you go to the settings for this repo and enable GitHub Pages, using the `gh-pages` branch? Then I think the docs should be visible at https://nodejs.github.io/node-api/

jasongin:
Docs are visible now at https://nodejs.github.io/node-api/

jasongin:
@digitalinfinity @gabrielschulhof @mhdawson @addaleax Any feedback on the documentation format?

README.md
```diff
@@ -5,6 +5,13 @@ Node.js API (N-API), along with library code that enables
 backward-compatibility with use with older versions of Node.js that do
 not have N-API built-in.
 
+### API Documentation
+
+ - [ABI-Stable C APIs in Node.js](https://nodejs.org/dist/latest-v7.x/docs/api/n-api.html)
+ - [C++ APIs in this package](https://nodejs.github.io/node-api/namespace_napi.html)
```
addaleax: Aaaah this is so cool! Thank you for setting it up!

README.md
```diff
@@ -5,6 +5,13 @@ Node.js API (N-API), along with library code that enables
 backward-compatibility with use with older versions of Node.js that do
 not have N-API built-in.
 
+### API Documentation
+
+ - [ABI-Stable C APIs in Node.js](https://nodejs.org/dist/latest-v7.x/docs/api/n-api.html)
```
addaleax: `latest-v7.x` only works by accident (and it might/should get deleted) – I would just stick with linking to https://nodejs.org/api/n-api.html and accept that it only works once the 8.0.0 release is out

napi.h
```diff
@@ -441,12 +494,105 @@ namespace Napi {
...
+  /// argument), then the N-API wrapper automatically converts and throws the error as a C++
+  /// exception of type `Napi::Error`. Or if a JavaScript function called by C++ code via N-API
+  /// throws a JavaScript exception, then the N-API wrapper automatically converts and throws it as
+  /// a C++ exception of type `Napi::Error`.
+  ///
+  /// If a C++ exception of type `Napi::Error` escapes from a N-API C++ callback, then the N-API
+  /// wrapper automatically converts and throws it as a JavaScript exception.
+  ///
+  /// Catching a C++ exception of type `Napi::Error` also clears the JavaScript exception. Of
+  /// course it may be then re-thrown, which restores the JavaScript exception.
```
addaleax: This paragraph sounds a bit confusing (it sounds like C++ exception management affects the JS VM’s state), I would be okay with just deleting it.

napi.h
```diff
@@ -441,12 +494,105 @@ namespace Napi {
...
+  /// thrown and caught as a C++ exception.
+  ///
+  /// If a N-API call fails without executing any JavaScript code (for example due to an invalid
+  /// argument), then the N-API wrapper automatically converts and throws the error as a C++
+  /// exception of type `Napi::Error`. Or if a JavaScript function called by C++ code via N-API
+  /// throws a JavaScript exception, then the N-API wrapper automatically converts and throws it as
+  /// a C++ exception of type `Napi::Error`.
+  ///
+  /// If a C++ exception of type `Napi::Error` escapes from a N-API C++ callback, then the N-API
+  /// wrapper automatically converts and throws it as a JavaScript exception.
```
addaleax: By the way, should we catch all exceptions coming from C++ callbacks and convert them to JS errors? I don’t think there’s a way to do anything *more* meaningful (or at least I would assume JS engines don’t like it if the call stack unwinds outside of their control)

napi.h
```diff
@@ -441,12 +494,105 @@ namespace Napi {
...
+  /// #### Example 1 - Throwing an exception:
+  ///
+  ///     Napi::Env env = ...
+  ///     throw Napi::Error::New(env, "Example exception");
+  ///
+  /// Following C++ statements will not be executed. The exception will bubble up as a C++
+  /// exception of type `Napi::Error`, until it is either caught while still in C++, or else
+  /// automatically re-thrown as a JavaScript exception when the callback returns to JavaScript.
+  ///
+  /// #### Example 2 - Ignoring a NAPI exception:
```
addaleax: s/Ignoring/not catching/?

napi.h
```diff
@@ -441,12 +494,105 @@ namespace Napi {
...
+  ///
+  ///     Napi::Function jsFunctionThatThrows = someObj.As<Napi::Function>();
+  ///     jsFunctionThatThrows({ arg1, arg2 });
+  ///
+  /// Following C++ statements will not be executed. The exception will bubble up as a C++
+  /// exception of type `Napi::Error`, until it is either caught while still in C++, or else
+  /// automatically re-thrown as a JavaScript exception when the callback returns to JavaScript.
+  ///
+  /// #### Example 3 - Handling a NAPI exception:
```
addaleax: N-API?

mhdawson:
Sorry for the late response, I've been out all this week.  Looks like you 
found somebody else to configure it.



Regards,

Michael (M.H.) Dawson
Runtime Technologies Node.js Technical Lead
Software Developer and Master Inventor


Phone: 1-613-356-5484 | Phone: 1-343-882-2473
E-mail: Michael_Dawson@ca.ibm.com
Find me on:   


3755 Riverside Drive
Ottawa, ON K1V 1B8
Canada

napi.h
jasongin: I'm not sure. I suppose we _could_ catch any kind of `std::exception` and then copy the message into a JS error. But would that be safe, or could some exceptions leave the module in an inconsistent state?

The alternative is to just let any _other_ kinds of uncaught C++ exceptions result in a crash, which is what happens now.

README.md
jasongin: Yes. I wanted to link to https://github.com/nodejs/node/blob/master/doc/api/n-api.md for now, but links to external markdown files confuse the Doxygen generator.

napi.h
jasongin: Reworded and combined with the previous paragraph.

napi.h
jasongin: Fixed.

napi.h
jasongin: Fixed.

PR #38, opened 2017-05-15
Exit test script with exit code 1 in case of failure

addaleax:
Just a small thing I noticed.

(I’m currently trying to set up Travis/AppVeyor CI support, btw – I might open a WIP PR shortly. It won’t work for the Node.js org repo, but it’s something that could be enabled in the forks we are working from.)

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -20,5 +20,6 @@ testModules.forEach(name => {
    }
    catch (e) {
       console.error(e);
+      process.exitCode = 1;
    }
 });


jasongin:
> I’m currently trying to set up Travis/AppVeyor CI support, btw – I might open a WIP PR shortly.

Cool. I was planning to do that soon, but since you're on that I'll work on more documentation instead.

> It won’t work for the Node.js org repo

Why not? Can we not enable the necessary integration in this repo?

addaleax:
> > It won’t work for the Node.js org repo

> Why not? Can we not enable the necessary integration in this repo?

We can’t give Travis access to the organization because doing so would give it access to private repos containing sensitive information… we’ve talked about this a lot :/

jasongin:
That's unfortunate. So should we look into setting up Jenkins (ci.nodejs.org) instead?

addaleax:
That’s what we have typically been doing so far, yes…

mhdawson:
The other issue with Travis is that it does not provide the platform coverage we have in the Node.js CI.

I do have it on my list to set something up, just have not had time to get to it yet.

Having said that @jasongin we have set to be able to let WG's manage edit jobs so if want to help setup the jenkins job I can create the empty job and you should be able to configure it.

kkoopa:
Travis still works without giving access, it is onlt the GitHub hooks that don't. We use Travis and AppVeyor in NAN. Just have to remember to check the build status from Travis or AppVeyor site. Travis were nice enough to also give us access to OS X as well when approached.

jasongin:
@kkoopa Without hooks, then do you have to manually _start_ the CI also?

kkoopa:
No, all it affects is that the CI status thing is not visible in pull requests.

kkoopa:
Regarding travis permissions, they list them:
https://docs.travis-ci.com/user/github-oauth-scopes/

read:org (read-only) When you’re logged in on Travis CI, we show you all of your repositories, including the ones from any organization you’re part of. The GitHub API hides any organizations you’re a private member of without this scope. So to make sure we show you all of your repositories, we require this scope. **Note that this scope allows access to the basic information about both private and public repositories, but not on any of the data or code stored in them.**

addaleax:
> Travis were nice enough to also give us access to OS X as well when approached.

Just fyi, I’m pretty sure it’s no longer necessary to ask them, OS X is now supported for all packages that want it. :)

addaleax:
Sure, I’ll close this. :)

PR #39, opened 2017-05-16
TypedArray improvements and tests

jasongin:
- Rename `TypedArray_<>` to `TypedArrayOf<>`.
 - Remove the second template parameter from `TypedArrayOf<>`, because it can be (usually) inferred from the first parameter and is only needed at object construction time.
 - Remove the `TypedArray::As*Array()` methods, which were redundant because they did the same thing as `Value::As()`.
 - Add doc-comments to `ArrayBuffer`, `TypedArray`, and `TypedArrayOf<>` classes.
 - Add tests for typed arrays.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -979,21 +979,23 @@ inline void ArrayBuffer::EnsureInfo() const {
 // TypedArray class
 ////////////////////////////////////////////////////////////////////////////////
 
-inline TypedArray::TypedArray() : Object(), _type(TypedArray::unknown_type), _length(0) {
+inline TypedArray::TypedArray()
+  : Object(), _type(TypedArray::unknown_array_type), _length(0) {
 }
 
 inline TypedArray::TypedArray(napi_env env, napi_value value)
-  : Object(env, value), _type(TypedArray::unknown_type), _length(0) {
+  : Object(env, value), _type(TypedArray::unknown_array_type), _length(0) {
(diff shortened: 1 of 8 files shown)


napi-inl.h
```diff
@@ -1054,99 +1056,75 @@ inline Napi::ArrayBuffer TypedArray::ArrayBuffer() const {
...
+inline TypedArrayOf<T>::TypedArrayOf(napi_env env,
+                                     napi_value value,
+                                     napi_typedarray_type type,
+                                     size_t length,
+                                     T* data)
+  : TypedArray(env, value, type, length), _data(data) {
+  if (!(type == TypedArrayTypeForPrimitiveType<T>() ||
+      (type == napi_uint8_clamped_array && std::is_same<T, uint8_t>::value))) {
+    throw TypeError::New(env, "Array type must match the template parameter. "
+    " (Uint8 arrays may optionally have the \"clamped\" array type.)");
```
addaleax: This will result in a double space when the strings are concatenated across lines, I’m not sure that’s intended? (Also, I think I’d just spell it as `Uint8Arrays`)

jasongin: I'll remove the extra space.

I didn't refer to `Uint8Array` specifically because that's just a typedef; Some people may prefer to just use `TypedArrayOf<uint8_t>` instead.

Issue #41, opened 2017-05-16
.defines instead of isNodeApiBuiltin


gabrielschulhof:
Would it make more sense to have a `require( "node-addon-api" ).defines` instead of `.isNodeApiBuiltin`? Conditionally adding the `EXTERNAL_NAPI` preprocessor directive is my only use case for examining the value of `.isNodeApiBuiltin`, and on Windows the define is required, otherwise there will be unresolved symbols.

gabrielschulhof:
Also, all the other `node-addon-api` exports can be added unconditionally to either the `"include_dirs"` or `"dependencies"` of a dependent target, because the conditionality is encapsulated in `index.js`.

gabrielschulhof:
I guess it doesn't really cost that much to keep `.isNodeApiBuiltin` so we export something that can be used as a condition statement for unforeseen use cases, but we should definitely have the `.defines` in addition.

jasongin:
The `.isNodeApiBuiltin` export was meant to be exposed as a runtime flag in case anybody wanted to write code to check that. But it's not used anywhere today, and there should be no need to use it at build time.

Did you follow the instructions in the README? You shouldn't need to define `EXTERNAL_NAPI` anywhere in _your_ module's `node.gyp`. That is taken care of by these two lines to be added to your `node.gyp`:

```
  'include_dirs': ["<!(node -p \"require('node-addon-api').include\")"],
  'dependencies': ["<!(node -p \"require('node-addon-api').gyp\")"],
```

Then when you `#include "napi.lh"` that should get `src/napi.h` that has `#define EXTERNAL_NAPI` at the top.

gabrielschulhof:
I don't use the C++ API.

gabrielschulhof:
https://github.com/gabrielschulhof/iotivity-node/blob/node-addon-api/binding.gyp#L221-L232

gabrielschulhof:
OK, I guess I can include `napi.h` directly, because I'm doing C++ anyway.

gabrielschulhof:
I can't use napi.h because that requires exception handling to be enabled, which, for the plain C API I don't need.

jasongin:
Hmm... using only the C APIs via this package wasn't a scenario I considered.

gabrielschulhof:
Well, adding `.defines` and a mention in the `README.md` should cover plain C.

jasongin:
[...]

PR #42, opened 2017-05-16
Include-path-juggle node_api.h instead of napi.h (#41)

gabrielschulhof:
Instead of making the path from which napi.h is included conditional
upon the Node.js version, make the path of node_api.h thusly
conditional. This way, no additional modification needs to be made to a
dependent's binding.gyp in order to add the EXTERNAL_NAPI preprocessor
directive when the C++ wrappers are not used.

By this mechanism the C++ wrapper napi.h will also refer to the correct
version of node_api.h.

~~Fixes https://github.com/nodejs/node-api/issues/40~~
Fixes https://github.com/nodejs/node-api/issues/41

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ To use N-API in a native module:
 
   2. Reference this package's include directory and gyp file in `binding.gyp`:
 ```gyp
-  'include_dirs': ["<!(node -p \"require('node-addon-api').include\")"],
+  'include_dirs': ["<!@(node -p \"require('node-addon-api').include\")"],
   'dependencies': ["<!(node -p \"require('node-addon-api').gyp\")"],
 ```
 
--- a/external-napi/node_api.h
+++ b/external-napi/node_api.h
@@ -1,0 +1,7 @@
+#ifndef EXTERNAL_NODE_API_H_
+#define EXTERNAL_NODE_API_H_
+
+#define EXTERNAL_NAPI
+#include "../src/node_api.h"
+
+#endif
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -16,13 +16,13 @@ To use N-API in a native module:
...
   "dependencies": {
-    "node-addon-api": "github:nodejs/node-addon-api",
+    "node-addon-api": "github:nodejs/node-api",
   }
 ```
 
   2. Reference this package's include directory and gyp file in `binding.gyp`:
 ```gyp
-  'include_dirs': ["<!(node -p \"require('node-addon-api').include\")"],
+  'include_dirs': ["<!@(node -p \"require('node-addon-api').include\")"],
```
jasongin: What does the `@` do?

README.md
```diff
@@ -16,13 +16,13 @@ To use N-API in a native module:
   1. Add a dependency on this package to `package.json`:
 ```json
   "dependencies": {
-    "node-addon-api": "github:nodejs/node-addon-api",
+    "node-addon-api": "github:nodejs/node-api",
```
jasongin: I was expecting this repo would actually get renamed to `node-addon-api`. @mhdawson what do you think?

README.md
gabrielschulhof: It indicates that the output of the command is to be used to fill out an array. In this case, it uses the double quotes I add in index.js to correctly split the output into array elements, stripping the quotes, so the paths are not misinterpreted.

gabrielschulhof:
@jasongin the PR no longer changes the README.md example showing how to depend on the repo.

jasongin:
@gabrielschulhof I published `node-addon-api@0.3.0` to NPM, which includes this change. So, your library can depend on it by name@version now instead of a repo URI.

PR #43, opened 2017-05-18
Fix reporting of API error messages

jasongin:
When a call to an N-API function caused an error for some reason other than a JS exception, the fallback error message "Error in native callback" was always reported because of incorrect logic in `Napi::Error::New()`.

Then that fix exposed a bug in `napi_get_last_error_info()`, which I have fixed here and also at https://github.com/nodejs/node/pull/13087

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1373,17 +1373,17 @@ inline void Buffer<T>::EnsureInfo() const {
 inline Error Error::New(napi_env env) {
   napi_status status;
   napi_value error = nullptr;
-  if (Napi::Env(env).IsExceptionPending()) {
-    status = napi_get_and_clear_last_exception(env, &error);
-    assert(status == napi_ok);
-  }
-  else {
-    // No JS exception is pending, so check for NAPI error info.
-    const napi_extended_error_info* info;
-    status = napi_get_last_error_info(env, &info);
-    assert(status == napi_ok);
 
-    if (status == napi_ok) {
+  const napi_extended_error_info* info;
+  status = napi_get_last_error_info(env, &info);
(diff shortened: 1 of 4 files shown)


napi-inl.h
```diff
@@ -1373,17 +1373,17 @@ inline void Buffer<T>::EnsureInfo() const {
...
-    const napi_extended_error_info* info;
-    status = napi_get_last_error_info(env, &info);
-    assert(status == napi_ok);
 
-    if (status == napi_ok) {
+  const napi_extended_error_info* info;
+  status = napi_get_last_error_info(env, &info);
+  assert(status == napi_ok);
+
+  if (status == napi_ok) {
```
addaleax: Does this make sense? `assert` + then make the code conditional on asserted statement?

jasongin: If we encounter another error while handling an error, there's not much we can do. Probably it should be a fatal error, except we haven't exposed `node::FatalError()` via N-API [yet](https://github.com/nodejs/abi-stable-node/issues/232). The assert will at least indicate when something went wrong in a debug build, but in a release build the conditional allows it to continue without crashing.

jasongin: This came up last time this code was reviewed: https://github.com/nodejs/node-addon-api/pull/32#discussion_r114903602

Probably we should go ahead and add a `napi_fatal_error()` API.

jasongin: I opened a new issue to track this: https://github.com/nodejs/node-addon-api/issues/48
I don't think the current PR needs to be blocked by it.

jasongin:
@addaleax @boingoing @mhdawson Can someone review this? Thanks!

PR #49, opened 2017-05-31
Update CI to include node 4, 6, 8, nightly, CC nightly

jasongin:
Travis run: https://travis-ci.org/jasongin/node-addon-api/builds/237732241
AppVeyor run: https://ci.appveyor.com/project/jasongin/node-api/build/3

All jobs in the matrix are passing with the exception of those that use node nightly builds, which fail due to https://github.com/nodejs/build/issues/743

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,21 @@ language: c++
 compiler:
   - clang
   - gcc
-env:
-  - TRAVIS_NODE_VERSION="4"
-  - TRAVIS_NODE_VERSION="6"
-  - TRAVIS_NODE_VERSION="7"
+# For Linux, use an Ubuntu 14 image
+dist: trusty
 os:
   - linux
   - osx
+env:
+  global:
+    # https://github.com/jasongin/nvs/blob/master/doc/CI.md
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -23,21 +31,22 @@ addons:
...
   - echo "CFLAGS=\"$CFLAGS\" LDFLAGS=\"$LDFLAGS\""
+  # nvs
+  - git clone --branch v$NVS_VERSION --depth 1 https://github.com/jasongin/nvs ~/.nvs
+  - . ~/.nvs/nvs.sh
+  - nvs --version
+  # node.js
+  - nvs add $NODEJS_VERSION
+  - nvs use $NODEJS_VERSION
+  - node --version
+  - npm --version
```
kunalspathak: curious - Is this just use to log which `node/npm` version was used?

kunalspathak:
:shipit:

.travis.yml
jasongin: Yes, just for reference when something fails. Most CI scripts I've seen do this.

PR #50, opened 2017-06-01
Update README conversion steps with package name

kirbysayshi:
node-api is an existing package, so I was pretty confused when there was no tools folder!


mathiask88:
This hit me just now :) Should be an easy merge.

jasongin:
Merged, thanks!

PR #54, opened 2017-06-08
Add more documentation

jasongin:
- Add doc-comments to the `Boolean`, `Number`, `Symbol`, `Object`,
   and `PropertyLValue` classes.
 - Make `PropertyLValue` a nested class inside `Object`, because it
   is only ever used by `Object` and would almost never be referenced
   explicitly.
 - Make some `Object` method parameter names more consistent.
 - Add missing `Object::Set()` overload.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ To use N-API in a native module:
   1. Add a dependency on this package to `package.json`:
 ```json
   "dependencies": {
-    "node-addon-api": "0.2.0",
+    "node-addon-api": "0.3.3",
   }
 ```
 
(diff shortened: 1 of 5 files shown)


mateodelnorte:
Are there any minimalist examples of using node-addon-api anywhere?

jasongin:
@mateodelnorte Check out [`generator-napi-module`](https://github.com/digitalinfinity/generator-napi-module) - it is a yeoman generator that helps you get started with a basic "hello world" module that uses `node-addon-api`.

mateodelnorte:
thanks @jasongin. that gets me part of the way there! but I still get syntax errors on node-gyp rebuild: 

https://gist.github.com/mateodelnorte/821335aeb1d4e6a4dfdb594f0386dc97

jasongin:
I published the updated documentation to https://nodejs.github.io/node-addon-api/

PR #60, opened 2017-06-13
Add API for getting object's property names

sampsongao:


napi.h
```diff
@@ -523,6 +523,8 @@ namespace Napi {
       double numberValue ///< Property value
     );
 
+    Napi::Array GetPropertyNames(); ///< Get all property names
```
jasongin: I think there's no need for the `Napi::` prefix here, and also in the implementation. Of course it doesn't hurt, but I'd prefer to keep the code less verbose when possible.

test/object.js
```diff
@@ -82,6 +82,16 @@ function test(binding) {
...
   }
 
+  {
+    const obj = {"one": 1, "two": 2, "three": 3};
+    var arr = binding.object.GetPropertyNames(obj);
+    var expected = ["one", "two", "three"];
+    assert.equal(arr.length, 3);
+    arr.forEach(function(property, i){
+      assert.equal(property, expected[i]);
+    });
```
jasongin: You can simplify the 4 lines above with `assert.deepStrictEqual()`.

jasongin:
Oh, do you need to take a dependency on this right away? If so, bump the version number to 0.3.4, then after merging I can publish the update to npm.

jasongin:
I published 0.3.4 to npm.

Issue #61, opened 2017-06-14
Compile failure after pull request #52


sampsongao:
After the commit 09c7b71274cf3cf6d18a8c9ea15e5ebfab653645
, I see this build error when porting node-sqlite3 on linux.
```
../src/database.cc:24:6:   required from here
/home/sampsongao/node_api/sqlite3/node_modules/node-addon-api/napi-inl.h:51:21: error: void value not ignored as it ought to be
     return callback();
```

Any clue what is the fix here?

jasongin:
What is at `database.cc:24` referenced in the error?

Looking through uses of `WrapCallback()` referenced in the error there, I do see what looks like a problem at `napi-inl.h:2435`. There should be a `return nullptr;` statement there. That could be the cause of that compile error.

sampsongao:
```
    Napi::Function t = DefineClass(env, "Database", {
      InstanceMethod("close", &Database::Close),
      InstanceMethod("exec", &Database::Exec),
      InstanceMethod("wait", &Database::Wait),
      InstanceMethod("loadExtension", &Database::LoadExtension),
      InstanceMethod("serialize", &Database::Serialize),
      InstanceMethod("parallelize", &Database::Parallelize),
      InstanceMethod("configure", &Database::Configure),
      InstanceMethod("interrupt", &Database::Interrupt),
      InstanceAccessor("open", &Database::OpenGetter, nullptr),
    });        //  L. 24
```

sampsongao:
Adding return nullptr after line 2435 works. Thanks!

sampsongao:
Closed #62

PR #62, opened 2017-06-16
Fix the build error on linux

sampsongao:
Addressing issue #61


sampsongao:
Landed https://github.com/nodejs/node-addon-api/commit/244abc346693ad744aabfd73412f1b729960efd1

PR #65, opened 2017-06-22
Add support for VS2013

kfarnung:
VS2013 issues
* No constexpr support
  - Conditionally use constexpr if the compiler supports it
  - Updated tests to use a macro for this purpose
* Requires copy constructor support for base classes of thrown objects
  - Added a protected copy constructor to ObjectReference and
    Reference<T>
* Incorrect overload resolution for std::initializer_list
  - Used explicit types to invoke the correct overload
* No literal support for char16_t
  - Added macro to convert the text properly
* Lambdas are unable to capture function non-pointers
  - Updated the typedef to create a function pointer
* Short lifetime for temporary objects passed to constructors
  - Updated the tests to create the objects and pass them to the
    constructor. This is closer to how they would be used anyway
    
General improvements
* Added console.log output to the test runner to make it more clear
  when and where tests are failing

[Build Status](https://travis-ci.org/kfarnung/node-addon-api)

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -555,13 +555,13 @@ inline std::string String::Utf8Value() const {
 inline std::u16string String::Utf16Value() const {
   size_t length;
   napi_status status = napi_get_value_string_utf16(_env, _value, nullptr, 0, &length);
-  NAPI_THROW_IF_FAILED(_env, status, u"");
+  NAPI_THROW_IF_FAILED(_env, status, NAPI_WIDE_TEXT(""));
 
   std::u16string value;
   value.reserve(length + 1);
   value.resize(length);
   status = napi_get_value_string_utf16(_env, _value, &value[0], value.capacity(), nullptr);
-  NAPI_THROW_IF_FAILED(_env, status, u"");
+  NAPI_THROW_IF_FAILED(_env, status, NAPI_WIDE_TEXT(""));
   return value;
 }
 
@@ -1490,7 +1490,7 @@ inline Error& Error::operator =(Error&& other) {
   return *this;
 }
 
-inline Error::Error(const Error& other) : Error(other.Env(), other.Value()) {
+inline Error::Error(const Error& other) : ObjectReference(other) {
 }
 
 inline Error& Error::operator =(Error& other) {
@@ -1645,6 +1645,22 @@ inline Reference<T>& Reference<T>::operator =(Reference<T>&& other) {
 }
 
 template <typename T>
+inline Reference<T>::Reference(const Reference<T>& other) {
+  _env = other.Env();
+  HandleScope scope(_env);
+
+  napi_value value = other.Value();
+  if (value != nullptr) {
+    // Copying is a limited scenario (currently only used for Error object) and always creates a
+    // strong reference to the given value even if the incoming reference is weak.
+    napi_status status = napi_create_reference(_env, value, 1, &_ref);
+
+    // TODO - Switch to napi_fatal_error() once it exists.
+    assert(status == napi_ok);
+  }
+}
+
+template <typename T>
 inline Reference<T>::operator napi_ref() const {
   return _ref;
 }
@@ -1776,6 +1792,10 @@ inline ObjectReference::ObjectReference(ObjectReference&& other)
 inline ObjectReference& ObjectReference::operator =(ObjectReference&& other) {
   static_cast<Reference<Object>*>(this)->operator=(std::move(other));
   return *this;
+}
+
+inline ObjectReference::ObjectReference(const ObjectReference& other)
+  : Reference<Object>(other) {
 }
 
 inline Napi::Value ObjectReference::Get(const char* utf8name) const {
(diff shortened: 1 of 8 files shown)


kfarnung:
I'm still investigating test failures that I'm seeing in Travis with node-chakracore, but all other legs complete successfully.

napi.h
```diff
@@ -7,6 +7,19 @@
 #include <string>
 #include <vector>
 
+// VS2015 RTM has bugs with constexpr, so require min of VS2015 Update 3 (known good version)
+#if !defined(_MSC_VER) || _MSC_FULL_VER >= 190024210
+#define NAPI_HAS_CONSTEXPR 1
+#endif
+
+// VS2015 RTM has bugs with constexpr, so require min of VS2015 Update 3 (known good version)
```
jasongin: Change this comment to reference the char16_t literal issue.

test/function.cc
```diff
@@ -62,7 +62,7 @@ Value CallWithVector(const CallbackInfo& info) {
 Value CallWithReceiverAndArgs(const CallbackInfo& info) {
    Function func = info[0].As<Function>();
    Value receiver = info[1];
-   return func.Call(receiver, { info[2], info[3], info[4] });
+   return func.Call(receiver, { static_cast<napi_value>(info[2]), static_cast<napi_value>(info[3]), static_cast<napi_value>(info[4]) });
```
jasongin: nit: Wrap lines at 100 cols.
(There is no linter set up yet...)

napi-inl.h
```diff
@@ -1645,6 +1645,21 @@ inline Reference<T>& Reference<T>::operator =(Reference<T>&& other) {
...
+inline Reference<T>::Reference(const Reference<T>& other) {
+  _env = other.Env();
+  HandleScope scope(_env);
+
+  napi_value value = other.Value();
+  if (value != nullptr) {
+    napi_status status = napi_create_reference(_env, value, 1, &_ref);
+
+    // TODO - What's the right way to handle errors in the constructor without
+    //        being able to throw exceptions?
```
jasongin: `napi_fatal_error()`... but it doesn't exist yet. :)

napi-inl.h
```diff
@@ -1645,6 +1645,21 @@ inline Reference<T>& Reference<T>::operator =(Reference<T>&& other) {
 }
 
 template <typename T>
+inline Reference<T>::Reference(const Reference<T>& other) {
+  _env = other.Env();
+  HandleScope scope(_env);
+
+  napi_value value = other.Value();
+  if (value != nullptr) {
+    napi_status status = napi_create_reference(_env, value, 1, &_ref);
```
jasongin: This is always creating a strong reference, even if `other` was a weak reference. I guess it's OK since this should only ever be called by the `Error` class which is always strong. At least add a comment explaining that is the purpose of this constructor.

test/function.cc
kfarnung: Are there specific guidelines for how to wrap lines?  I looked through the existing code and didn't see enough forced wraps to draw solid conclusions.

jasongin: No specific guidelines.

kfarnung:
@jasongin The CI is clean now, but I'm running into a test failure with VS2013 (now that the tests actually run).

I'll pick this up again tomorrow and figure out why the test is failing.

kfarnung:
@jasongin @boingoing I have resolved all build and test issues, is there anything else I should be testing with?

kfarnung:
Everything is passing now locally building in VS2013 and VS2015 and the [CI](https://travis-ci.org/kfarnung/node-addon-api/builds/246406247) is green for all legs.

test/object.cc
```diff
@@ -39,14 +39,26 @@ void DefineProperties(const CallbackInfo& info) {
...
+
     obj.DefineProperties({
-      PropertyDescriptor::Accessor(std::string("readonlyAccessor"), TestGetter),
-      PropertyDescriptor::Accessor(std::string("readwriteAccessor"), TestGetter, TestSetter),
-      PropertyDescriptor::Value(std::string("readonlyValue"), trueValue),
-      PropertyDescriptor::Value(std::string("readwriteValue"), trueValue, napi_writable),
-      PropertyDescriptor::Value(std::string("enumerableValue"), trueValue, napi_enumerable),
-      PropertyDescriptor::Value(std::string("configurableValue"), trueValue, napi_configurable),
-      PropertyDescriptor::Function(std::string("function"), TestFunction),
+      PropertyDescriptor::Accessor(str1, TestGetter),
```
jasongin: This is unfortunate, and seems likely to confuse many people (if they are using VS 2013). The way I wanted this to work for the most common case is you should be able to just pass in a string literal:

    PropertyDescriptor::Accessor("readonlyAccessor", TestGetter),

Did you try that? I'd assume the behavior is the same as the explicit call to `std::string()`, but I'm not sure.

If that doesn't work, then I think it would be best to add overloads of each of those methods that accept a `const char*` instead.

kfarnung: @jasongin there is already an overload for const char *, that's what the other leg of the conditional tests. It just gets chained to for the std::string constructor. Does that resolve your concerns?

jasongin: Oh, sorry I didn't look at the broader context in this file! Yes, that resolves my concerns then.

napi-inl.h
```diff
@@ -2697,11 +2716,11 @@ inline FunctionReference& AsyncWorker::Callback() {
 }
 
 inline void AsyncWorker::OnOK() {
-  _callback.MakeCallback(_receiver.Value(), {});
+  _callback.MakeCallback(_receiver.Value(), {{}});
```
kfarnung: This should have been changed in the latest revision.  This actually had side-effects that I encountered during testing.

kfarnung:
@jasongin As far as I'm concerned this is ready to land, feel free to land it at your convenience.

Issue #63, opened 2017-06-16
ObjectWrap construction flow


sampsongao:
In node-sqlite3, they have code call Ref() on Database in the Database::New. Similar to node-canvas port, I change the Database::New to Database::Database. So the problem is in the ConstructorCallbackWrapper, Ref() is called before the ref is set in the ObjectReference.
```
  T* instance;
  napi_value wrapper;
  details::WrapCallback([&] {
    CallbackInfo callbackInfo(env, info);
    instance = new T(callbackInfo);
    wrapper = callbackInfo.This();   // ==> calling db->Ref() here
    return nullptr;
  });

  napi_ref ref;
  status = napi_wrap(env, wrapper, instance, FinalizeCallback, nullptr, &ref);
  if (status != napi_ok) return nullptr;

  Reference<Object>* instanceRef = instance;
  *instanceRef = Reference<Object>(env, ref);
```

So I guess better way to go is to have Database::Database() for creating napi_ref and then call a Database::New (callbackInfo) to initialize?
```
  T* instance = new T();
  napi_value wrapper;
  details::WrapCallback([&] {
    CallbackInfo callbackInfo(env, info);
    wrapper = callbackInfo.This();
    return nullptr;
  });

  napi_ref ref;
  status = napi_wrap(env, wrapper, instance, FinalizeCallback, nullptr, &ref);
  if (status != napi_ok) return nullptr;

  Reference<Object>* instanceRef = instance;
  *instanceRef = Reference<Object>(env, ref);

  details::WrapCallback([&] {
    CallbackInfo callbackInfo(env, info);
    instance->New(callbackInfo);
    return nullptr;
  });
```

sampsongao:
So the idea here is that the whatever function the original code use to initialize a new instance (in this case is the Database::New) should be the last thing get called in the ConstructoCallbackWrapper. By the time we call it, all preparation should have been done.

sampsongao:
To clarify this problem:
So in [Database::New()](https://github.com/mhdawson/node-sqlite3/blob/master/src/database.cc#L137) it creates a new Baton, which calls [ObjectWrap->Ref()](https://github.com/mhdawson/node-sqlite3/blob/master/src/database.h#L42). However, if we change Database::New to be Database::Database (similar to what is done in Canvas), then in the [ConstructorCallbackWrapper](https://github.com/nodejs/node-addon-api/blob/master/napi-inl.h#L2434), the napi_ref haven't be created by the time Ref() is called.

[...]

PR #67, opened 2017-06-26
Move napi_wrap into ObjectWrap constructor

sampsongao:
For #63

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2274,7 +2274,16 @@ inline PropertyDescriptor::operator const napi_property_descriptor&() const {
 ////////////////////////////////////////////////////////////////////////////////
 
 template <typename T>
-inline ObjectWrap<T>::ObjectWrap() {
+inline ObjectWrap<T>::ObjectWrap(Napi::CallbackInfo callbackInfo) {
+  napi_env env = callbackInfo.Env();
+  napi_value wrapper = callbackInfo.This();
+  napi_status status;
+  napi_ref ref;
+  status = napi_wrap(env, wrapper, this, FinalizeCallback, nullptr, &ref);
+  if (status != napi_ok) return;
+
+  Reference<Object>* instanceRef = this;
+  *instanceRef = Reference<Object>(env, ref);
 }
 
 template<typename T>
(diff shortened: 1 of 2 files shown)


napi-inl.h
```diff
@@ -2242,7 +2242,16 @@ inline PropertyDescriptor::operator const napi_property_descriptor&() const {
 ////////////////////////////////////////////////////////////////////////////////
 
 template <typename T>
-inline ObjectWrap<T>::ObjectWrap() {
+inline ObjectWrap<T>::ObjectWrap(Napi::CallbackInfo callbackInfo) {
+  napi_env env = callbackInfo.Env();
+  napi_value wrapper = callbackInfo.This();
+  napi_status status;
+  napi_ref ref;
+  status = napi_wrap(env, wrapper, this, FinalizeCallback, nullptr, &ref);
```
sampsongao: Just want to make sure that it is safe to use `this` instead of passing an instance pointer `T*`.

jasongin: I think it would be more correct to use `static_cast<T*>(this)`, though I'm not sure it will make a difference.

napi-inl.h
```diff
@@ -2242,7 +2242,16 @@ inline PropertyDescriptor::operator const napi_property_descriptor&() const {
...
+  if (status != napi_ok) return;
```
jasongin: `NAPI_THROW_IF_FAILED(env, status)`

napi-inl.h
sampsongao: The type is void* anyway. Wouldn't static_cast here be redundant?

jasongin: `static_cast<>` can _change the pointer_ for objects that have a vtable. Because `ObjectWrap::Unwrap()` interprets the `void*` as a `T*`, we should be sure to store it as a `T*`.

sampsongao:
I guess I should also update the version number as this is a significant change?

jasongin:
I'll update the version number next time I publish to npm. That will probably be after the current round of PRs all land.

jasongin:
But yes this is a breaking change due to the different `ObjectWrap` constructor. Any modules that use `Napi::ObjectWrap` will need to be updated. The conversion tool also.

Issue #72, opened 2017-07-02
Reference:  copy-initialization miss _suppressDestruct(false)


langxiong:
`Reference` member variable `_suppressDestruct` is only initialized in default constructor.
It should be initialized in other constructor. 
See [code](https://github.com/nodejs/node-addon-api/blob/master/napi-inl.h#L1619)

If `_suppressDestruct` is not initialized to `false`, the destructor may not call `napi_delete_reference`, then the gc is not work as expected.

jasongin:
Fixed by https://github.com/nodejs/node-addon-api/commit/b0f2c6fe59d08f2b0c0bd44a43f3362640b84059

PR #74, opened 2017-07-02
Reference: fix constructor miss some member variables initialize.

langxiong:


langxiong:
#72

napi-inl.h
```diff
@@ -1632,25 +1632,28 @@ inline Reference<T>::~Reference() {
...
   other._env = nullptr;
   other._ref = nullptr;
+  other._suppressDestruct = false;
   return *this;
 }
 
 template <typename T>
-inline Reference<T>::Reference(const Reference<T>& other) {
+inline Reference<T>::Reference(const Reference<T>& other) 
+  : _env(other._env), _ref(nullptr), _suppressDestruct(other._suppressDestruct) {
```
jasongin: I think `_suppressDestruct` should be initialized to `false` in this case. When copying a reference, the newly created reference should be destructed normally by default, even if the original's destruction was suppressed.

napi-inl.h
```diff
@@ -1632,25 +1632,28 @@ inline Reference<T>::~Reference() {
...
   _env = other.Env();
```
jasongin: This line can be deleted now that `_env` is initialized above.

langxiong: This line has been deleted now.

napi-inl.h
langxiong: Yes, I agreed with you.
Actually i don't know what dose `_suppressDestruct ` used for.
I can not find any example code about it.

jasongin: The purpose of that is explained in the header comment:
```C++
    // Call this on a reference that is declared as static data, to prevent its destructor
    // from running at program shutdown time, which would attempt to reset the reference when
    // the environment is no longer valid.
    void SuppressDestruct();
```

PR #70, opened 2017-06-30
Add support for Error::Fatal

kfarnung:
* Added static method `Error::Fatal` to invoke `napi_fatal_error`
* Added `node_internals.cc/h` to shim missing internal functions
* Added a test for `Error::Fatal`

PR #78, opened 2017-07-12
Update to new error helper signatures in n-api

mhdawson:
Update to use new signatures
Update copies of n-api files so they match. I've not done
a direct copy because more work is needed given that the
n-api files now use node internals.  I have updated so that
they are consistent and will build ok with respect to
this specific chaange. I took this approach to minimize
the time the wrapper is broken by the breaking change
on the n-api side.

Also note that we'll need follow on work to expose the ability to add the code
when using the wrapper but I think that is best left to a follow on PR
so that we can limit the time the wrapper is broken.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1452,10 +1452,10 @@ inline Error Error::New(napi_env env) {
         case napi_string_expected:
         case napi_boolean_expected:
         case napi_number_expected:
-          status = napi_create_type_error(env, message, &error);
+          status = napi_create_type_error(env, nullptr, message, &error);
           break;
         default:
-          status = napi_create_error(env, message, &error);
+          status = napi_create_error(env, nullptr,  message, &error);
           break;
         }
         assert(status == napi_ok);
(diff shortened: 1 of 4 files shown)


mhdawson:
Related to https://github.com/nodejs/node/pull/13988

jasongin:
> more work is needed given that the n-api files now use node internals

That work is done already in https://github.com/nodejs/node-addon-api/pull/70. Looks like it's almost ready to merge. Maybe you should wait to rebase on top of that?

kfarnung:
@jasongin I could split out the changes from my PR to land it sooner.

kfarnung:
I opened #79 with my changes split out from #70.

mhdawson:
Landed https://github.com/nodejs/node-addon-api/pull/79 and then rebased this PR and validated with latest version of node that has the error changes.

mhdawson:
@jasongin @kfarnung once I get one approval I'll land

kfarnung:
Yesterday I was seeing a test failure on windows with node 8.1.4, I'll take a look today and see if I can figure out what's going on.

mhdawson:
Landed as 179d13529cca85ef7228ba779bc153cb6925003f

Issue #97, opened 2017-08-02
Following instructions, depending module do not compile on Node 6.11 


mcollina:
```
CXX(target) Release/obj.target/node-api/node_modules/node-addon-api/src/node_api.o
In file included from ../node_modules/node-addon-api/src/node_api.cc:20:
In file included from ../node_modules/node-addon-api/src/node_internals.h:8:
../node_modules/node-addon-api/src/util-inl.h:9:8: error: use of undeclared identifier 'Napi'
inline Napi::String OneByteString(v8::Isolate* isolate,
       ^
../node_modules/node-addon-api/src/util-inl.h:12:10: error: no viable conversion from returned value of type 'MaybeLocal<v8::String>' to function return type 'int'
  return v8::String::NewFromOneByte(isolate,
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../node_modules/node-addon-api/src/util-inl.h:18:8: error: use of undeclared identifier 'Napi'
inline Napi::String OneByteString(v8::Isolate* isolate,
       ^
../node_modules/node-addon-api/src/util-inl.h:21:10: error: no viable conversion from returned value of type 'MaybeLocal<v8::String>' to function return type 'int'
  return v8::String::NewFromOneByte(isolate,
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

(and others)

Launching `npm test`  from this module works, so maybe there is something to be updated in the README about the instructions on how to update `binding.gyp`.

Here is my `binding.gyp`:

```
{
  "targets": [
    {
      "target_name": "native-hdr-histogram",
      "cflags!": [ "-fno-exceptions" ],
      "cflags_cc!": [ "-fno-exceptions" ],
      "sources": [
        "src/hdr_encoding.h",
        "src/hdr_encoding.c",
        "src/hdr_histogram.h",
        "src/hdr_histogram.c",
        "src/hdr_histogram_log.h",
        "src/hdr_histogram_log.c",
        "hdr_histogram_wrap.cc",
        "histogram.cc"
      ],
      "include_dirs": [
        "<!@(node -p \"require('node-addon-api').include\")",
        "src/"
      ],
      "dependencies": [
        "<(module_root_dir)/zlib/zlib.gyp:zlib",
        "<!(node -p \"require('node-addon-api').gyp\")"
      ],
      'xcode_settings': {
        'GCC_ENABLE_CPP_EXCEPTIONS': 'YES',
[...]

PR #112, opened 2017-08-21
Avoid node_modules during conversion

mcollina:
Fixes https://github.com/nodejs/node-addon-api/issues/97


mhdawson:
Landed as 3efce08f8080834af735d5782615c4c2a2ea5f64

PR #113, opened 2017-08-21
Replace napi_is_construct_call with napi_get_new_target

sampsongao:
For https://github.com/nodejs/node/pull/14698

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2026,7 +2026,7 @@ inline bool CallbackInfo::IsConstructCall() const {
   napi_value newTarget;
   napi_status status = napi_get_new_target(_env, _info, &newTarget);
   NAPI_THROW_IF_FAILED(_env, status, false);
-  return (new_target != nullptr);
+  return (newTarget != nullptr);
 }
 
 inline Napi::Env CallbackInfo::Env() const {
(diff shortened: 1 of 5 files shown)


napi-inl.h
```diff
@@ -2015,9 +2015,16 @@ inline CallbackInfo::~CallbackInfo() {
...
+  return Value(_env, newTarget);
+}
+
 inline bool CallbackInfo::IsConstructCall() const {
-  bool isConstructCall;
-  napi_status status = napi_is_construct_call(_env, _info, &isConstructCall);
+  napi_value newTarget;
+  napi_status status = napi_get_new_target(_env, _info, &newTarget);
+  bool isConstructCall = (new_target != nullptr);
   NAPI_THROW_IF_FAILED(_env, status, false);
```
addaleax: Shouldn’t this line be part of `NewTarget()` as well?

By the way, what I had in mind was just replacing `IsConstructCall` with
```c++
inline bool CallbackInfo::IsConstructCall() const {
  return !NewTarget().IsEmpty();
}
```

sampsongao:
Landed 458f5760abb38b2eafb281ebc8716945c57c6aa0

Issue #86, opened 2017-07-20
Object::PropertyLValue<>::operator= chooses wrong overload of Set for const char* argument


mika-fischer:
`Object.Set(key_type, bool)` is called instead of `Object.Set(key_type, const char*)`.

addaleax:
I assume you’re right about this but can you post an example piece of code since you seem to already have that?

mika-fischer:
```
auto result = Napi::Object::New(env);
result["foo"] = "bar";
result["type"] = static_cast<double>(result.Get("foo").Type());
return result;
```

Yields
```
{ foo: true, type: 2 }
```

instead of

```
{ foo: "bar", type: 4 }
```

This is with node 8.2.0, node-addon-api 0.4.0 (can't use master because of the error API changes) and MSVC 2015 Update 3

bnoordhuis:
I bet it's something like this:
```c++
std::string key("foo");
const char* value = "bar";
obj.Set(key, value);
```
There are Set() overloads that take `const char*` or `const std::string&` but not a mix of both.  Since pointers can be coerced to bool, the compiler will pick the overload that takes a bool as the second argument.

mika-fischer:
@bnoordhuis Yes, that's probably it, `PropertyLValue` always uses `string` as key type...

jasongin:
> Does this just mean we need another overload ?

Yes, I think so.

addaleax:
I've suggested what I would consider a proper fix (rather than adding a gazillion more overloads) in #123, could you try that? :)

PR #123, opened 2017-08-28
Refactor property setting

addaleax:
Previously, various utility overloads of `Object::Set()` were provided. This was problematic, because accidental unexpected conversions could happen (e.g. `const char*` to `bool`) and overloads were ambiguous (e.g. `int` could be cast to `double` and `bool` and therefore could not be passed directly).

To address this, pass a template parameter instead, and introduce a helper `Value::From` method that could create a value based on the type of object it receives, and that uses clear precedence rules for determining the kind of JS value to create.

Fixes: https://github.com/nodejs/node-addon-api/issues/86

Also: Note that we can, if we want, extend this facility a lot, e.g. by turning iterable C++ types into arrays or `std::unordered_map`s into objects, and so on. I'd be happy to work on that, but since that would be a bit more extensive, it's probably good to get other people's buy-in first.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -10,6 +10,7 @@
 // Note: Do not include this file directly! Include "napi.h" instead.
 
 #include <cstring>
+#include <type_traits>
 
 namespace Napi {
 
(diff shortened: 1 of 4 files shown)


napi-inl.h
```diff
@@ -605,6 +606,102 @@ inline Symbol::Symbol(napi_env env, napi_value value) : Name(env, value) {
...
+template <typename T>
+struct vf_fallback {
+  static Value From(napi_env env, const T& value) {
+    return Value(env, value);
+  }
+};
+
+template <typename T>
+struct can_make_string {
+  static constexpr bool value =
```
jasongin: There are problems with `constexpr` support on some older compilers. See `NAPI_HAS_CONSTEXPR` elsewhere in this codebase. If this approach doesn't not work at all without `constexpr`, I'm not sure what to do about it.

(Again, we really need a CI system to verify changes work across all supported platforms and versions.)

jasongin:
I really like the idea, I'm just concerned about whether it will work across all platforms/compilers that need to be supported.

napi-inl.h
addaleax: > There are problems with `constexpr` support on some older compilers. See `NAPI_HAS_CONSTEXPR` elsewhere in this codebase. If this approach doesn't not work at all without `constexpr`, I'm not sure what to do about it.

There's definitely a way around this, people did this kind of thing before C++11 :) I am not sure I can do much without access to those older compilers, though, because any attempt at "fixing" this  would be a shot in the dark. I think the obvious approaches would include using `const` instead, using a type instead of a `bool`, or just looking at how `<type_traits>` is implemented by those compilers.

> (Again, we really need a CI system to verify changes work across all supported platforms and versions.)

Yes. :)

kkoopa: If all you want to do is avoid this particular constexpr, it can always be rewritten through old-fashioned template metaprogramming. A generic solution would be to implement a disjunction template for testing the `std::is_convertible`s:
```c++
template<class...> struct disjunction : std::false_type { };
template<class B> struct disjunction<B> : B { };
template<class B, class... Bs> struct disjunction<B, Bs...> :
    std::conditional_t<bool(B::value), B, disjunction<Bs...>>  { };
```

kkoopa: I made a quick implementation that appears to work: kkoopa@54f5a39ee236b60e82b12d4f3c480e4f48fb0386
```diff
-template <typename T>
 -struct can_make_string {
 -  static constexpr bool value =
 -      std::is_convertible<T, const char*>::value ||
 -      std::is_convertible<T, const char16_t*>::value ||
 -      std::is_convertible<T, std::string>::value ||
 -      std::is_convertible<T, std::u16string>::value;
 -};
 +template <typename...> struct disjunction : std::false_type {};
 +template <typename B> struct disjunction<B> : B {};
 +template <typename B, typename... Bs>
 +struct disjunction<B, Bs...>
 +    : std::conditional<bool(B::value), B, disjunction<Bs...>>::type {};
 +
 +template <typename T>
 +struct can_make_string
 +    : disjunction<typename std::is_convertible<T, const char *>::type,
 +                  typename std::is_convertible<T, const char16_t *>::type,
 +                  typename std::is_convertible<T, std::string>::type,
 +                  typename std::is_convertible<T, std::u16string>::type> {};
```

addaleax: I can't push right now, but if anybody wants to include that into this PR, feel free to do so

addaleax:
ping @jasongin @kkoopa any chance this could get reviewed? :)

napi-inl.h
```diff
@@ -605,6 +606,106 @@ inline Symbol::Symbol(napi_env env, napi_value value) : Name(env, value) {
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+// Automagic value creation
+////////////////////////////////////////////////////////////////////////////////
+
+namespace details {
+template <typename T>
+struct vf_number {
+  static Number From(napi_env env, T value) {
```
kkoopa: How about an assert that `value` does not exceed 2^53? This should, however, only be done for integer `T` wider than 53 bits. On the other hand, if this check were added, it might be better off at a lower level.

addaleax: Maybe … I think it’s kind of up to the user to make sure that their values are actually representable in JS?

PR #125, opened 2017-08-31
Fix crash due to ObjectWrap copying CallbackInfo

davedoesdev:
`ObjectWrap` constructor takes `CallbackInfo` by value so `_dynamicArgs` gets released twice. Occurs with >6 args to the `ObjectWrap` constructor.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2284,7 +2284,7 @@ inline PropertyDescriptor::operator const napi_property_descriptor&() const {
 ////////////////////////////////////////////////////////////////////////////////
 
 template <typename T>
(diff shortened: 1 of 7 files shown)


davedoesdev:
I'm happy to reformat. Can you recommend a utility to help?

davedoesdev:
Reformatted by hand.

addaleax:
> Can you recommend a utility to help?

I don’t think we have any tooling, but we should probably look into copying Node core’s C++ linter to this repo

mhdawson:
Landed as ec33379baa358990b37b41df2cac49b8554aec2b

PR #128, opened 2017-09-05
Make TypedArrayOf take both types as template parameters

kkoopa:
This allows a compile-time check for matching array types.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1123,73 +1123,68 @@ inline Napi::ArrayBuffer TypedArray::ArrayBuffer() const {
 // TypedArrayOf<T> class
 ////////////////////////////////////////////////////////////////////////////////
 
-template <typename T>
-inline TypedArrayOf<T> TypedArrayOf<T>::New(napi_env env,
-                                            size_t elementLength,
-                                            napi_typedarray_type type) {
-  Napi::ArrayBuffer arrayBuffer = Napi::ArrayBuffer::New(env, elementLength * sizeof (T));
-  return New(env, elementLength, arrayBuffer, 0, type);
+template <typename U, napi_typedarray_type V>
+inline TypedArrayOf<U, V> TypedArrayOf<U, V>::New(napi_env env,
+                                            size_t elementLength) {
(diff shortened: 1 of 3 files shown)


jasongin:
An earlier version of the code had this design. But I removed the 2nd type parameter in https://github.com/nodejs/node-addon-api/commit/bc683eb1960346e46c8a97f95b5de82f139a88dc to make the `TypedArrayOf` class easier to use.

Mismatching array types generally shouldn't be a problem, because the `napi_typedarray_type` parameter to the constructor will almost always be inferred using `TypedArray::TypedArrayTypeForPrimitiveType<T>()`. Given that, I think a runtime check to ensure the class isn't mis-used is sufficient. Or maybe it could be changed to a `static_assert` on platforms that support `constexpr`.

kkoopa:
Well, it could still be made with only one template type, which would have to be `Uint8Array` etc, to be able to handle `Uint8ClampedArray` as a special case.

Then use it as, say, `TypedArrayOf<Uint8Array>()`. The `napi_typedarray_type` could be automatically inferred in every case.

jasongin:
I don't like `TypedArrayOf<Uint8Array>` very much either. `TypedArrayOf<uint8_t>` just seems much more natural because the type parameter is the element type.

What is the benefit you're after here? Merely preventing someone from doing something like the first line here instead of the second?
```C++
Uint8Array::New(env, N, napi_int32_array);
Uint8Array::New(env, N);
```

That doesn't seem like it's likely to be a big problem, and would be caught at runtime anyway. So the usability tradeoff isn't worth it in my opinion.

DaAitch:
I think impl can be simplified as "V" is only needed for ctor (as far as I could see), so no class template needed for it.

When I use current master impl with:

```cpp
// int8_t with napi_float32_array
Napi::TypedArrayOf<int8_t>::New(info.Env(), 10, napi_float32_array);
```
I get "Error: Unknown failure" to the NAPI entrypoint in JS which is not very helpful. (Why actually "unknown failure", instead of invalid arg?)

While the PR will not compile code with a helpful message.
As a lib maintainer I'd always prefer compile-time errors.

PR #134, opened 2017-09-13
Set conversion from '->' to '.' for object dereference [conversion]

anisha-rohra:
Uncertain if searching for "->Set(" would over-match in some circumstances. Alternative is to search exclusively for "exports->Set(" and "module->Set(", which will under-match in some circumstances.


mhdawson:
Landed as ed6782ab137bd0a2da903dbbb39b6d32ad08b9d8

PR #136, opened 2017-09-14
Fix conversion tool that created two return statements after throwing error [conversion]

anisha-rohra:
The conversion tool didn't match a return statement but added one anyways when converting a ThrowAsJavaScriptException, which often resulted in two return statements.


mhdawson:
Landed as c98fe89b75edd9e7f6176cafbc90195271669580

PR #140, opened 2017-09-19
update for latest n-api changes

mhdawson:
There is a better way to handle the changes for async hooks, this gets it compiling/working but I know its not right @jasongin can you suggest what we should be doing in terms of properly supporting the new async hook parameters.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -155,26 +155,21 @@ struct AccessorCallbackData {
 ////////////////////////////////////////////////////////////////////////////////
 
 #define NODE_API_MODULE(modname, regfunc)                 \
-  void __napi_ ## regfunc(napi_env env,                   \
-                          napi_value exports,             \
-                          napi_value module,              \
-                          void* priv) {                   \
-    Napi::RegisterModule(env, exports, module, regfunc);  \
+  napi_value __napi_ ## regfunc(napi_env env,             \
+                                napi_value exports) {     \
+    return Napi::RegisterModule(env, exports, regfunc);   \
   }                                                       \
   NAPI_MODULE(modname, __napi_ ## regfunc);
 
 // Adapt the NAPI_MODULE registration function:
 //  - Wrap the arguments in NAPI wrappers.
 //  - Catch any NAPI errors and rethrow as JS exceptions.
-inline void RegisterModule(napi_env env,
-                           napi_value exports,
-                           napi_value module,
-                           ModuleRegisterCallback registerCallback) {
-  details::WrapCallback([&] {
-    registerCallback(Napi::Env(env),
-                     Napi::Object(env, exports),
-                     Napi::Object(env, module));
-    return nullptr;
+inline napi_value RegisterModule(napi_env env,
+                                 napi_value exports,
+                                 ModuleRegisterCallback registerCallback) {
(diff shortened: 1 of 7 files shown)


mhdawson:
@gabrielschulhof the other thing of note in this is that I had to add uv.h and node.h to node_internals.h because of refactoring that took place in the node files.  

I'm wondering who we could have built the wrapper without the vm as node_api.cc would have needed uv.h and node.h ? 

Maybe its just late and I'm tired though ?

jasongin:
Following is an outline of the updates required for the async hooks changes:
 - Define a C++ type for the new `napi_async_context` type; its initialization calls `napi_async_init()` and destruction calls `napi_async_delete()`. (Think carefully about how the async context lifetime is managed using C++ constructor/destructor semantics.)
 - Add an async context parameter to all `MakeCallback()` function overloads.
 - Add async resource and resource-name parameters to the `AsyncWorker` constructor and pass those to `napi_create_async_work()`. (Optionally the async resource object can be the same as the `receiver` parameter.)
 - Change `AsyncWorker::OnOK()` and `AsyncWorker::OnError()` callbacks to _not_ use `MakeCallback()`. An ordinary function call (`_callback::Call()`) is now correct.

Since it's not trivial, it's probably better to do as a separate PR.

mhdawson:
@jasongin so are you suggesting that we land this PR as is and then do a follow on PR to cover the steps you have outline ?   That would make sense to me.

jasongin:
Yes.

mhdawson:
Ok then @jasongin need a review on this PR so I can land.

napi-inl.h
```diff
@@ -155,26 +155,21 @@ struct AccessorCallbackData {
...
-    registerCallback(Napi::Env(env),
-                     Napi::Object(env, exports),
-                     Napi::Object(env, module));
-    return nullptr;
+inline napi_value RegisterModule(napi_env env,
+                                 napi_value exports,
+                                 ModuleRegisterCallback registerCallback) {
+  return details::WrapCallback([&] {
+    return napi_value(registerCallback(Napi::Env(env),
+                      Napi::Object(env, exports)));
```
jasongin: nit: The last line here should be indented further to align with the previous argument.

Or, the explicit `napi_value()` constructor can be omitted because that conversion can be implicit.

napi-inl.h
```diff
@@ -2685,8 +2680,14 @@ inline AsyncWorker::AsyncWorker(const Object& receiver, const Function& callback
   : _env(callback.Env()),
     _receiver(Napi::Persistent(receiver)),
     _callback(Napi::Persistent(callback)) {
-  napi_status status = napi_create_async_work(
-    _env, OnExecute, OnWorkComplete, this, &_work);
+
+  napi_value resource_id;
+  napi_status status = napi_create_string_latin1(
+    _env, "generic", -1, &resource_id);
```
jasongin: I don't like this much, but it's fine for now. The follow-up changes will move this resource name string to a parameter of the `AsyncWorker` constructor.

mhdawson:
Will land and address @jasongin comments.

mhdawson:
Landed as 8b667302a6b8d8e3efe29bb4d70efd54499e59ed

PR #141, opened 2017-09-19
Fix Node n-api support check

digitalinfinity:
The current n-api capability check works only if Node's major version is 8. Switches this to check for unsupported versions instead and invert the result of that to determine whether `isNodeApiBuiltin` is `true`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,8 +14,8 @@ var versionArray = process.version
 // We know which version of Node.js first shipped the incarnation of the API
 // available in *this* package. So, if we find that the Node.js version is below
 // that, we indicate that the API is missing from Node.js.
-var isNodeApiBuiltin =
-  (versionArray[0] >= 8 && versionArray[1] >= 4 && versionArray[2] >= 0);
+var isNodeApiBuiltin = 
+  (versionArray[0] > 8 || (versionArray[0] == 8 && versionArray[1] > 5));
 
 // So far it looks like even version 9 will need the flag. We need to adjust
 // this for the version where the flag is dropped whenever that version lands.


digitalinfinity:
The one question I do have with this change is whether I should consider 8.6+ to have n-api support builtin. @jasongin @mhdawson @gabrielschulhof thoughts?

gabrielschulhof:
Absolutely. We're checking not only whether N-API is available, but also
whether it has the exact same binary signature as that provided by
node-addon-api. If it does not, then the node-addon-api-provided symbols
must be used.

jasongin:
@gabrielschulhof  But assuming we stop making breaking changes, any later version should still be binary-compatible.

gabrielschulhof:
That's right. So, up to and not including 8.6.0 we resolve the symbols with
node-addon-api and starting with 8.6.0 we yield to node.

digitalinfinity:
Ok, so the suggestion then is to change this PR to check for 8.6 instead, right? I'll update once either of you confirms.

gabrielschulhof:
Yeah, I think if we just change the 4 to a 6 it'll do the Right Thing™.

digitalinfinity:
You mean the existing code? i.e the following?
```
var isNodeApiBuiltin = 
  (versionArray[0] >= 8 && versionArray[1] >= 4 && versionArray[2] >= 0); 
```

I don't think changing that 4 to 6 works because it's a conjunction- for 9.0, the first check would pass but the second (`versionArray[1] >= 4`) would fail since in this case `versionArray[1]` would be 0. Hence this PR 😄

gabrielschulhof:
I think the simplest expression there would be
```js
var isNodeApiBuiltin = 
    (versionArray[0] > 8 || (versionArray[0] == 8 && versionArray[1] > 5));
```
As it is, it would evaluate true for 8.5.0, which does not include our recent breaking changes.

I *was* a bit short-sighted to not think beyond 8 :)

digitalinfinity:
Updated to that 😄

mhdawson:
Landed as 646e1b68039d3271f58fa713f36ee88fbbcabb37

PR #143, opened 2017-09-25
Allow easy definition of iterators and other methods that are defined by (public) symbols

rolftimmermans:
This PR introduces two API additions that together makes it a lot easier to define iterator methods and other functions that are defined by public symbols.

The first addition is `Napi::Symbol::Public` which returns a public symbol defined by name on `Symbol`. For example: `Napi::Symbol::Public(env, "iterator")`.

The second addition is the addition of `InstanceMethod(Symbol name, ...)` which allows defining an instance method with a symbol key instead of a string key.

It is now possible to do:

```
DefineClass(env, "Test", {
    InstanceMethod(Napi::Symbol::Public(env, "iterator"), &Test::Iterator)
});
```

This defines the method `Test.prototype[Symbol.iterator]()`.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -593,6 +593,10 @@ inline Symbol Symbol::New(napi_env env, napi_value description) {
   return Symbol(env, value);
 }
 
+inline Symbol Symbol::Public(napi_env env, const std::string& name) {
+  return Napi::Env(env).Global().Get("Symbol").As<Object>().Get(name).As<Symbol>();
+}
+
 inline Symbol::Symbol() : Name() {
 }
 
@@ -1657,7 +1661,7 @@ inline Reference<T>& Reference<T>::operator =(Reference<T>&& other) {
 }
 
 template <typename T>
-inline Reference<T>::Reference(const Reference<T>& other) 
+inline Reference<T>::Reference(const Reference<T>& other)
   : _env(other._env), _ref(nullptr), _suppressDestruct(false) {
   HandleScope scope(_env);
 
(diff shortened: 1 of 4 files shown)


napi-inl.h
```diff
@@ -593,6 +593,10 @@ inline Symbol Symbol::New(napi_env env, napi_value description) {
   return Symbol(env, value);
 }
 
+inline Symbol Symbol::Public(napi_env env, const std::string& name) {
+  return Napi::Env(env).Global().Get("Symbol").As<Object>().Get(name).As<Symbol>();
```
jasongin: Is this right? It's getting a property on the global `Symbol` object with the given `name`, which by my reading of the JavaScript spec is undefined (unless `name` happens to be one of the defined static methods on `Symbol`). I'm not super experienced with symbols, but I would have expected this to invoke the `Symbol.for()` function, which creates a new public symbol.

rolftimmermans: This is intentional. The idea is that instead of:

```Napi::Env(env).Global().Get("Symbol").As<Object>().Get("iterator").As<Symbol>();```

you can now do:

```Symbol::Public("iterator");```

The goal of this PR is to simplify adding symbol properties to classes; and one very common use case is defining `Symbol.iterator`, `Symbol.asyncIterator`.

I believe that the equivalent of `Symbol.for()` is `Symbol::New(...)`; so that API is already present.

Comment (jasongin) [reply napi-inl.h]: Oh, thanks. I see now, this is to reference what are commonly known as "well-known symbols".

Reference:
https://tc39.github.io/ecma262/#sec-well-known-symbols
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Well-known_symbols

Based on some searching, the "public symbols" terminology appears to have been used some earlier in the development of the JS symbols concept, but now is not widely used anymore. So should this method be called `Symbol::WellKnown()` instead?

Comment (rolftimmermans) [reply napi-inl.h]: Sure, that makes sense!

File: napi-inl.h [hunk 2]
```diff
@@ -2424,6 +2428,41 @@ inline ClassPropertyDescriptor<T> ObjectWrap<T>::InstanceMethod(
...
+  return desc;
+}
+
+template <typename T>
+inline ClassPropertyDescriptor<T> ObjectWrap<T>::InstanceMethod(
+    Symbol name,
+    InstanceMethodCallback method,
+    napi_property_attributes attributes,
+    void* data) {
+  // TODO: Delete when the class is destroyed
```
mhdawson: Wondering about the TODO's here and in the other place.  Is there some reason we can't add the deletion as part of this PR ?

rolftimmermans: Defining a class just creates a constructor function, I'm not sure how can we attach a cleanup action when it is garbage collected?

An alternative would be to use `data` for a pointer to the method callback and disallow an additional `void* data` in the C++ API; although that still wouldn't solve the case for the getter/setter (two pointers instead of one).

mhdawson: I see this is not specific to your change, as there were already other instances.

mhdawson:
Landed as 4173b09ac743f5a80646698a4f134a597e80b8e6

PR #148, opened 2017-09-29
Fix segmentation fault when using setters or getters that are not defined

rolftimmermans:
Currently the following scenario aborts with a segmentation fault:

```
exports.Set("Text", DefineClass(env, "Test", {
    InstanceAccessor("readonly_prop", &Test::GetReadonlyProp, nullptr),
}));
```

```
new Test().readonly_prop = 1;
```

This PR fixes that scenario by checking for a null getter/setter and setting the property descriptor accordingly. Includes a test case for readonly properties.

I had to modify the test runner to correctly report SIGSEGV in order for it to actually fail the tests.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -28,5 +28,11 @@ if (typeof global.gc === 'function') {
   const child = require('./napi_child').spawnSync(process.argv[0], [ '--expose-gc', __filename ], {
     stdio: 'inherit',
   });
-  process.exitCode = child.status;
+
+  if (child.signal) {
+    console.log(`Tests aborted with ${child.signal}`);
+    process.exitCode = 1;
+  } else {
+    process.exitCode = child.status;
+  }
 }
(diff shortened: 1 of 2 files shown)


mhdawson:
Sorry I've not had a chance to check this out yet.  I'm out at Node Interactive this week but will try to get to it when I have spare time.

rolftimmermans:
I fixed the conflicts, added a write only test and fixed a segfault that was caused by a missing `SuppressDestruct`. 

If the other PR is merged we can still update the segfault logic and fix of course.

mhdawson:
Ok, my plan is to land the other PR, ask you to rebase and then I can go ahead and land this one.  Sorry for taking so long to review this one....

mhdawson:
Ok other PR has landed, if you can rebase that would be great.

rolftimmermans:
👍 done

mhdawson:
Landed as 22be6092d84ad78efd5aac84620db0545e88ad6c

PR #151, opened 2017-10-02
Add API for deletion of object properties

rolftimmermans:
An API that uses `napi_delete_property` was missing.

This adds a few overloaded `Object::Delete()` methods that remove a given property.

### DIFF
--- a/napi.h
+++ b/napi.h
@@ -491,6 +491,26 @@ namespace Napi {
       double numberValue           ///< Property value
     );
 
+    /// Delete property.
+    bool Delete(
+      napi_value key ///< Property key primitive
+    );
+
+    /// Delete property.
+    bool Delete(
+      Value key ///< Property key
+    );
+
+    /// Delete property.
+    bool Delete(
+      const char* utf8name ///< UTF-8 encoded null-terminated property name
+    );
+
+    /// Delete property.
+    bool Delete(
+      const std::string& utf8name ///< UTF-8 encoded property name
+    );
+
     /// Checks whether an indexed property is present.
     bool Has(
       uint32_t index ///< Property / element index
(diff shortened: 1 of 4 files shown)


napi.h
```diff
@@ -491,6 +491,26 @@ namespace Napi {
...
 
+    /// Delete property.
+    bool Delete(
+      napi_value key ///< Property key primitive
+    );
+
+    /// Delete property.
+    bool Delete(
+      Value key ///< Property key
+    );
```
addaleax: I don’t think we need this key since `Value` has an `operator napi_value` member, right?

rolftimmermans: Indeed I guess so; I just copied the definitions from `Object::Has`. Should I remove all of them?

jasongin: I think I had to add the overloads that take `Value` in order to avoid compiler errors in some situations about ambiguous calls to overloaded methods.

addaleax: @jasongin I think it should be safe to remove those – if you’re talking about #86, that’s happening because there were two multiple arguments that were used for overloading …

@rolftimmermans I’m not sure, but it certainly doesn’t have to happen in this PR :)

addaleax: Oh, wait, maybe that is about `String`s that also have an `operator std::string`? That would make sense but we should have tests for that :)

rolftimmermans:
@jasongin Done – I couldn't find any tests though, is that correct?

jasongin:
@rolftimmermans Test coverage is not great right now but you are welcome to contribute to improving it!

mhdawson:
@rolftimmermans the tests are in https://github.com/nodejs/node-addon-api/tree/master/test and you can run them with `npm test`.

At some point we should require tests for any new APIs' but as @jasongin mentions we don't have great coverage yet so at this point it would be nice but not required.

mhdawson:
Landed as b1c4cb940c1afea104717afb8d276f30a9b1b471

Issue #98, opened 2017-08-02
convert.js does not add node-addon-api to the dependencies in binding.gyp


mcollina:
As titled, after launching `covert.js`  the updated `binding.gyp` is not trying to build this specific module, resulting in missing symbols at runtime (in Node 6.11).

mcollina:
I'm trying to see how do add this. Is it fine if I parse `binding.gyp`? I also need to add this module within the dependencies in `NAPI_CPP_EXCEPTIONS`.

sampsongao:
Can you post an example of the output binding.gyp?

mcollina:
Here it is:

```
{
  "targets": [
    {
      "target_name": "native-hdr-histogram",
      "cflags!": [ "-fno-exceptions" ],
      "cflags_cc!": [ "-fno-exceptions" ],
      "sources": [
        "src/hdr_encoding.h",
        "src/hdr_encoding.c",
        "src/hdr_histogram.h",
        "src/hdr_histogram.c",
        "src/hdr_histogram_log.h",
        "src/hdr_histogram_log.c",
        "hdr_histogram_wrap.cc",
        "histogram.cc"
      ],
      "dependencies": [
        "<(module_root_dir)/zlib/zlib.gyp:zlib"
      ],
      "include_dirs": [
        "<!@(node -p \"require('node-addon-api').include\")",
        "src/"
      ]
    }
  ]
}
```

What it's missing is the line in `dependencies`, like so:

```
{
  "targets": [
    {
      "target_name": "native-hdr-histogram",
      "cflags!": [ "-fno-exceptions" ],
      "cflags_cc!": [ "-fno-exceptions" ],
      "sources": [
        "src/hdr_encoding.h",
        "src/hdr_encoding.c",
        "src/hdr_histogram.h",
        "src/hdr_histogram.c",
        "src/hdr_histogram_log.h",
        "src/hdr_histogram_log.c",
        "hdr_histogram_wrap.cc",
        "histogram.cc"
      ],
      "dependencies": [
        "<(module_root_dir)/zlib/zlib.gyp:zlib",
        "<!(node -p \"require('node-addon-api').gyp\")"
      ],
      "include_dirs": [
        "<!@(node -p \"require('node-addon-api').include\")",
        "src/"
      ]
    }
  ]
}
```

mhdawson:
@sampsongao have you had a chance to look at this ?

sampsongao:
[...]

PR #158, opened 2017-10-22
Add node-addon-api into dependencies

sampsongao:
Addressing https://github.com/nodejs/node-addon-api/issues/98 where node-addon-api may not be added to dependencies if the module initially is not depending on nan


mhdawson:
Landed as a87509329b4bceb1e41a37e36acd7ba5697165ef

PR #159, opened 2017-10-23
Fix a crash when setter is nullptr in InstanceAccessor().

romandev:
In the current implementation, if we want to make a readonly property,
we should pass empty setter that is actual function but has no body
as follows:
```c++
  void HelloSetter(const Npai::CallbackInfo& info,
                   const Npai::Value& value) {
    // This is an empty setter to make a readonly property.
    // Do nothing
  }

  InstanceAccessor("hello_property", &HelloGetter, &HelloSetter);
```

If we allows taking `nullptr` as the argument, we can simplify user's
code more as follows:
```c++
  InstanceAccessor("hello_property", &HelloGetter, nullptr);
```

Currently, above code makes a crash. So, this patch makes
InstanceAccessor() take `nullptr` as the argument and then add some
tests for them.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2562,8 +2562,8 @@ inline ClassPropertyDescriptor<T> ObjectWrap<T>::InstanceAccessor(
 
   napi_property_descriptor desc = {};
   desc.utf8name = utf8name;
-  desc.getter = T::InstanceGetterCallbackWrapper;
-  desc.setter = T::InstanceSetterCallbackWrapper;
+  desc.getter = getter != nullptr ? T::InstanceGetterCallbackWrapper : nullptr;
+  desc.setter = setter != nullptr ? T::InstanceSetterCallbackWrapper : nullptr;
   desc.data = callbackData;
   desc.attributes = attributes;
   return desc;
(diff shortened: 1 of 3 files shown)


romandev:
@jasongin PTAL

napi-inl.h
```diff
@@ -2523,8 +2523,8 @@ inline ClassPropertyDescriptor<T> ObjectWrap<T>::InstanceAccessor(
 
   napi_property_descriptor desc = {};
   desc.utf8name = utf8name;
-  desc.getter = T::InstanceGetterCallbackWrapper;
-  desc.setter = T::InstanceSetterCallbackWrapper;
+  desc.getter = getter ? T::InstanceGetterCallbackWrapper : nullptr;
+  desc.setter = setter ? T::InstanceSetterCallbackWrapper : nullptr;
```
addaleax: Just fyi, I think we have a slight preference for using explicit nullptr comparisons (i.e. `getter != nullptr ? …`). You can update the PR if you want but this should be fine either way.

romandev: Thank you for review.
I addressed your comment :)

mhdawson:
Landed as d3f62e32746486f06a05f54d44ab3d2e55e78e87

PR #165, opened 2017-10-26
Basic data types documentation for nodejs/abi-stable-node#280.

jschlight:
This is my initial hack at documentation for Basic Types. I'm open to all suggestions and criticisms to improve this document. Please let me know how it can be improved with additional content or format changes.


doc/N-API Basic Types.md
```diff
@@ -0,0 +1,373 @@
...
+Returns an `Error` object representing the pending exception, if any.
+
+## Value
+
+A Value is the C++ manifestation of a JavaScript value.
+
+Value is a the base class upon which other JavaScipt values like Number, Boolean, String, and Object are based. 
+
+```
+Value();
```
mhdawson: For each of the methods I would add a heading. In this example:

### Default Constructor

and we might want to add the `Added in: ` which is used in the standard Node doc as well.

We may want to include the Returns line, if if it is just `Returns void` (or whatever we think is the best way to say returns nothing)

See https://nodejs.org/dist/latest-v8.x/docs/api/n-api.html#n_api_napi_create_reference in the N-API doc as the example I looked at to come up with these suggestions.

doc/N-API Basic Types.md
```diff
@@ -0,0 +1,373 @@
+# N-API Documentation
```
mhdawson: A general thought is that some intro at the top that introduces the basic types and how they are used might be useful before the section with the details.  We can add that later though so what you have here would be good to land without that.

mhdawson:
Overall a good start just a few suggestions.   I also think we can incrementally land changes so if you like my suggestions we could integrate those (and any others made in the mean time), land what you have so far and then continue to add content in follow on PRs.

doc/N-API Basic Types.md
mhdawson: We might want to put a big "WORK IN PROGRESS, NOT YET COMPLETE" at the top just to set peoples expectations.

jschlight:
I've split the earlier single large md file into four separate files that fit into the new documentation scaffolding structure. I've implemented Michael's suggestions with the exception of the `Added in:` marker which will require further research.

doc/callbackinfo.md
```diff
@@ -0,0 +1,99 @@
...
+
+Returns a `bool` indicating if the CallbackInfo is a constructor call.
+
+### Length
+
+```cpp
+size_t Length() const;
+```
+
+Returns the number of arguemnts passed in the Callabckinfo object.
```
mhdawson: Callbackinfo -> CallbackInfo  (capital I)

doc/callbackinfo.md
```diff
@@ -0,0 +1,99 @@
...
+
+Returns the `Env` object in which the request is being made.
+
+### NewTarget
+
+```cpp
+Value NewTarget() const;
+```
+
+Returns the `new.target` value of the constructor call. If the CallbackInfo is not a constructor call, a call to `IsEmpty()` on the returned value returns true.
```
mhdawson: This reads a bit strangely (to my ears).  Maybe:

If the function that was invoked (and for which the CallbackInfo was passed) is not a constructor call....

There is also similar wording in a few other places.

doc/callbackinfo.md
```diff
@@ -0,0 +1,99 @@
...
+
+### This
+
+```cpp
+Value This() const;
+```
+
+Returns the JavaScript `this` value for the call
+
+### Data
```
mhdawson: Would be good to provide a bit more details on what/how Data is used but that can be added in a follow on PR.

doc/env.md
```diff
@@ -0,0 +1,65 @@
+**WORK IN PROGRESS, NOT YET COMPLETE**
+
+# Env
+
+The opaque data structure containing the environment in which the request is being run.
+
+The Env object is usually created and passed by the calling system.
```
mhdawson: Would it be better to say 

"is usually created and passed by the Node.js runtime or node-addon-api infrastructure"

The same comment also applies other places were we say "calling system"

doc/external.md
```diff
@@ -0,0 +1,68 @@
...
+- `[in] data`: The arbitrary C++ data to be held by the External object.
+
+Returns the created `External<T>` object.
+
+### New
+
+```cpp
+template <typename T>
+static External New(napi_env env,
+			T* data,
```
mhdawson: The parameters don't line up here.

doc/external.md
```diff
@@ -0,0 +1,68 @@
...
+- `[in] finalizeCallback`: A function called when the External object is released by the garbage collector accepting a T* and returning void.
+
+Returns the created `External<T>` object.
+
+### New
+
+```cpp
+template <typename T>
+static External New(napi_env env,
+			T* data,
```
mhdawson: parameters don't line up here

doc/external.md
```diff
@@ -0,0 +1,68 @@
...
+### Data
+
+```cpp
+T* Data() const;
+```
+
+Returns a pointer to the arbitrary C++ data held by the External object.
+
+### Not documented here
+
```
mhdawson: Is this just because the doc will be added in future PR ?  Or do you have some questions that need to be answered before you can add the doc for these ?

doc/callbackinfo.md
jschlight: I'm not clear how Data is used. I'm assuming that it's a place the C++ programmer can store general information with the C++ object, but I'm not sure. If so, I can add that to the documentation.

doc/external.md
jschlight: I wanted the reviewers to know that they're not documented. I'm pretty sure there's no need to document them. If so, I'll remove the note.

jschlight:
Michael, thank you for your comments. I've implemented your recommendations, have pushed updated files, and added a couple of comments above.

doc/external.md
mhdawson: I'm thinking best to remove.  It confused me so it might confuse others as well.

doc/callbackinfo.md
mhdawson: I think we can add the info as a follow on PR we just need to record somewhere that we should come back and do that.  We should generate a test that uses it which would provide an example to base the doc on.

doc/callbackinfo.md
mhdawson: Just noticed arguemnts -> arguments

mhdawson: arguemnts -> arguments

jschlight:
Thanks, Michael. I've made another push to my branch.

mhdawson:
Landed as 7ecab460df01413f523fb7216d4b03db1fd2b8c9

Issue #164, opened 2017-10-25
Code cleanup -warnings


mhdawson:
There are a bunch of warnings when I run the tests on linux with `npm test`

Would be great to clean these up:

```
In file included from /home/mhdawson/newpull/land/node-addon-api/napi.h:1442:0,
                 from ../objectwrap.cc:1:
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h: In instantiation of âstatic Napi::ClassPropertyDescriptor<T> Napi::ObjectWrap<T>::InstanceMethod(const char*, Napi::ObjectWrap<T>::InstanceMethodCallback, napi_property_attributes, void*) [with T = TestIter; Napi::ObjectWrap<T>::InstanceMethodCallback = Napi::Value (TestIter::*)(const Napi::CallbackInfo&)]â:
../objectwrap.cc:15:45:   required from here
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h:2509:36: warning: missing initializer for member ânapi_property_descriptor::utf8nameâ [-Wmissing-field-initializers]
   napi_property_descriptor desc = {};
                                    ^
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h:2509:36: warning: missing initializer for member ânapi_property_descriptor::nam â [-Wmissing-field-initializers]
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h:2509:36: warning: missing initializer for member ânapi_property_descriptor::methodâ [-Wmissing-field-initializers]
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h:2509:36: warning: missing initializer for member ânapi_property_descriptor::getterâ [-Wmissing-field-initializers]
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h:2509:36: warning: missing initializer for member ânapi_property_descriptor::setterâ [-Wmissing-field-initializers]
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h:2509:36: warning: missing initializer for member ânapi_property_descriptor::valueâ [-Wmissing-field-initializers]
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h:2509:36: warning: missing initializer for member ânapi_property_descriptor::attributesâ [-Wmissing-field-initializers]
/home/mhdawson/newpull/land/node-addon-api/napi-inl.h:2509:36: warning: missing initializer for member ânapi_property_descriptor::dat â [-Wmissing-field-initializers]
[...]

PR #166, opened 2017-10-27
Fix warnings when running npm test

romandev:
The warnings occur due to a bug in gcc(<5.1.1)[1]. To fix this issue,
we can use constructor initializer instead of empty braces initializer.

This change is fixing #164 issue.

[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61489


romandev:
@mhdawson PTAL (this idea comes from https://github.com/sass/libsass/pull/1542)

mhdawson:
Landed as 51f00b1007210939b1b8e458002f53335ebcbc5f

PR #185, opened 2017-11-15
Number doc

corymickelson:
First pass, tried to add examples I thought may be valuable to someone, not sure if they belong here or would make more sense somewhere else. If this is ok I will do the same for the other docs,

### DIFF
--- a/doc/number.md
+++ b/doc/number.md
@@ -1,0 +1,53 @@
+# Number
+
+A Javascript number value.
+## Methods
+
+<b>Constructors</b>
+</br>
+Create a new Javascript Number from an int, float or double.
+</br>
+<i>Parameters</i>: [Env](),  ```double```
+</br>
+<i>Returns</i>:  ```Number``` 
+```cpp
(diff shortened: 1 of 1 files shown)


doc/number.md
```diff
@@ -0,0 +1,53 @@
+# Number
```
mhdawson: The doc should use markdown formatting as opposed to html formatting to be consistent with existing Node.js docs.

doc/number.md
```diff
@@ -0,0 +1,53 @@
...
+
+A Javascript number value.
+## Methods
+
+<b>Constructors</b>
+</br>
+Create a new Javascript Number from an int, float or double.
+</br>
+<i>Parameters</i>: [Env](),  ```double```
```
mhdawson: I think we agreed formatting would be similar to the existing Node.js docs which are along these lines in terms of how the parameters are defined: https://github.com/jschlight/node-addon-api/blob/ea4e5bf988ecb4a2239f93622d52c04605093a06/doc/callbackinfo.md#constructor

and https://github.com/nodejs/node-addon-api/pull/165/files

corymickelson:
ok will update

PR #186, opened 2017-11-17
Introduce new build targets for debugging purpose

romandev:
In this project, we can use the following command to test examples.
`$ npm test`

It might be very inefficient, especially, if the number of files
increases. So, this patch introduces new build targets for debugging
purpose as follows:
```
$ npm run-script dev                # Build with --debug option
$ npm run-script dev:incremental    # Incremental dev build
```

This idea comes from @DaAitch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,20 @@ npm install
 npm test --disable-deprecated
 ```
 
+### **Debug**
+
+To run the **node-addon-api** tests with `--debug` option:
+
+```
+npm run-script dev
+```
+
+If you want faster build, you might use the following option:
+
+```
+npm run-script dev:incremental
+```
+
 Take a look and get inspired by our **[test suite](https://github.com/nodejs/node-addon-api/tree/master/test)**
 
 <a name="resources"></a>
--- a/package.json
+++ b/package.json
@@ -55,6 +55,10 @@
   "scripts": {
     "pretest": "node-gyp rebuild -C test",
     "test": "node test",
+    "predev": "node-gyp rebuild -C test --debug",
+    "dev": "node test",
+    "predev:incremental": "node-gyp configure build -C test --debug",
+    "dev:incremental": "node test",
     "doc": "doxygen doc/Doxyfile"
   },
   "version": "1.6.2"


romandev:
@mhdawson, and others PTAL

I'm not sure if this patch is useful to you. If you want to always make a clean full build for use in CI tool, this patch might not be useful. But if someone like me uses the command in local environment, doing clean full build every time might be very inefficient. WDYT?

mhdawson:
So far the build time is not a big deal, but that does not mean it won't be later on.  We do want full builds in the CI.  How about adding a new target which would be test-incremental that could be used when people want an incremental instead of full build ?

DaAitch:
Some days before I worked on a PR and I also find out that `npm test` seems not the right way to go for dev because of the rebuild, so I `cd test; node-gyp configure build; node mytest.js` but this worked stable only for release. with `node-gyp configure --debug` sometimes the incremental builds getting me segfaults in the test. Although this PR is not merged or closed within the last 7months I think it's worth to improve the build.

gabrielschulhof:
I personally use ccache to drastically reduce build time. It essentially does an incremental build behind node-gyp's back.

gabrielschulhof:
@DaAitch I also noticed that, when working on a test, it's no fun watching all of them rebuild. A solution outside of a change to the build system is to use ccache, however, that requires setting up, and I'm not sure how well it works on platforms outside of Linux.

mhdawson:
@gabrielschulhof "configure build === rebuild" would go into the CI job, right?

gabrielschulhof:
@mhdawson do you mean that the build should fail if it turns out that configure build !== rebuild? If not, I don't think we need to make any changes to the CI. I believe that as long as the CI runs `npm install && npm test` and the tests pass the changes under test should be considered validated by a CI run.

mhdawson:
@gabrielschulhof what I was asking is if by adding a command to the CI run we could ensure it does a full rebuild (even without a clean checkout which I don't think jenkins does by default) instead of an incremental rebuild then I'd be comfortable.

gabrielschulhof:
@mhdawson can we simply `rm -rf test/build`? That should ensure a full rebuild.

DaAitch:
Wouldn't it just be easier to have more build commands, then it's also documented.

It would also be nice to test "Debug" with `--debug` because it has tests that are ignored with "Release" and compilation should be faster. On CI I think also the debug tests are needed as there are

Something like this:
```json
{
  "scripts": {
    "test": "node test",

    "test:dev": "node test",
    "pretest:dev": "node-gyp rebuild -C test --debug",

    "test:dev:incremental": "node test",
    "pretest:dev:incremental": "node-gyp configure build -C test --debug",

    "clean": "node-gyp clean -C test",
    "doc": "doxygen doc/Doxyfile"
  }
}
```

CI runs
- `npm run test:dev` rebuild with debug tests
- `npm run test` rebuild release

Dev runs:
- `npm run test:dev:incremental` for dev
- ~~`npm run test:incremental`~~ I think no incremental release build is needed

mhdawson:
+1 for adding additional targets.

mhdawson:
Landed as fcfc612728fad0533aaa63f5aaad25bae712eeb3

Issue #109, opened 2017-08-17
Make Buffer extend Uint8Array?


TimothyGu:
It's this way in JS, so might be appropriate to make the C++ class extend Uint8Array as well.

PR #188, opened 2017-11-20
Make Buffer extend Uint8Array instead of directly extending Object

sampsongao:
For issue: https://github.com/nodejs/node-addon-api/issues/109


mhdawson:
@sampsongao can you confirm that you don't think this change will require updates to any existing code using node-addon-api ?

sampsongao:
No change on modules is needed.

mhdawson:
@sampsongao thanks landing.

mhdawson:
Landed as 74de0cb69f0af845195e242fe4dd5e4d12eb53c2

PR #190, opened 2017-11-21
doc: minor fixes to README.md

digitalinfinity:


README.md
```diff
@@ -5,13 +5,12 @@ Node.js API**  also known as **N-API**, providing C++ object model and exception
 handling semantics with low overhead. It guarantees backward compatibility with
 use with older versions of Node.js that do not have N-API built-in.
```
aruneshchandra: It guarantees backward compatibility when  used with older versions of Node.js that do not have N-API built-in.

aruneshchandra: How about rearranging the text like this:

This package contains **header-only C++ wrapper classes** for the **ABI-stable Node.js API** also known as **N-API**, providing C++ object model and exception handling semantics with low overhead. It guarantees backward compatibility when  used with older versions of Node.js that do not have N-API built-in.
N-API is an API for building native addons. It is independent from the underlying JavaScript runtime (e.g. V8 or ChakraCore) and is maintained as part of Node.js itself. This API will be Application Binary Interface (ABI) stable across versions and flavors of Node.js. It is intended to insulate native addons from changes in the underlying JavaScript engine and allow modules compiled for one version to run on later versions of Node.js without recompilation. N-API guarantees the API and ABI compatibility across different versions of Node.js. So if you switched to a different version of Node.js, you would not need to reinstall or recompile the native addon.

README.md
```diff
@@ -5,13 +5,12 @@ Node.js API**  also known as **N-API**, providing C++ object model and exception
...
-version of Node.js. So if you switch to a
-different version of Node.js you must not reinstall and maybe recompile the
-Addon.
+N-API guarantees the **API** and **ABI** compatibility across different
+versions of Node.js. So if you switched to a different version of Node.js,
+you would not need to reinstall or recompile the Addon.
 
 N-API is an API for building native Addons. It is independent from the underlying
-JavaScript runtime (ex V8) and is maintained as part of Node.js itself. This API
+JavaScript runtime (e.g. v8) and is maintained as part of Node.js itself. This API
```
mhdawson: I think this should be V8 as I believe I saw a change somewhere else to make updates to have it be consistently V8 as opposed to v8.

README.md
```diff
@@ -29,11 +28,11 @@ values. Concepts and operations generally map to ideas specified in the
 - **[Contributors](#contributors)**
 - **[License](#license)**
 
-## **Current version: 1.0.0**
+## **Current version: 1.1.0**
```
mhdawson: Thanks for catching this.

mhdawson:
@digitalinfinity just wondering if you are going to update.

digitalinfinity:
Ah completely forgot I had this PR open- I'll update today

digitalinfinity:
I keep forgetting about this PR- updated it with Arunesh's proposed text

digitalinfinity:
Thanks- landed in 5441b3c

PR #182, opened 2017-11-06
Added working progress message

NickNaso:
Added a working progress message to empty documentation files.

PR #192, opened 2017-11-23
Start documentation

NickNaso:
Start work on documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,11 @@ values. Concepts and operations generally map to ideas specified in the
 - **[Examples](#examples)**
 - **[Tests](#tests)**
 - **[More resource and info about native Addons](#resources)**
+- **[Code of Conduct](CODE_OF_CONDUCT.md)**
 - **[Contributors](#contributors)**
 - **[License](#license)**
 
-## **Current version: 1.0.0**
+## **Current version: 1.1.0**
 
 (See [CHANHELOG.md](CHANGELOG.md) for complete Changelog)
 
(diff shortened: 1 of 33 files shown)


mhdawson:
Landed as 5543c01f2d7739beb2ac9ae817d1c55e0c170e61

addaleax:
Uh … this was merged *fast*.

@NickNaso How is this documentation going to differ from https://nodejs.github.io/node-addon-api/ ?

NickNaso:
Hi @addaleax,
we are writing the documentation through markdown. We started to organize the **readme** and to split the documentation in different sections. It's only the start. At the end of writing we will use the markdown file to generate a documentation web site to publish on https://nodejs.github.io/node-addon-api/. 
I really want go aver the normal documentation reference and explain in a more expressive way how to use some API provided by the **node-addon-api**
I hope that you share the ideas

addaleax:
> I really want go aver the normal documentation reference and explain in a more expressive way how to use some API provided by the node-addon-api

I am very much a fan of that! What I’m asking is why we’re not integrating that with the *existing* API documentation, which is rather complete and pretty useful imo?

NickNaso:
At the end we need combining both maybe like proposed here:
https://github.com/nodejs/abi-stable-node/issues/280#issuecomment-337331871
Personally I'm thinking on how to do that even if I haven't an idea on how to do that.

mhdawson:
@addaleax the main reason in my mind that we might not integrate with the existing API doc (and why I've not thought of doing that) is that node-addon-api is not "officially" part of Node.js.  It's a `helper` modules that happens to be provided by the Node.js community.  

Having said that, I'm not opposed to integrating if we think that's the right thing to do.

In terms of this PR landing fast, it was opened to replace https://github.com/nodejs/node-addon-api/pull/182 which had been open for quite a while.  We could just not get it into a form that we could land.  Of course we missed adding that context to this PR which would have been good.

addaleax:
@mhdawson Just to make sure we’re talking about the same thing: When I said “existing” documentation, I was referring to the linked https://nodejs.github.io/node-addon-api/ (i.e. documenation for this very module), not the docs for the C interface that are part of Node.

PR #195, opened 2017-11-29
Write tests for Number class

romandev:
The new tests cover a part of the basic type conversion from JavaScript
type to native type.

### DIFF
--- a/test/basic_types/number.cc
+++ b/test/basic_types/number.cc
@@ -1,0 +1,57 @@
+#include <cfloat>
+
+#include "napi.h"
+
+using namespace Napi;
+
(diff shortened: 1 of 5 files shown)


romandev:
@mhdawson PTAL

romandev:
@mhdawson,
It's enough short, but I reduced the count more in my latest patch.
The average loop count will be 400.

mhdawson:
@romandev thanks, will plan to land when I have some free time.

mhdawson:
Landed as 8e31e242098c61db4bf3e9a0dc801b5e51dd7fc8

Issue #196, opened 2017-11-29
Add support for DataView feature


romandev:
`DataView` objects provide an array-like view over an underlying data buffer, but one which allows items of different size and type in the ArrayBuffer.

So, we should provide wrapper implementation for the following APIs.

- napi_is_dataview() [1]
- napi_create_dataview() [2]
- napi_get_dataview_info() [3]

[1] https://nodejs.org/dist/latest-v8.x/docs/api/n-api.html#n_api_napi_is_dataview
[2] https://nodejs.org/dist/latest-v8.x/docs/api/n-api.html#n_api_napi_create_dataview
[3] https://nodejs.org/dist/latest-v8.x/docs/api/n-api.html#n_api_napi_get_dataview_info

romandev:
@mhdawson Can I start to work on this? or If we need discussions or agreements, where can I see the process for adding these new APIs?

mhdawson:
Sounds good.  We should have coverage for all of the things you can do in the core N-API.

romandev:
Okay, I'll take this :)

NickNaso:
Hi @romandev,
what do you think? Could be possible close the issue?

NickNaso:
@romandev 
I think that now that *[https://github.com/nodejs/node-addon-api/pull/331](https://github.com/nodejs/node-addon-api/pull/331)** is landed we can close the issue.
Good work!

PR #202, opened 2017-12-11
Implement Value.isDataView()

romandev:
The method is to check if the given value is a data view object. This ia
an initial implementation of DataView feature(#196).

This change also adds the NAPI_DATA_VIEW_FEATURE flag to expose only to
test modules until all features are implemented.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -347,6 +347,19 @@ inline bool Value::IsPromise() const {
   return result;
 }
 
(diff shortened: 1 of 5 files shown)


mhdawson:
Using the define to make less visible during incremental addition of features seems reasonable to me. 

Will leave for another day to allow 48 hours comment period before landing.

mhdawson:
Landed as b47cce85867c4770369bb12cd7371ea0580d0de7

PR #205, opened 2017-12-19
Implement an initial implementation of DataView class

romandev:
This is an initial implementation of DataView class. This change
includes the following things:
  - DataView::New() methods
  - Getters for DataView (ArrayBuffer(), ByteOffset(), ByteLength())
  - Tests for them

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1171,6 +1171,88 @@ inline void ArrayBuffer::EnsureInfo() const {
   }
 }
 
+#if NAPI_DATA_VIEW_FEATURE
+////////////////////////////////////////////////////////////////////////////////
+// DataView class
+////////////////////////////////////////////////////////////////////////////////
+inline DataView DataView::New(napi_env env,
+                              Napi::ArrayBuffer arrayBuffer) {
+  return New(env, arrayBuffer, 0, arrayBuffer.ByteLength());
+}
+
+inline DataView DataView::New(napi_env env,
+                              Napi::ArrayBuffer arrayBuffer,
+                              size_t byteOffset) {
+  if (byteOffset > arrayBuffer.ByteLength()) {
+    NAPI_THROW(RangeError::New(env,
+        "Start offset is outside the bounds of the buffer"));
+    return DataView();
+  }
+  return New(env, arrayBuffer, byteOffset,
+      arrayBuffer.ByteLength() - byteOffset);
+}
+
+inline DataView DataView::New(napi_env env,
+                              Napi::ArrayBuffer arrayBuffer,
+                              size_t byteOffset,
+                              size_t byteLength) {
+  if (byteOffset + byteLength > arrayBuffer.ByteLength()) {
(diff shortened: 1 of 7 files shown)


romandev:
@mhdawson PTAL

romandev:
FYI, I found a bug in upstream. So, I sent a pull request (https://github.com/nodejs/node/pull/17869).

romandev:
Gentle ping :)

mhdawson:
Just catching up after being away for a few weeks.  Will take me a couple more days before I can review.

napi-inl.h
```diff
@@ -1171,6 +1171,88 @@ inline void ArrayBuffer::EnsureInfo() const {
...
+  size_t byteOffset;
+  napi_status status = napi_get_dataview_info(
+    _env,
+    _value      /* dataView */,
+    nullptr     /* byteLength */,
+    nullptr     /* data */,
+    nullptr     /* arrayBuffer */,
+    &byteOffset /* byteOffset */);
+  NAPI_THROW_IF_FAILED(_env, status, 0);
+  return byteOffset;
```
mhdawson: Does it make any sense to get the data once and cache it  as opposed to requesting it every time, I think at least some of them will not change.

napi.h
```diff
@@ -778,6 +778,32 @@ namespace Napi {
...
+                        size_t byteLength);
+
+    DataView();                               ///< Creates a new _empty_ DataView instance.
+    DataView(napi_env env, napi_value value); ///< Wraps a N-API value primitive.
+
+    Napi::ArrayBuffer ArrayBuffer() const;    ///< Gets the backing array buffer.
+    size_t ByteOffset() const;    ///< Gets the offset into the buffer where the array starts.
+    size_t ByteLength() const;    ///< Gets the length of the array in bytes.
+
+    // TODO: Should implement more methods to read/write data into array buffer.
```
mhdawson: Maybe explain here that this is a first step and support for additional methods is being added incrementally.

mhdawson:
A few comments.  One other general thought is that it would be great to start adding the doc at the same time when we add new APIs.

napi-inl.h
romandev: I considered that but I thought that it would be better to request the data everytime than spending additional memory. I think that ByteOffset(), ByteLength(), and ArrayBuffer() methods are necessary but I'm not sure if they are very often used. So, isn't it better to reduce memory usage? (Also, if users want to cache it, they can do it themselves.) WDYT?

napi.h
romandev: Done.

mhdawson:
Makes sense to make sure doc is added before we remove the gating flag.

napi-inl.h
mhdawson: I'm ok with keeping it simple for now, if we find it makes sense later on we can always add the cache.

romandev:
Gentle ping :)

mhdawson:
Landed as 673b59d31995091e83037ab53234778bcefb1d21

PR #212, opened 2018-01-12
Add ObjectReference test case

anisha-rohra:


test/objectreference.cc
```diff
@@ -0,0 +1,163 @@
+/* ObjectReference cant be used to create references to Values that
```
mhdawson: Should be can as opposed to cant

test/objectreference.js
```diff
@@ -0,0 +1,255 @@
...
+      binding.objectreference.setObjects("hello", "world");
+      assert.throws(
+        () => {
+          var rcount = binding.objectreference.refObjects("weak");
+          assert.equal(rcount, 1);
+          rcount = binding.objectreference.unrefObjects("weak");
+          assert.equal(rcount, 0);
+          binding.objectreference.unrefObjects("weak");
+        },
+        Error
```
mhdawson: If the first 2 asserts are supposed to pass I'm not sure this would catch failures of those asserts ?  (Since they throw and overall the assert.throws does not check what Error occured ?

mhdawson:
I see the following crash after the tests run when this new test is added:

```
Running test 'name'
Running test 'object/delete_property'
Running test 'object/get_property'
Running test 'object/has_own_property'
Running test 'object/has_property'
Running test 'object/object'
Running test 'object/set_property'
Running test 'promise'
Running test 'typedarray'
Running test 'objectwrap'
Running test 'objectreference'

All tests passed!
(node:60515) Warning: N-API is an experimental feature and could change at any time.
Tests aborted with SIGSEGV
npm ERR! Test failed.  See above for more details.
-sh-4.2$

```
If I revert this PR the failure goes away.

anisha-rohra:
@mhdawson when you get a chance, can you check if my recent changes still causes a segfault? Checking on macOS and an ubuntu machine, the segfault no longer occurs.

mhdawson:
Landed as 4d92a6066f9d0c3dffe9349728558433a55bec02

PR #214, opened 2018-01-14
n-api: throw RangeError in napi_create_dataview() with invalid range

romandev:
The API is required that `byte_length + byte_offset` is less than or
equal to the size in bytes of the array passed in. If not, a RangeError
exception is raised[1].

This is a partial cherry-pick from upstream[2].

[1] https://nodejs.org/api/n-api.html#n_api_napi_create_dataview
[2] https://github.com/nodejs/node/pull/17869/commits/2e329e9b776e0bea5e045a370858fb908dbdea5d


mhdawson:
Landed as 17c74e5a5e12752ee05a55e2f3c3724b56705088

PR #216, opened 2018-01-22
n-api: throw RangeError napi_create_typedarray()

romandev:
According to the ECMA spec, we should throw a RangeError in the
following cases:
  - `(length * elementSize) + offset` > the size of the array passed in
  - `offset % elementSize` != `0`

In the current implementation, this check was omitted. So, the following
code will cause a crash.
  ```
  napi_create_typedarray(env, napi_uint16_array, 2 /* length */,
                         buffer, 1 /* byte_offset */, &output_array);
  ```

This is a partial cherry-pick from upstream.

Refs:
https://tc39.github.io/ecma262/#sec-typedarray-buffer-byteoffset-length

### DIFF
--- a/src/node_api.cc
+++ b/src/node_api.cc
@@ -144,6 +144,30 @@ struct napi_env__ {
   (!try_catch.HasCaught() ? napi_ok \
                          : napi_set_last_error((env), napi_pending_exception))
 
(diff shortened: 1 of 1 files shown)


romandev:
@mhdawson Could you please check this patch?

mhdawson:
Landed as 5a39fdca6f510f52c9657fd4a89108d88be156a7

PR #218, opened 2018-01-22
Implement data manipulation methods for dataview

romandev:
This change implements additional methods for manipulating data in the
dataview object. This change includes the following things:
  - Get/Set${type} methods to manipulate data
  - Tests for the methods

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1211,6 +1211,14 @@ inline DataView::DataView() : Object() {
 }
 
 inline DataView::DataView(napi_env env, napi_value value) : Object(env, value) {
+  napi_status status = napi_get_dataview_info(
+    _env,
+    _value   /* dataView */,
+    &_length /* byteLength */,
+    &_data   /* data */,
+    nullptr  /* arrayBuffer */,
+    nullptr  /* byteOffset */);
+  NAPI_THROW_IF_FAILED(_env, status);
 }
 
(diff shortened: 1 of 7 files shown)


romandev:
@mhdawson Could you please check this patch?

napi-inl.h
```diff
@@ -1240,16 +1248,97 @@ inline size_t DataView::ByteOffset() const {
...
+  WriteData<uint16_t>(byteOffset, value);
+}
+
+inline void DataView::SetUint32(size_t byteOffset, uint32_t value) const {
+  WriteData<uint32_t>(byteOffset, value);
+}
+
+template <typename T>
+inline T DataView::ReadData(size_t byteOffset) const {
+  if (byteOffset + sizeof(T) > _length) {
```
sethbrenith: Looks like this could overflow and let somebody read up to 7 bytes before `_data`. I'm not sure how an attacker could do anything useful with that, but it might be surprising.

romandev: Good point. I've addressed your comment.

Thanks.

mhdawson:
@romandev just back from being out for a week and still catching up.  Will try and review later this week.

romandev:
@mhdawson Gentle ping.

mhdawson:
@romandev thanks for the reminder, have been pretty busy and at a conference this week but will try to find time to take a look.

mhdawson:
Landed as 9c4d321b57327de69e1f95bff06256174f001153

PR #228, opened 2018-02-11
Fix/nullptr error message

### LINKED ISSUE #224: Error constructor initializes _message std::string to nullptr; Visual C++ Warnings
In napi-inl.h, the Error:Error constructors initialize _message to nullptr. 

In Visual Studio C++ with Static Analysis enabled ('VCCLCompilerTool: { 'AdditionalOptions': ['/analyze'] } in bindings.gyp)) a warning is given that the nullptr could be zero and doesn't adhere to the specification of std::basic_string. 

> warning C6387: '_Param_(1)' could be '0'; this does not adhere to the specification for the function 'std::basic_string<char,std::char_traits<char>,std::allocator<char> >::{ctor}'. Lines: 1650

_message is std::string and not a pointer; it doesn't make sense to assign a null pointer literal to it. The explicit initialization of _message should be removed.

ebickle:
Fixes #224.

The Napi::Error constructor was initializing the `std::string _message` private member variable to `nullptr`. As a class, std::string automatically initializes to the equivalent of a blank string. `_message` isn't a pointer; assigning a pointer literal (nullptr) to it causes static analysis warnings.

I've also enabled Visual C++ Static Analysis (Prefast) in the test suite to detect these sorts of problems. If that isn't desirable let me know and I'll back out that part of the PR.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1729,7 +1729,7 @@ inline NAPI_NO_RETURN void Error::Fatal(const char* location, const char* messag
   napi_fatal_error(location, NAPI_AUTO_LENGTH, message, NAPI_AUTO_LENGTH);
 }
 
-inline Error::Error() : ObjectReference(), _message(nullptr) {
+inline Error::Error() : ObjectReference() {
 }
 
 inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullptr) {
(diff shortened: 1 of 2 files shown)


ebickle:
@mhdawson Prefast is clean but there are regular compiler warnings relating to lossy implicit conversions to double. Only affects the unit test code, not the main source code itself.

mhdawson:
Landed as faf19c4f7aafc416410e1ae26537c8ec4ba018e6

PR #243, opened 2018-03-29
New Promise and Reference docs

jschlight:
Continuing work for https://github.com/nodejs/abi-stable-node/issues/280

### DIFF
--- a/doc/promises.md
+++ b/doc/promises.md
@@ -1,5 +1,70 @@
-# Promise
-
 You are reading a draft of the next documentation and it's in continuos update so
 if you don't find what you need please refer to:
 [C++ wrapper classes for the ABI-stable C APIs for Node.js](https://nodejs.github.io/node-addon-api/)
+
+# Promise
+
+The Promise class, along with its Promise::Deferred class, implement the ability to create, resolve, and reject Promise objects.
+
+The basic approach is to create a Promise::Deferred object and return to your caller the value returned by the Promise::Deferred::Promise method. For example:
+
+```cpp
+Value YourFunction(const CallbackInfo& info) {
+  // your code goes here...
+  Promise::Deferred deferred = Promise::Deferred::New(info.Env());
+  // deferred needs to survive this call...
+  return deferred.Promise();
+}
+```
+
+Later, when the asynchronous process completes, call either the `Resolve` or `Reject` method on the Promise::Deferred object created earlier:
+
+```cpp
+  deferred.Resolve(String::New(info.Env(), "OK"));
+```
+
+## Promise::Deferred Methods
+
+### Factory Method
+
+```cpp
+static Promise::Deferred Promise::Deferred::New(napi_env env);
+```
+
+* `[in] env`: The `napi_env` environment in which to create the Deferred object.
+
+### Constructor
+
+```cpp
+Promise::Deferred(napi_env env);
+```
+
+* `[in] env`: The `napi_env` environment in which to construct the Deferred object.
+
+### Promise
+
+```cpp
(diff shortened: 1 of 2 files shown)


doc/reference.md
```diff
@@ -1,5 +1,115 @@
...
+
+## Methods
+
+### Factory Method
+
+```cpp
+static Reference<T> New(const T& value, uint32_t initialRefcount = 0);
+```
+
+* `[in] value`: The value which is to be referenced.
```
mhdawson: I think we should say

The Value object which is to be referenced.

mhdawson:
@gabrielschulhof  ping

doc/promises.md
```diff
@@ -1,5 +1,70 @@
-# Promise
-
 You are reading a draft of the next documentation and it's in continuos update so
```
kfarnung: nit: common typo: 'continuos' -> 'continuous'

doc/promises.md
```diff
@@ -1,5 +1,70 @@
...
+```cpp
+static Promise::Deferred Promise::Deferred::New(napi_env env);
+```
+
+* `[in] env`: The `napi_env` environment in which to create the Deferred object.
+
+### Constructor
+
+```cpp
+Promise::Deferred(napi_env env);
```
kfarnung: @gabrielschulhof Do you remember why we expose both a factory and a constructor?

doc/reference.md
```diff
@@ -1,5 +1,115 @@
...
+Creates a new _empty_ Reference instance.
+
+### Constructor
+
+```cpp
+Reference(napi_env env, napi_value value);
+```
+
+* `[in] env`: The `napi_env` environment in which to construct the Reference object.
+
```
kfarnung: nit: extra whitespaces between list items

doc/promises.md
```diff
@@ -1,5 +1,70 @@
...
+
+## Promise::Deferred Methods
+
+### Factory Method
+
+```cpp
+static Promise::Deferred Promise::Deferred::New(napi_env env);
+```
+
+* `[in] env`: The `napi_env` environment in which to create the Deferred object.
```
kfarnung: nit: looks like other docs use ` - ` (space-hyphen-space) for list items

doc/promises.md
gabrielschulhof: If you mean the fact that we have both a deferred and a promise, then the reason for that is that with V8 N-API cannot resolve a promise coming in from JavaScript, because it doesn't have the corresponding `v8::Resolver`.

In reality we could have gone with a single object, as @addaleax [points out](https://github.com/nodejs/node/pull/13717#issuecomment-309044219), but I chose not to because the coincidence between `v8::Promise` and `v8::Promise::Resolver` is an internal V8 detail and thus not guaranteed.

kfarnung: Sorry, I should have been more clear.  It seems like these objects have a lot of different mechanisms for construction.  Buffer (which I'm documenting) has four factory methods and two constructors.  In the process of writing the documentation the factory methods seem clear, but I haven't quite understood the purpose of the public constructors.  They seem like the end up putting the object into a weird (and potentially undesirable state).

mhdawson: @gabrielschulhof I don't think it was anything like that.

mhdawson: @kfarnung I'd suggest you ask @jasongin as he did a lot of the early work on the module.

mhdawson:
Going to land this and then we can improve through follow on PRs.

mhdawson:
Landed as 9d38f61afbd9f574c96b6a3ee401399a3e4bbb21

Issue #233, opened 2018-03-05
Async handles and Errors


fivdi:
Async handles can be used to "wakeup" the event loop and get a callback called from another thread. However, if the callback throws an error it doesn't appear to be possible to handle the error and have it displayed to the user.

An example of such an async handle can be seen [here](https://github.com/fivdi/node-addon-api-experiments/blob/master/src/addon.cc#L6). This async handle is [initialized to call DispatchEvent](https://github.com/fivdi/node-addon-api-experiments/blob/master/src/addon.cc#L38) when [told to do so once per second](https://github.com/fivdi/node-addon-api-experiments/blob/master/src/addon.cc#L29). The callback is [called by DispatchEvent](https://github.com/fivdi/node-addon-api-experiments/blob/master/src/addon.cc#L14). If the callback throws an error what needs to be done to display the error to the user?

mhdawson:
Thanks for the test case, this will help investigate and come up with an solution.

mhdawson:
Believe these are related: https://github.com/nodejs/node/issues/15371

mhdawson:
Made some progress on the plane today. Likely needs 2 PRs.  One to fix up escaping of handles when the handle is undefined and then one to add napi_throw_fatal.  Managed to to the test case to throw the expected error but still have work to validate what I have makes sense and then generate tests.

@fivdi thanks for the test case it was very helpful.

mhdawson:
This will add the function needed to get the exception to come out: https://github.com/nodejs/node/pull/19337, but there is still another fix needed for the right exception to come out in this test case.  I'll work on the PR for that.

mhdawson:
PR to address escape issue that was preventing the right exception from coming out.

https://github.com/nodejs/node/pull/19434

mhdawson:
Once the 2 PRs above land we can look at the PR to add a method in node-addon-api to be able to through an exception and validate the test case in this issue passes.

mhdawson:
This along with the earlier PR (nodejs/node#19337).

https://github.com/nodejs/node-addon-api/pull/245

Will let you get the correct exception if you use:

  `` napi_fatal_exception(env, e.Value());```

Next step will be to add function to through the fatal exception with node-addon-api iteself.

mhdawson:
@fivdi I think we'll need a test case added to the node-addon-api test suite which is pretty much what you had.  Do you want to contribute a PR to add it to the test suite ?

fivdi:
[...]

PR #245, opened 2018-04-03
don't try to escape null

mhdawson:
If an exception is pending from Call Or MakeCallback on
FunctionReference the return value may be nullptr.
Check for a pending exception and don't try to escape in these cases.

Refs:  https://github.com/nodejs/node-addon-api/issues/233

Next step after this would be to add support for throwing fatal exception with node-addon-api which can include a test that validates the exception is correct.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2285,36 +2285,60 @@ inline Napi::Value FunctionReference::operator ()(
 
 inline Napi::Value FunctionReference::Call(const std::initializer_list<napi_value>& args) const {
   EscapableHandleScope scope(_env);
-  return scope.Escape(Value().Call(args));
+  Napi::Value result = Value().Call(args);
+  if (scope.Env().IsExceptionPending()) {
+    return Value(); 
+  }
+  return scope.Escape(result);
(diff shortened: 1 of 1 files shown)


digitalinfinity:
Might be a dumb question but I thought the general model in node-addon-api is to have script exceptions propagate as C++ exceptions (i.e we'd thrown an exception)- so why doesn't Call or MakeCallback throw if an exception's pending?

digitalinfinity:
Right, I understand that the script method that was called threw an exception- my question was more along the lines of, why not propagate that exception as a C++ exception? I guess your change only applies to the non-exception cases?

mhdawson:
It applies to the default case where the test case was failing.  The problem was that it tries to unwrap then it throws a different exception so you get the wrong exception being returned.  That would likely be a problem in both cases.

mhdawson:
@digitalinfinity @gabrielschulhof can you review?

mhdawson:
Landed as cf6c93e4ee3d3e1d8b436f6090fb868501c4f286

PR #253, opened 2018-04-30
Add resource_name and resource parameters to AsyncWorker constructor

romandev:
This PR is initiated from https://github.com/nodejs/node-addon-api/pull/140#discussion_r140302349.

### DIFF
--- a/napi.h
+++ b/napi.h
@@ -1515,7 +1515,20 @@ namespace Napi {
 
   protected:
     explicit AsyncWorker(const Function& callback);
-    explicit AsyncWorker(const Object& receiver, const Function& callback);
+    explicit AsyncWorker(const Function& callback,
+                         const char* resource_name);
+    explicit AsyncWorker(const Function& callback,
+                         const char* resource_name,
+                         const Object& resource);
+    explicit AsyncWorker(const Object& receiver,
+                         const Function& callback);
+    explicit AsyncWorker(const Object& receiver,
+                         const Function& callback,
+                         const char* resource_name);
+    explicit AsyncWorker(const Object& receiver,
+                         const Function& callback,
+                         const char* resource_name,
+                         const Object& resource);
 
     virtual void Execute() = 0;
     virtual void OnOK();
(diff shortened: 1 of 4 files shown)


napi.h
```diff
@@ -1515,7 +1515,20 @@ namespace Napi {
...
+                const Object& resource);
+    AsyncWorker(const Object& receiver,
+                const Function& callback);
+    AsyncWorker(const Object& receiver,
+                const Function& callback,
+                const char* resource_name);
+    AsyncWorker(const Object& receiver,
+                const Function& callback,
+                const char* resource_name,
+                const Object& resource);
```
mhdawson: Overall looks good but do we need to add explicit (like there was for the existing methods) to avoid unwanted conversions?

test/asyncworker.js
```diff
@@ -1,21 +1,85 @@
...
 const buildType = process.config.target_defaults.default_configuration;
 const assert = require('assert');
+const async_hooks = require('async_hooks');
 
-test(require(`./build/${buildType}/binding.node`));
-test(require(`./build/${buildType}/binding_noexcept.node`));
+(async() => {
+  await test(require(`./build/${buildType}/binding.node`));
+  await test(require(`./build/${buildType}/binding_noexcept.node`));
+})();
```
mhdawson: I assume this was needed to make sure the tests ran sequentially.   Could you do it without depending on await as we run the tests  with older versions of Node.js that may not have support (ex 6.x I think)

test/asyncworker.js
```diff
@@ -1,21 +1,85 @@
...
-    assert.strictEqual(typeof this, 'object');
-    assert.strictEqual(this.data, 'test data');
-  }, 'test data');
+function installAsyncHooksForTest() {
+  return new Promise((resolve, reject) => {
+    let id;
+    const events = [];
+    const hook = async_hooks.createHook({
+      init(asyncId, type, triggerAsyncId, resource) {
+        if (type === 'TestResource'){
```
mhdawson: missing space between ) and {

romandev:
Thank you for review. I'm on a trip until this week. So, I'll update this patch until early next week. Thank you.

mhdawson:
@romandev thanks.  @NickNaso any additional comments before we land?

mhdawson:
Landed as 4b8918b3524bb8498dbac7811faf3e3c759d2054

PR #254, opened 2018-05-01
Updated Object documentation

anisha-rohra:


doc/object.md
```diff
@@ -1,5 +1,197 @@
...
+- `std::string&`
+- `bool`
+- `double`
+
+### Get()
+
+```cpp
+Value Napi::Object::Get(____ key);
+```
+- `[in] key`: The property whose assigned value is being returned.
```
mhdawson: How about
returns the value for the property with name `key` on the object.  Returns ?? if the property does not exist. 

We need to see what exactly it returns in that case where the property does not exist and update ??

doc/object.md
```diff
@@ -1,5 +1,197 @@
...
+
+Returns the [Value](value.md) associated with the key property.
+The `key` can be any of the following types:
+- `napi_value`
+- [Value](value.md)
+- `const char *`
+- `const std::string &`
+- `uint32_t`
+
```
mhdawson: We should say what it returns if the property does not exist.

doc/object.md
```diff
@@ -1,5 +1,197 @@
...
+### Has()
+
+```cpp
+bool Napi::Object::Has (____ key) const;
+```
+- `[in] key`: The property that is being checked for having an assigned value.
```
mhdawson: How about
the name of the property to check.

doc/object.md
```diff
@@ -1,5 +1,197 @@
...
+
+Returns a `bool` that is *true* if the `key` has a value property associated with it and *false* otherwise.
```
mhdawson: ..if the `object has a property name `key`  and *false* otherwise.

doc/object.md
```diff
@@ -1,5 +1,197 @@
...
+- `[in] properties`: A list of [PropertyDescriptor](propertydescriptor.md). Can be one of the following types:
+	- const std::initializer_list<PropertyDescriptor>&
+	- const std::vector<PropertyDescriptor>&
+
+Defines properties on the object.
+
+### Operator[]()
+
+```cpp
+PropertyLValue<std::string> Napi::Object::operator[] (const char* utf8name);
```
mhdawson: Some examples in the introduction which shows using these might help .

doc/object.md
```diff
@@ -1,5 +1,197 @@
...
+Returns the named property as a [Value](value.md).
+
+```cpp
+Value Napi::Object::operator[] (uint32_t index) const;
+```
+- `[in] index`: Element index.
+
+Returns an indexed property or array element as a [Value](value.md).
+
+### Example
```
mhdawson: I think moving the examples up into the intro would be better

doc/object.md
anisha-rohra: Not sure how to use these myself, do you know of any existing examples I can take a look at?

mhdawson:
Landed as 43ff9fa8366236d4e67be177cd794238f3de0bf1

PR #256, opened 2018-05-02
doc: ArrayBuffer and Buffer documentation

kfarnung:
* Also fixed a typo in the draft message

### DIFF
--- a/doc/array_buffer.md
+++ b/doc/array_buffer.md
@@ -1,5 +1,127 @@
-# Array buffer
+# ArrayBuffer
 
-You are reading a draft of the next documentation and it's in continuous update so
-if you don't find what you need please refer to:
-[C++ wrapper classes for the ABI-stable C APIs for Node.js](https://nodejs.github.io/node-addon-api/)
+The `ArrayBuffer` class corresponds to the JavaScript `ArrayBuffer` class.
+
+## Methods
+
+### New
+
+Allocates a new `ArrayBuffer` object with a given length.
+
+```cpp
+static ArrayBuffer New(napi_env env, size_t byteLength);
+```
+
+- `[in] env`: The environment in which to create the ArrayBuffer object.
+- `[in] byteLength`: The length to be allocated, in bytes.
+
+Returns a new `ArrayBuffer` object.
+
+### New
+
+Wraps the provided external data into a new `ArrayBuffer` object.
+
+The `ArrayBuffer` object does not assume ownership for the data and expects it
+to be valid for the lifetime of the object. Since the `ArrayBuffer` is subject
+to garbage collection this overload is only suitable for data which is static
+and never needs to be freed.
+
+```cpp
+static ArrayBuffer New(napi_env env, void* externalData, size_t byteLength);
+```
+
+- `[in] env`: The environment in which to create the `ArrayBuffer` object.
+- `[in] externalData`: The pointer to the external data to wrap.
+- `[in] byteLength`: The length of the `externalData`, in bytes.
+
+Returns a new `ArrayBuffer` object.
+
+### New
+
+Wraps the provided external data into a new `ArrayBuffer` object.
+
+The `ArrayBuffer` object does not assume ownership for the data and expects it
+to be valid for the lifetime of the object. The data can only be freed once the
+`finalizeCallback` is invoked to indicate that the `ArrayBuffer` has been
+released.
+
+```cpp
+template <typename Finalizer>
+static ArrayBuffer New(napi_env env,
+                       void* externalData,
+                       size_t byteLength,
+                       Finalizer finalizeCallback);
+```
+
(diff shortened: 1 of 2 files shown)


doc/buffer.md
```diff
@@ -1,5 +1,147 @@
...
+Returns a new Buffer object.
+
+### New
+
+Wraps the provided `data` into a Buffer object.
+
+The Buffer object does not assume ownership for the data and expects it to be
+valid for the lifetime of the Buffer object. Since the Buffer is subject to
+garbage collection this overload is only suitable for data which is static and
+never needs to be freed.
```
kfarnung: Can someone fact-check this?  I did some digging in the code and this appears to be true.

If so, it's a departure from NAN which would take ownership of the data and later free it.  Something to keep in mind for the conversion script, if someone was depending on this behavior there would be a memory leak in the converted code.

kfarnung: I opened #258 to discuss

doc/array_buffer.md
```diff
@@ -1,5 +1,127 @@
...
+- `[in] externalData`: The pointer to the external data to wrap.
+- `[in] byteLength`: The length of the `externalData`, in bytes.
+
+Returns a new `ArrayBuffer` object.
+
+### New
+
+Wraps the provided external data into a new `ArrayBuffer` object.
+
+The `ArrayBuffer` object does not assume ownership for the data and expects it
```
mhdawson: It is not more accurate that it effectively does take ownership, taking responsibility to run the finalizer to provided to delete the data when the class is deleted?

doc/array_buffer.md
```diff
@@ -1,5 +1,127 @@
...
+to be valid for the lifetime of the object. Since the `ArrayBuffer` is subject
+to garbage collection this overload is only suitable for data which is static
+and never needs to be freed.
```
mhdawson: I like your description here.  

I know we discussed doing a free, but thinking about it I don't like us trying to free an object that we don't know was allocated with malloc or new.

What would be good to figure out though is how to avoid conversion errors from NaN which result in leaks.  I'd be tempted to remove or deprecate this one and force callers to indicate they don't need finalization by passing a null to one of the other methods.  Since the wrapper is not part of the abi contract I think we could delete the method and provide instructions on what to use instead.

mhdawson: Maybe just deprecate as opposed to remove with the doc saying "don't use"

kfarnung: My only concern about "doc deprecated" is that it requires people to see the doc when converting their code.

doc/array_buffer.md
kfarnung: I toyed with the wording here a bit, but since the module code is still responsible for cleanup I feel like it's more accurate to say that ownership is never transferred.  The caller is merely lending the data to the `ArrayBuffer`.  Doing a bit more reading, it seems to align pretty well with the way V8 describes these "externalized" `ArrayBuffer` objects.

https://v8.paulfryzel.com/docs/master/classv8_1_1_array_buffer.html#a166848a34653afd4502e4cc33443815d

> After ArrayBuffer had been externalized, it does no longer own the memory block. The caller should take steps to free memory when it is no longer needed.

mhdawson: @kfarnung  did we close on what we wanted to do?  It would be good to get this landed.

mhdawson:
@kfarnung can you take a pass through to update to what you think we should do for now.  Would be good to get an initial version out and then we can iterate.

kfarnung:
Will do, I was on vacation for a week and I'm just getting back up to speed.

doc/buffer.md
```diff
@@ -1,5 +1,144 @@
...
+- `[in] data`: The pointer to the external data to expose.
+- `[in] length`: The number of `T` elements in the external data.
+
+Returns a new `Buffer` object.
+
+### New
+
+Wraps the provided external data into a new `Buffer` object.
+
+The Bu`Buffer`ffer object does not assume ownership for the data and expects it
```
gabrielschulhof: There's a typo here.

doc/buffer.md
```diff
@@ -1,5 +1,144 @@
...
+- `[in] finalizeHint`: The hint to be passed as the second parameter of the
+  finalize callback.
+
+Returns a new `Buffer` object.
+
+### Copy
+
+Allocates a new `Buffer` object and copies the provided external data into it.
+
+The `Buffer` object does not assume ownership for the data and expects it to be
```
gabrielschulhof: I don't get it. If it copies the data, how does it not assume ownership of it.

kfarnung: The Buffer doesn't assume ownership for the data you're providing to the `Copy` call, it copies it to memory that it does own.  This was supposed to contrast with the other Buffer factory functions, but maybe it's not necessary?

mhdawson:
Landed as 11697fcecd6e6b7519fda09a135397cea14dd701

PR #257, opened 2018-05-03
Update CI configs

kfarnung:
* Fix missing `pip` in macOS builds
* Add new builds to the matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,15 @@ os:
 env:
   global:
     # https://github.com/jasongin/nvs/blob/master/doc/CI.md
-    - NVS_VERSION=1.2.0
+    - NVS_VERSION=1.4.2
   matrix:
     - NODEJS_VERSION=node/4
     - NODEJS_VERSION=node/6
     - NODEJS_VERSION=node/8
+    - NODEJS_VERSION=node/9
+    - NODEJS_VERSION=node/10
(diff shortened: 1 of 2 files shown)


kfarnung:
https://travis-ci.org/kfarnung/node-addon-api

The matrix is pretty large now, I'm open to suggestions for reductions.

mhdawson:
We actually run the tests across platforms here: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api/

Right now we run master nightly, and the others when we create releases.  We should at least document that there are 2 places to look, if not try to cover them both in the community CI.

kfarnung:
I wasn't aware of that, I just had Travis-CI setup for my fork and it was failing for a while.  Does the that CI run include older versions of node as well?

kfarnung:
I think it's either worth landing these changes or deleting the files.  Do we have any node-addon-api CI coverage for ChakraCore in the official CI?

mhdawson:
If you mean in the official CI for the n-api tests, no.   I assume it would be a clone of the existing job, with possibly a smaller set of supported platforms, and then have it pull the chakra published versions.  It does depend on binaries being already built/available.

mhdawson:
Landed as 12b2cdeed3230ebeaa32caa67ec269e4f4d85388

PR #259, opened 2018-05-03
Documentation for error handling

NickNaso:
First step on writing documentation for error handling.

### DIFF
--- a/doc/error.md
+++ b/doc/error.md
@@ -1,5 +1,114 @@
 # Error
 
-You are reading a draft of the next documentation and it's in continuos update so
-if you don't find what you need please refer to:
-[C++ wrapper classes for the ABI-stable C APIs for Node.js](https://nodejs.github.io/node-addon-api/)
+The **Error** class is a representation of the JavaScript Error object that is thrown
+when runtime errors occur. The Error object can also be used as a base object for
+user defined exceptions.
+
+The **Error** class is a persistent reference to a JavaScript error object and
+inherits its behaviour from ObjectReference class (for more info see: [ObjectReference](object_reference.md)
+section).
+
+If C++ exceptions are enabled (for more info see: [Setup](setup.md) section),
+then the **Error** class extends `std::exception` and enables integrated
+error-handling for C++ exceptions and JavaScript exceptions.
+
+For more details about error handling refer to the section titled [Error handling](error_handling.md).
+
+## Methods
+
+### New
+
+Creates a new instance empty of `Error` object for the specified environment.
+
+```cpp
+Error::New(Napi:Env env);
+```
+
+- `[in] Env`: The environment in which to construct the Error object.
+
+Returns an instance of `Error` object.
+
+### New
+
+Creates a new instance of `Error` object
+
+```cpp
+Error::New(Napi:Env env, const char* message);
+```
+
+- `[in] Env`: The environment in which to construct the Error object.
+- `[in] message`: Null-terminated strings to be used as the message for the Error.
+
+Returns an instance of `Error` object.
+
+### New
+
+Creates a new instance of `Error` object
+
+```cpp
+Error::New(Napi:Env env, const std::string& message);
+```
+
+- `[in] Env`: The environment in which to construct the Error object.
+- `[in] message`: Reference string to be used as the message for the Error.
+
+Returns an instance of `Error` object.
+
+### Fatal
+
+In case of an unrecoverable error in a native module, a fatal error can be thrown
+to immediately terminate the process.
+
+```cpp
+static NAPI_NO_RETURN void Fatal(const char* location, const char* message);
+```
+
+The function call does not return, the process will be terminated.
+
+### Constructor
+
+Creates a new empty instance of `Error`
+
+```cpp
+Error();
+```
+
+### Constructor
+
+Initializes a `Error` instance from an existing JavaScript error object.
+
+```cpp
+Error(napi_env env, napi_value value);
+```
+
+- ```[in] Env```: The environment in which to construct the Error object.
+- ```[in] value```: The ```Error``` reference to wrap.
+
+Returns an instance of ```Error``` object.
+
+### Message
+
+```cpp
+std::string& Message() const NAPI_NOEXCEPT;
+```
+
+Returns the reference to string that represent the message of the error.
+
+### ThrowAsJavaScriptException
+
+Throw the error as JavaScript exception.
+
+```cpp
+void ThrowAsJavaScriptException() const;
+```
+
+Throws the error as JavaScript exception.
+
+### what
+
+```cpp
+const char* what() const NAPI_NOEXCEPT override;
+```
+
+Returns a pointer to a null-terminated string that is used to identify the
+exception. This method can be used only if the exception mechanism is enabled.
--- a/doc/error_handling.md
+++ b/doc/error_handling.md
@@ -1,5 +1,151 @@
 # Error handling
 
-You are reading a draft of the next documentation and it's in continuos update so
-if you don't find what you need please refer to:
-[C++ wrapper classes for the ABI-stable C APIs for Node.js](https://nodejs.github.io/node-addon-api/)
+Error handling represents one of the most important considerations when
+implementing a Node.js native add-on. When an error occurs in your C++ code you
+have to handle and dispatch it correctly. **N-API** uses return values and
+JavaScript exceptions for error handling. You can choose return values or
+exception handling based on the mechanism that works best for your add-on.
+
+The **Error** is a persistent reference (for more info see: [Object reference](object_reference.md)
+section) to a JavaScript error object. Use of this class depends on whether C++
+exceptions are enabled at compile time.
+
+If C++ exceptions are enabled (for more info see: [Setup](setup.md) section),
+then the **Error** class extends `std::exception` and enables integrated
+error-handling for C++ exceptions and JavaScript exceptions.
+
+The following sections explain the approach for each case:
+
+- [Handling Errors With C++ Exceptions](#exceptions)
+- [Handling Errors Without C++ Exceptions](#noexceptions)
+
+<a name="exceptions"></a>
+
+## Handling Errors With C++ Exceptions
+
+When C++ exceptions are enabled try/catch can be used to catch exceptions thrown
+from calls to JavaScript and then they can either be handled or rethrown before
+returning from a native method.
+
+If a N-API call fails without executing any JavaScript code (for example due to
+an invalid argument), then the N-API wrapper automatically converts and throws
+the error as a C++ exception of type **Error**.
+
+If a JavaScript function called by C++ code via N-API throws a JavaScript
+exception, then the N-API wrapper automatically converts and throws it as a C++
+exception of type **Error** on return from the JavaScript code to the native 
+method.
+
+If a C++ exception of type **Error** escapes from a N-API C++ callback, then
+the N-API wrapper automatically converts and throws it as a JavaScript exception.
+
+On return from a native method, N-API will automatically convert a pending C++
+exception to a JavaScript exception.
+
+When C++ exceptions are enabled try/catch can be used to catch exceptions thrown
+from calls to JavaScript and then they can either be handled or rethrown before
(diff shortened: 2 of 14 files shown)


doc/error.md
```diff
@@ -1,5 +1,101 @@
...
+The **Error** class is a representation of the JavaScript Error object that is thrown
+when runtime errors occur. The Error object can also be used as a base object for
+user defined exceptions.
+
+The **Error** class is a persistent reference to a JavaScript error object and
+inherits its behaviour from ObjectReference class (for more info see: [ObjectReference](object_reference.md)
+section).
+
+If C++ exceptions are enabled (for more info see: [Setup](setup.md) section),
+then the **Error** class extends ```std::exception``` and enables integrated
```
kfarnung: For single line code I think you can just use a single backtick.

doc/error.md
```diff
@@ -1,5 +1,101 @@
...
+
+Returns the reference to string that represent the message of the error.
+
+### ThrowAsJavaScriptException
+
+```cpp
+void ThrowAsJavaScriptException() const;
+```
+
+Throw the error as JavaScript exception.
```
kfarnung: Nit: I would move this above where the description normally goes and then update it to say "Throws"

doc/error.md
```diff
@@ -1,5 +1,101 @@
...
+
+### what
+
+```cpp
+const char* what() const NAPI_NOEXCEPT override;
+```
+
+Returns a pointer to a null-terminated string that is used to identify the
+exception. This method can be used only if the eceptions mechanis is enabled.
```
kfarnung: nit: "exception mechanism"

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
 # Error handling
 
-You are reading a draft of the next documentation and it's in continuos update so
-if you don't find what you need please refer to:
-[C++ wrapper classes for the ABI-stable C APIs for Node.js](https://nodejs.github.io/node-addon-api/)
```
kfarnung: What is our feeling about this?  I've seen some of the documentation updates leave it, but maybe it's the right time to start removing these designations.

/cc @mhdawson

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
\ No newline at end of file
+The error handling represents one of the most important thing on implementing a
```
kfarnung: nit: "Error handling represents one of the most important considerations when implementing a Node.js native add-on."

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+Node.js native add-on. When an error occurs in your C++ code you have to handle
+and dispatch it correctly. **N-API** uses return values and JavaScript exceptions
+for error handling. You can choice one method or other based on your preferences
+or on the C/C++, library that you are integrating.
+
+The **Error** is a persistent reference (for more info see: [Object reference](object_reference.md)
+section) to a JavaScript error object. Use of this class depends somewhat on
+whether C++ exceptions are enabled at compile time.
```
kfarnung: nit: avoid "somewhat", we want to be as descriptive as possible. If there are more nuances we should try to enumerate them.

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+
+The following sections explain the approach for each case:
+
+- [Handling Errors With C++ Exceptions](#exceptions)
+
```
kfarnung: nit: I don't think this extra newline is necessary here

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+- [Handling Errors Without C++ Exceptions](#noexceptions)
+
+<a name="exceptions"></a>
+
```
kfarnung: nit: extra newline

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+```cpp
+Napi::Env env = ...
+throw Napi::Error::New(env, "Example exception");
+// other C++ statements
+// ...
+```
+
+Following C++ statements will not be executed. The exception will bubble up as a
+C++ exception of type **Error**, until it is either caught while still in C++, or
+else automatically propataged as a JavaScript exception when returns to
```
kfarnung: nit: "propagated"

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+Since the exception was caught here, it will not be propagated as a JavaScript
+exception.
+
+<a name="noexceptions"></a>
+
+## Handling Errors Without C++ Exceptions
+
+If C++ exceptions are disabled (for more info see: [Setup](setup.md) section),
+then the **Error**  class does not extend ```std::exception```. This means that
+any call to node-addon-api functions does not throw C++ exception.
```
kfarnung: nit: "any calls to node-addon-api function do not throw C++ exceptions."

doc/error_handling.md
mhdawson: That was my thought, once we fill in the section lets remove the comment.

doc/error_handling.md
mhdawson: +1

doc/error_handling.md
mhdawson: Replied above at this point I'd say take it out.

doc/error.md
```diff
@@ -1,5 +1,101 @@
...
+```cpp
+Error();
+```
+
+### Constructor
+
+Initializes a ```Error``` instance from an existing ```Error``` object.
+
+```cpp
+TypeError(napi_env env, napi_value value);
```
mhdawson: Believe this should Error not TypeError

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+section) to a JavaScript error object. Use of this class depends somewhat on
+whether C++ exceptions are enabled at compile time.
+
+The following sections explain the approach for each case:
+
+- [Handling Errors With C++ Exceptions](#exceptions)
+
+- [Handling Errors Without C++ Exceptions](#noexceptions)
+
+<a name="exceptions"></a>
```
mhdawson: ? wondering why there is html here

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+
+## Handling Errors With C++ Exceptions
+
+If C++ exceptions are enabled (for more info see: [Setup](setup.md) section),
+then the **Error** class extends ```std::exception``` and enables integrated
+error-handling for C++ exceptions and JavaScript exceptions.
+
```
mhdawson: If C++ exceptions are enabled then exceptions are converted and handled as C++ exceptions while in native code and as JavaScript exceptions in JavaScript code.  What this means is that: 

and then make the following bullet points:

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+If a N-API call fails without executing any JavaScript code (for example due to
+an invalid argument), then the N-API wrapper automatically converts and throws
+the error as a C++ exception of type **Error**.
+
+If a JavaScript function called by C++ code via N-API throws a JavaScript
+exception, then the N-API wrapper automatically converts and throws it as a C++
+exception of type **Error**.
```
mhdawson: on return from the JavaScript code to the native method.

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+
+If a C++ exception of type **Error** escapes from a N-API C++ callback, then
```
mhdawson: On return from a native method, N-API will automatically convert a pending C++ exception to a JavaScript exception.

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+the N-API wrapper automatically converts and throws it as a JavaScript exception.
+
+Therefore, catching a C++ exception of type **Error** prevents a JavaScript
```
mhdawson: When C++ exceptions are enabled try/catch can be used to catch exceptions thrown from calls to JavaScript and then they can either be handled or rethrown before returning from a native method.

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+Napi::Value result = jsFunctionThatThrows({ arg1, arg2 });
+// other C++ statements
+// ...
+```
+
+Following C++ statements will not be executed. The exception will bubble up as a
+C++ exception of type **Error**, until it is either caught while still in C++, or
+else automatically propagated as a JavaScript exception when returns to 
+JavaScript.
+
```
mhdawson: An additional example where it catches (try/catch)  might be useful here.

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+} catch (const Napi::Error& e) {
+    cerr << "Caught JavaScript exception: " + e.what();
+}
+```
+
+Since the exception was caught here, it will not be propagated as a JavaScript
+exception.
+
+<a name="noexceptions"></a>
+
```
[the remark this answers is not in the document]
mhdawson: Maybe an example that shows rethrowing ?

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+After throwing a JS exception, the code should generally return immediately from
+the native callback, after performing any necessary cleanup.
+
+### Propagating a N-API JS exception
+
+```cpp
+Napi::Function jsFunctionThatThrows = someObj.As<Napi::Function>();
+Napi::Value result = jsFunctionThatThrows({ arg1, arg2 });
+if (result.IsEmpty()) return;
+```
```
mhdawson: I think it would be better to suggest that they should check using IsExceptionPending as that will work for methods that don't return a value as well.

doc/error_handling.md
```diff
@@ -1,5 +1,138 @@
...
+JavaScript exception is pending. To let the exception propagate, the code should
+generally return immediately from the native callback, after performing any
+necessary cleanup.
+
+### Handling a N-API JS exception
+
+```cpp
+Napi::Function jsFunctionThatThrows = someObj.As<Napi::Function>();
+Napi::Value result = jsFunctionThatThrows({ arg1, arg2 });
+if (result.IsEmpty()) {
```
mhdawson: Same goes here to check with IsExceptionPending

mhdawson:
Overall looking good, just a few comments and suggestions.

doc/error_handling.md
kfarnung: This is to create an anchor with a friendly name that's not affected by changes to the name of the header.  The same approach is used in the official [CHANGELOG](https://raw.githubusercontent.com/nodejs/node/master/doc/changelogs/CHANGELOG_V10.md)

mhdawson: ok, just have not see it in other docs.

doc/error.md
NickNaso: I will provide to remove where necessary.

doc/error_handling.md
NickNaso: I'm using markdownlint and it reports
```
[markdownlint] MD022/blanks-around-headers: Headers should be surrounded by blank lines
```
but if you want I can remove the extra space. Maybe in general we need to have a standard to follow.

doc/error.md
```diff
@@ -1,5 +1,101 @@
 # Error
 
-You are reading a draft of the next documentation and it's in continuos update so
-if you don't find what you need please refer to:
-[C++ wrapper classes for the ABI-stable C APIs for Node.js](https://nodejs.github.io/node-addon-api/)
\ No newline at end of file
+The **Error** class is a representation of the JavaScript Error object that is thrown
```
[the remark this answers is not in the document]
NickNaso: No at moment we haven't  a stylistic standard to follow. My opinion is that we need one.

NickNaso:
Completed fix proposed on first review

mhdawson:
@NickNaso I'm having trouble landing.  Can you squash the commits and then I'll try again.

NickNaso:
@mhdawson Could you try to land now?

NickNaso:
Hi @mhdawson I combined the work on this PR with https://github.com/nodejs/node-addon-api/pull/272. Please don't land this PR and if you want close it.

mhdawson:
Closing in favour of #272

PR #268, opened 2018-05-10
First pass at basic Node Addon API docs

digitalinfinity:
Added docs for Basic Types, Strings and Symbols. Feedback welcome, I think this will take a couple more iterations to get right.

### DIFF
--- a/doc/basic_types.md
+++ b/doc/basic_types.md
@@ -1,5 +1,360 @@
 # Basic Types
 
-You are reading a draft of the next documentation and it's in continuous update so
-if you don't find what you need please refer to:
-[C++ wrapper classes for the ABI-stable C APIs for Node.js](https://nodejs.github.io/node-addon-api/)
+Node Addon API consists of a few fundamental data types. These allow a user of 
+the API to create, convert and introspect fundamental JavaScript types, and
+interop with their C++ counterparts.
+
+## Value
+
+Value is the base class of Node Addon API's fundamental object type hierarchy.
+It represents a JavaScript value of an unknown type. It is a thin wrapper around
+the N-API datatype `napi_value`. Methods on this class can be used to check
+the JavaScript type of the underlying `napi_value` and also convert to C++ 
+types.
+
+### Constructor
+
+```cpp
+Value();                               
+```
+
+Used to create a Node Addon API `Value` that represents an **empty** value.
+
+```cpp
+Value(napi_env env, napi_value value); 
+```
+
+- `[in] env` - The `napi_env` environment in which to construct the Value object.
+- `[in] value` - The underlying JavaScript value that the `Value` instance represents
+
+Used to create a Node addon API `Value` that represents the `napi_value` passed
+in.
+
+### Operators
+
+#### operator napi_value
+
+```cpp
+operator napi_value() const;
+```
+
+Returns the underlying N-API value primitive. If the instance is _empty_,
+this returns  `nullptr`.
+
+#### operator ==
+
+```cpp
+bool operator ==(const Value& other) const;
+```
+
+Returns `true` if this value strictly equals another value, or `false` otherwise
+
+#### operator !=
+
+```cpp
+bool operator !=(const Value& other) const;
+```
+
+Returns `false` if this value strictly equals another value, or `true` otherwise
+
+### Methods
+
+#### From
+```cpp
+template <typename T>
+static Value From(napi_env env, const T& value);
+```
+
+- `[in] env` - The `napi_env` environment in which to construct the Value object.
+- `[in] value` - The C++ type to represent in JavaScript.
+
+This method is used to convert from a C++ type to a JavaScript value.
+Here, `value` may be any of:
+- `bool` - returns a `Napi::Boolean`
+- Any integer type - returns a `Napi::Number`
+- Any floating point type - returns a `Napi::Number`
+- `const char*` (encoded using UTF-8, null-terminated) - returns a `Napi::String`
+- `const char16_t*` (encoded using UTF-16-LE, null-terminated) - returns a `Napi::String`
+- `std::string` (encoded using UTF-8) - returns a `Napi::String`
+- `std::u16string` - returns a `Napi::String`
+- `napi::Value` - returns a `Napi::Value`
+- `napi_value` - returns a `Napi::Value`
+
+#### As
+```cpp
+template <typename T> T As() const;
+```
+
+Casts to another type of `Napi::Value`, when the actual type is known or assumed.
+
+This conversion does NOT coerce the type. Calling any methods inappropriate for 
+the actual value type will throw `Napi::Error`.
+
+#### StrictEquals
+```cpp
+bool StrictEquals(const Value& other) const;
+```
+
+- `[in] other` - The value to compare against
+
+Tests if this value strictly equals another value.
+
+#### Env
+```cpp
+Napi::Env Env() const;
+```
+
+Gets the environment the value is associated with. See `napi_env` for more 
+details about environments.
+
+#### IsEmpty
+```cpp
+bool IsEmpty() const;
+```
+
+Checks if the value is uninitialized.
+
+An empty value is invalid, and most attempts to perform an operation on an 
+empty value will result in an exception. Note an empty value is distinct from
+JavaScript `null` or `undefined`, which are valid values.
+
+When C++ exceptions are disabled at compile time, a method with a `Value` 
+return type may return an empty value to indicate a pending exception. So when 
+not using C++ exceptions, callers should check whether the value is empty 
+before attempting to use it.
+
+#### Type
+```cpp
+napi_valuetype Type() const;
+```
+
+Gets the underlying `napi_valuetype` of the value.
+
+#### IsUndefined
+```cpp
+bool IsUndefined() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `undefined` or `false`
+otherwise.
+
+#### IsNull
+```cpp
+bool IsNull() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `null` or `false`
+otherwise.
+
+#### IsBoolean
+```cpp
+bool IsBoolean() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `true` or JavaScript 
+`false` or `false` if the value is not a Boolean value in JavaScript.
+
+#### IsNumber
+```cpp
+bool IsNumber() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `Number` or `false`
+otherwise.
+
+#### IsString
+```cpp
+bool IsString() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `String` or `false`
+otherwise.
+
+#### IsSymbol
+```cpp
+bool IsSymbol() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `Symbol` or `false`
+otherwise.
+
+#### IsArray
+```cpp
+bool IsArray() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `Array` or `false`
+otherwise.
+
+#### IsArrayBuffer
+```cpp
+bool IsArrayBuffer() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `ArrayBuffer` or `false`
+otherwise.
+
+#### IsTypedArray
+```cpp
+bool IsTypedArray() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `TypedArray` or `false`
+otherwise.
+
+#### IsObject
+```cpp
+bool IsObject() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `Object` or `false`
+otherwise.
+
+#### IsFunction
+```cpp
+bool IsFunction() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `Function` or `false`
+otherwise.
+
+#### IsPromise
+```cpp
+bool IsPromise() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `Promise` or `false`
+otherwise.
+
+#### IsDataView
+```cpp
+bool IsDataView() const;
+```
+
+Returns `true` if the underlying value is a JavaScript `DataView` or `false`
+otherwise.
+
+#### IsBuffer
+```cpp
+bool IsBuffer() const;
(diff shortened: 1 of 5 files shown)


doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+Used to create a Node Addon API `Value` that represents an **empty** value.
+
+```cpp
+Value(napi_env env, napi_value value); 
+```
+
+- `[in] env` - The `napi_env` environment in which to construct the Value object.
+- `[in] value` - The underlying JavaScript value that the `Value` instance represents
+
+Used to create a Node addon API `Value` that represents the `napi_value` passed
```
NickNaso: Maybe here we need to use: Returns a Node addon API `Value` that represents the `napi_value` passed in.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+the actual value type will throw `Napi::Error`.
+
+#### StrictEquals
+```cpp
+bool StrictEquals(const Value& other) const;
+```
+
+- `[in] other` - The value to compare against
+
+Tests if this value strictly equals another value.
```
NickNaso: Maybe it's better move `Tests if this value strictly equals another value.` on top and here something like this: Returns `true` if the value isn strictly equals another value, or `false` otherwise.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+Checks if the value is uninitialized.
+
+An empty value is invalid, and most attempts to perform an operation on an 
+empty value will result in an exception. Note an empty value is distinct from
+JavaScript `null` or `undefined`, which are valid values.
+
+When C++ exceptions are disabled at compile time, a method with a `Value` 
+return type may return an empty value to indicate a pending exception. So when 
+not using C++ exceptions, callers should check whether the value is empty 
+before attempting to use it.
```
NickNaso: Maybe it's better to move this part on top and here: Returns `true` if the `Value` is an _empty_ value. 

About the error if the function that you are calling doesn't return value it' more appropriate use `Env::IsExceptionPending()` to retrieve if there is an active error.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+
+#### Type
+```cpp
+napi_valuetype Type() const;
+```
+
+Gets the underlying `napi_valuetype` of the value.
```
NickNaso: Maybe it's more appropriate use: Returns

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+
+#### ToBoolean
+```cpp
+Boolean ToBoolean() const;
+```
+
+Converts a `Napi::Value` to a JavaScript boolean value. This is a wrapper 
+around `napi_coerce_to_boolean`. This will throw a JavaScript exception if the 
+coercion fails.
+
```
NickNaso: Move this part on top and here something like this: Returns a JavaScript boolean value.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+#### ToNumber
+```cpp
+Number ToNumber() const;
+```
+
+Converts a `Napi::Value` to a JavaScript Number value. Note that this can cause
+script code to be executed according to JavaScript semantics. This is a wrapper 
+around `napi_coerce_to_number`. This will throw a JavaScript exception if the 
+coercion fails.
+
```
NickNaso: Move this part on top and here something like this: Returns a JavaScript Number value.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+#### ToString
+```cpp
+String ToString() const;
+```
+
+Converts a `Napi::Value` to a JavaScript String value. Note that this can cause
+script code to be executed according to JavaScript semantics. This is a wrapper 
+around `napi_coerce_to_string`. This will throw a JavaScript exception if the 
+coercion fails.
+
```
NickNaso: Move this part on top and here something like this: Returns a JavaScript String value.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+#### ToObject
+```cpp
+Object ToObject() const;
+```
+
+Converts a `Napi::Value` to a JavasScript Object. This is a wrapper around
+`napi_coerce_to_object`. This will throw a JavaScript exception if the 
+coercion fails.
+
```
NickNaso: Move this part on top and here something like this: Returns a JavaScript Object.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+## Name
+
+Names are JavaScript values that can be used as a property name. There are two
+specialized types of names supported in Node Addon API- `String` and `Symbol`.
```
NickNaso: Maybe you can link the doc of String and Symbol:
Names are JavaScript values that can be used as a property name. There are two
specialized types of names supported in Node Addon API- **[String](string.md)** and **[Symbol](symbol.md)**.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+```cpp
+Name();                               
+Name(napi_env env, napi_value value);
+```
+
+If the default constructor is invoked with no parameters, then an empty Name is
+created. If the constructor is invoked passing in an N-API value, then a Name is
+created from the JavaScript primitive. Note that the value is not coerced to a
+string.
+
```
NickNaso: Maybe it could be better to split this two cases

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+Arrays are native representations of JavaScript Arrays. `Napi::Array` is sugar
+around `napi_value` representing a JavaScript Array.
+
+### Constructor
+```cpp
+Array();
+```
+
+Returns an empy array. If an error occurs, a `Napi::Error` will get thrown.
+
```
NickNaso: empy -> empty

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+```cpp
+Array(napi_env env, napi_value value);
+```
+- `[in] env` - The environment in which to create the array.
+- `[in] value` - The primitive to wrap
+
+Wraps a `napi_value` as a `Napi::Array`. If an error occurs, a `Napi::Error` 
+will get thrown.
+
```
NickNaso: Move this on top and here something like this: Returns the  primitive top wrap.

doc/basic_types.md
mhdawson: Agreed we to follow the standard format we should have a Returns after the parameters.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+### Operators
+
+#### operator napi_value
+
+```cpp
+operator napi_value() const;
+```
+
+Returns the underlying N-API value primitive. If the instance is _empty_,
```
mhdawson: How about `Returns the underlying N-API napi_value'. If ...

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+this returns  `nullptr`.
+
+#### operator ==
+
+```cpp
+bool operator ==(const Value& other) const;
+```
+
+Returns `true` if this value strictly equals another value, or `false` otherwise
```
mhdawson: I think we need full sentences so needs a period at the end of the sentence.  Say goes throughout the rest of the doc.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+
+#### From
+```cpp
+template <typename T>
+static Value From(napi_env env, const T& value);
+```
+
+- `[in] env` - The `napi_env` environment in which to construct the Value object.
+- `[in] value` - The C++ type to represent in JavaScript.
+
```
mhdawson: Needs a Returns... before the explanation of how it works.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+- `std::string` (encoded using UTF-8) - returns a `Napi::String`
+- `std::u16string` - returns a `Napi::String`
+- `napi::Value` - returns a `Napi::Value`
+- `napi_value` - returns a `Napi::Value`
+
+#### As
+```cpp
+template <typename T> T As() const;
+```
+
```
mhdawson: Needs Returns.... same for other places, won't mention it specifically in those other places.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+Casts to another type of `Napi::Value`, when the actual type is known or assumed.
+
+This conversion does NOT coerce the type. Calling any methods inappropriate for 
+the actual value type will throw `Napi::Error`.
```
mhdawson: I think an example here would help.

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+```
+
+Checks if the value is uninitialized.
+
+An empty value is invalid, and most attempts to perform an operation on an 
+empty value will result in an exception. Note an empty value is distinct from
+JavaScript `null` or `undefined`, which are valid values.
+
+When C++ exceptions are disabled at compile time, a method with a `Value` 
+return type may return an empty value to indicate a pending exception. So when
```
mhdawson: So when -->  Therefore when

doc/basic_types.md
mhdawson: I'd have to double check the core node documentation but I think the explanation goes under the Returns statement.

mhdawson: What is the impact of script code being executed?

doc/basic_types.md
```diff
@@ -1,5 +1,360 @@
...
+```cpp
+static Array New(napi_env env, size_t length);
+```
+- `[in] env` - The environment in which to create the array.
+- `[in] length` - The lengthgn of the array
+
+This method returns a native representation of a JavaScript array. If an error
+occurs, a `Napi::Error` will get thrown.
+
+#### New
```
mhdawson: This should be #### Length

doc/string.md
```diff
@@ -39,13 +33,41 @@ operator std::u16string() const;
 ```
```
mhdawson: I think an example or more description might be useful here.

doc/string.md
```diff
@@ -39,13 +33,41 @@ operator std::u16string() const;
...
+String::New();
+```
+Creates a new empty String
+
+### New
+```cpp
+String::New(napi_env env, const std::string& value);
+String::New(napi_env env, const std::u16string& value);
+String::New(napi_env env, const char* value);
+String::New(napi_env env, const char16_t* value);
```
mhdawson: We should talk about this.  Its not consistent with anything I've seen elsewhere in the node documentation

mhdawson: But with overrides it does make sense to me

mhdawson:
Left a few comments but a really good first cut.

mhdawson:
@digitalinfinity are you going to get a chance to update for the comments?

digitalinfinity:
@mhdawson yes- so sorry, I've been sidetracked with other things for the past couple weeks but I should be able to find some cycles this week to incorporate all the great feedback provided so far.

digitalinfinity:
Ok, finally got a chance to incorporate the feedback. @NickNaso @mhdawson great feedback btw. @mhdawson I still need to add a couple examples but I think all of the other comments have been addressed.

doc/string.md
digitalinfinity: Yes, that was kind of what I was thinking. Anyone else have thoughts?

doc/basic_types.md
digitalinfinity: This could cause re-entrancy, and therefore unexpected behavior in the users module if their native module is not re-entrant proof. Thought calling it out explicitly in the doc could be helpful.

doc/basic_types.md
```diff
@@ -1,5 +1,409 @@
 # Basic Types
 
-You are reading a draft of the next documentation and it's in continuous update so
-if you don't find what you need please refer to:
-[C++ wrapper classes for the ABI-stable C APIs for Node.js](https://nodejs.github.io/node-addon-api/)
+Node Addon API consists of a few fundamental data types. These allow a user of 
+the API to create, convert and introspect fundamental JavaScript types, and
+interop with their C++ counterparts.
```
gabrielschulhof: I think you should expand this to "interoperate".

doc/basic_types.md
```diff
@@ -1,5 +1,409 @@
...
+
+#### IsEmpty
+```cpp
+bool IsEmpty() const;
+```
+
+Returns `true` if the value is uninitialized.
+
+An empty value is invalid, and most attempts to perform an operation on an 
+empty value will result in an exception. Note an empty value is distinct from
```
gabrielschulhof: I don't think you need "Note" here.

doc/basic_types.md
```diff
@@ -1,5 +1,409 @@
...
+
+Returns `true` if the underlying value is a JavaScript `DataView` or `false`
+otherwise.
+
+#### IsBuffer
+```cpp
+bool IsBuffer() const;
+```
+
+Returns `true` if the underlying value is a Node `Buffer` or `false`
```
gabrielschulhof: We should write "Node.js" instead of "Node".

doc/basic_types.md
```diff
@@ -1,5 +1,409 @@
...
+- `[in] value` - The primitive to wrap.
+
+Returns a Name created from the JavaScript primitive.
+
+Note:
+The value is not coerced to a string.
+
+## Array
+
+Arrays are native representations of JavaScript Arrays. `Napi::Array` is sugar
```
gabrielschulhof: I think we should say "a wrapper" instead of "sugar".

doc/basic_types.md
```diff
@@ -1,5 +1,409 @@
...
+around `napi_value` representing a JavaScript Array.
+
+### Constructor
+```cpp
+Array();
+```
+
+Returns an empty array.
+
+If an error occurs, a `Napi::Error` will get thrown. If C++ exceptions are not
```
gabrielschulhof: "will *be* thrown" might be better – here, and subsequently.

doc/basic_types.md
mhdawson: +1

doc/basic_types.md
mhdawson: +1

mhdawson:
I believe all comments are now addressed will land.

mhdawson:
Landed as e029a076c6bc5ae2303f4027c3c556ddc976918d

PR #255, opened 2018-05-01
doc: fix wrong link in readme

Miloas:

PR #273, opened 2018-05-18
Update README.md

amincheloh:
fix wrong example URL


NickNaso:
@mhdawson @kfarnung I think that we already have a PR that fix the same problem here [https://github.com/nodejs/node-addon-api/pull/255](https://github.com/nodejs/node-addon-api/pull/255)

mhdawson:
Yes that earlier on fixes the same issue.

PR #278, opened 2018-06-09
Update HTML Documentation

joshgarde:


mhdawson:
For some reason this fails to apply for me

joshgarde:
Something on my end?

mhdawson:
Not really sure, when I try to apply with git am it complains an aborts with conflicts, even though here in github it stays there are no conflicts.

PR #287, opened 2018-06-18
Fix HandleScope docs

rivertam:
It turns out `HandleScope::New` doesn't actually exist, so I'm not sure exactly how that happened.

Side note: Maybe it should, because I'm getting the error "Entering the V8 API without proper locking in place" when I just construct `HandleScope` so I think I'm missing something important that could be hidden using a factory function, but I don't know because I don't know why I'm running into this issue. =)

### DIFF
--- a/doc/handle_scope.md
+++ b/doc/handle_scope.md
@@ -15,10 +15,10 @@ the section titled (Object lifetime management)[object_lifetime_management].
 Creates a new handle scope.
 
 ```cpp
-HandleScope HandleScope::New(Napi:Env env);
+HandleScope::HandleScope(Napi:Env env);
 ```
 
-- `[in] Env`: The environment in which to construct the HandleScope object.
+- `[in] env`: The environment in which to construct the HandleScope object.
 
 Returns a new HandleScope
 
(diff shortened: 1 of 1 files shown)


doc/handle_scope.md
```diff
@@ -15,10 +15,10 @@ the section titled (Object lifetime management)[object_lifetime_management].
 Creates a new handle scope.
 
 ```cpp
-HandleScope HandleScope::New(Napi:Env env);
+HandleScope::HandleScope(Napi:Env env);
```
NickNaso:
```cpp
HandleScope(Napi::Env env);
 ```

File: doc/handle_scope.md [hunk 2]
```diff
@@ -28,11 +28,11 @@ Returns a new HandleScope
 Creates a new handle scope.
 
 ```cpp
-HandleScope HandleScope::New(napi_env env, napi_handle_scope scope);
+HandleScope::HandleScope(Napi::Env env, Napi::HandleScope scope);
```
NickNaso:
```cpp
HandleScope(Napi::Env env, Napi::HandleScope scope);
 ```

File: doc/handle_scope.md [hunk 3]
```diff
@@ -15,10 +15,10 @@ the section titled (Object lifetime management)[object_lifetime_management].
 Creates a new handle scope.
```
mhdawson: I think we should say "Creates a new handle scope on the stack"

doc/handle_scope.md
```diff
@@ -28,11 +28,11 @@ Returns a new HandleScope
...
```
mhdawson: Same comment as above

mhdawson:
@rivertam thanks for catching that.  You are correct that New does not exist and HandleScopes should be declared on the stack as opposed to allocated.

For example: 
```
Value createScope(const CallbackInfo& info) {
  {
    HandleScope scope(info.Env());
    String::New(info.Env(), "inner-scope");
  }
  return String::New(info.Env(), "scope");
}
```

mhdawson:
@rivertam made a few additional suggestions, if you can update PR to include those I'll go ahead and land.

mhdawson:
@rivertam do you have time to update this PR based on the comments?

rivertam:
No I'm sorry. I've been working like crazy. Might be able to get to it next week sometime, but it's probably faster and more effective if you just take it from here.

mhdawson:
Landed as 86be13a611c97f18dacedd689af0be08d5ad691e

PR #288, opened 2018-06-19
doc: Clarify positioning versus N-API

mhdawson:


README.md
```diff
@@ -1,24 +1,30 @@
...
-underlying JavaScript runtime (e.g. V8 or ChakraCore) and is maintained as 
-part of Node.js itself. This API will be Application Binary Interface (ABI) 
-stable across versions and flavors of Node.js. It is intended to insulate 
+# **node-addon-api module** 
+This module contains **header-only C++ wrapper classes** which simplify
+the use of the C based [N-API](https://nodejs.org/dist/latest/docs/api/n-api.html)
+provided by Node.js when using C++. It provides a C++ object model
+and exception handling semantics with low overhead.
+
+N-API is an ABI stale API for building native addons in C++. It is
```
rivertam: I think you just mean "N-API is a stable API" rather than "N-API is an ABI stale API"

Webern: I think you should also say `N-API is a ABI stable C interface provided by Node.js for building native addons...` It is not necessary that such addons be written in C++ and saying `in C++` muddies the waters a little bit IMO.

README.md
```diff
@@ -1,24 +1,30 @@
...
-recompilation. N-API guarantees the **API** and **ABI** compatibility across 
-different versions of Node.js. So if you switched to a different version of 
-Node.js, you would not need to reinstall or recompile the native addon.
-
-APIs exposed by N-API are generally used to create and manipulate JavaScript
-values. Concepts and operations generally map to ideas specified in the
-**ECMA262 Language Specification**.
+recompilation.
+
+The node-addon-api module preserves the benefits of the N-API as it consists
```
Webern: add ticks around node-addon-api like `node-addon-api`.

Recommendation:

> The `node-addon-api` module, which is not a part of Node.js, preserves the benefits of the C `N-API` by using inline `C++` code which only depends on the stable C ABI provided by `N-API`.

>

README.md
```diff
@@ -1,24 +1,30 @@
...
+only of inline code that depends only on the stable API provided by N-API.
+As such, modules built against one version of Node.js using node-addon-api
+should run without having to be rebuilt with newer versions of Node.js.
+As new APIs are added to N-API, node-addon-api must be updated to provide
```
Webern: Start a new paragraph at `As new APIs are added to...`

README.md
mhdawson: Thanks for catching those 2,  will update to address

mhdawson:
Landed as c2a620dc111765d20d8e958cdca024dd6618e234

Issue #296, opened 2018-07-11
node-addon-api test broken on 6.x 


mhdawson:
Seems the tests were broken on 6.x when I landed https://github.com/nodejs/node-addon-api/commit/4b8918b3524bb8498dbac7811faf3e3c759d2054.

It is just the test, not an ABI breakage:

```
Starting test suite

Running test 'arraybuffer'
Running test 'asyncworker'
module.js:478
    throw err;
    ^

Error: Cannot find module 'async_hooks'
    at Function.Module._resolveFilename (module.js:476:15)
    at Function.Module._load (module.js:424:25)
    at Module.require (module.js:504:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/mhdawson/newpull/node-addon-api/test/asyncworker.js:4:21)
    at Module._compile (module.js:577:32)
    at Object.Module._extensions..js (module.js:586:10)
    at Module.load (module.js:494:32)
    at tryModuleLoad (module.js:453:12)
    at Function.Module._load (module.js:445:3)
npm ERR! Test failed.  See above for more details.

```

mhdawson:
Working on fix.

PR #298, opened 2018-07-11
test: fix asyncworker test so it runs on 6.x

mhdawson:
Fixes: https://github.com/nodejs/node-addon-api/issues/296

### DIFF
--- a/test/asyncworker.js
+++ b/test/asyncworker.js
@@ -1,8 +1,21 @@
 'use strict';
 const buildType = process.config.target_defaults.default_configuration;
 const assert = require('assert');
-const async_hooks = require('async_hooks');
 const common = require('./common');
+
+// we only check async hooks on 8.x an higher were
+// they are closer to working properly
+const nodeVersion = process.versions.node.split('.')[0]
+let async_hooks = undefined;
+function checkAsyncHooks() {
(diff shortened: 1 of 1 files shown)


mhdawson:
@gabrielschulhof can you please review/approve.

gabrielschulhof:
Maybe we should refine this to perform the async work irrespective of whether hooks are available or not, but to not create, nor check the result of the promise if the hooks are unavailable.

mhdawson:
@gabrielschulhof I think it now does what you suggested.  It checks which version its running on and runs different tests on 6.x.

mhdawson:
Going to land this without waiting the usual 48 hours as it fixes a regression and we have 2 approvals.

mhdawson:
Also as an FYI, figured out how to have a run of the tests nightly on all of the LTS versions. Job is here:  https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-LTS%20versions/

The way it reports failures is not great as it's not clear which version the failure was on, but at least its a start and the info is available if you look at the console output for the top level job.

mhdawson:
Landed as dd1191e086fc4beb42de8d359ede9ad3e6776c56

PR #315, opened 2018-08-14
Tighten up warnings

mcheshkov:
This PR enable more warnings in tests and mark them as errors, to ensure that headers would not interfere with build environment of project using this library.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -171,7 +171,7 @@ struct AccessorCallbackData {
                                 napi_value exports) {     \
     return Napi::RegisterModule(env, exports, regfunc);   \
   }                                                       \
-  NAPI_MODULE(modname, __napi_ ## regfunc);
+  NAPI_MODULE(modname, __napi_ ## regfunc)
 
 // Adapt the NAPI_MODULE registration function:
 //  - Wrap the arguments in NAPI wrappers.
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2437,7 +2437,7 @@ inline PropertyDescriptor
 PropertyDescriptor::Accessor(const char* utf8name,
                              Getter getter,
                              napi_property_attributes attributes,
-                             void* data) {
+                             void* /*data*/) {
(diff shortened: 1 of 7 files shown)


napi-inl.h
```diff
@@ -3207,7 +3218,7 @@ inline void AsyncWorker::OnExecute(napi_env env, void* this_pointer) {
 }
 
 inline void AsyncWorker::OnWorkComplete(
-    napi_env env, napi_status status, void* this_pointer) {
+    napi_env /*env*/, napi_status status, void* this_pointer) {
```
mhdawson: Why is it necessary to comment out some of the parameter names like this?

mhdawson:
CI run to test across platforms: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api/653/

napi-inl.h
mcheshkov: Parameter of a function definition, that is not used in function body generate warning with `-Wunused-parameter` enabled.

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

Since library is header-only, even function definitions will be compiled with library user (as opposed to library author) flags. If library user has this warning enabled one will get warnings coming from library code, and to fix/hide them one will have to disable warnings for every compilation unit that includes this library, or wrap include with `#pragma push-ignore-pop` dance. Or fork/vendor library and fix warnings.

napi-inl.h
```diff
@@ -40,6 +43,14 @@ namespace details {
     return __VA_ARGS__;                                  \
   }
 
+// Usually ony could use NAPI_THROW_IF_FAILED(env, status, void()),
```
gabrielschulhof: How about we re-phrase this to
```C++
// We need a _VOID version of this macro to avoid warnings resulting from
// leaving the NAPI_THROW_IF_FAILED `...` argument empty..
```

gabrielschulhof:
Landed in 622ffaea76ffc21f1e8ecf659a77d79dd66a5e46.

PR #329, opened 2018-09-06
Added cpp highlight for string.md

yjaeseok:
There is no cpp highlight for some function in string.md file.


mhdawson:
Landed as 7cdd78726afeacb2d61c98fbc0ea7f8caacebe91

PR #339, opened 2018-09-12
Napi::ObjectWrap example does not compile

BotellaA:
The current example does not compile using GCC 4.8.


BotellaA:
@mhdawson Yes, I compiled it and it compiles, assuming you provide at some point an implementation for the methods but it is OK for the example.

mhdawson:
Landed as c7d54180ff0ef742445ed39ba20bd66aae815c96.  @BotellaA thanks for helping improve the documentation.

Issue #332, opened 2018-09-06
Methods without corresponding tests


mhdawson:
Need to document then list of methods that still need tests.

andrew-cc-chen:
| class                   | method                                                                                                                          |
|-------------------------|----------------------------------------------------------------------------------------------------------------------------------|
| Function                | Function()                                                                                                                       |
|                         | Value Call(size_t argc, const napi_value* args) const                                                                            |
|                         | Value Call(napi_value recv, size_t argc, const napi_value* args) const                                                           |
|                         | Value MakeCallback(napi_value recv, const std::initializer_list<napi_value>& args) const                                         |
|                         | Value MakeCallback(napi_value recv, const std::vector<napi_value>& args) const                                                   |
|                         | Value MakeCallback(napi_value recv, size_t argc, const napi_value* args) const                                                   |
|                         | Object New(size_t argc, const napi_value* args) const                                                                            |
| Object                  | Object()                                                                                                                         |
|                         | Object(napi_env env, napi_value value)                                                                                           |
|                         | PropertyLValue<std::string> operator [](const char* utf8name)                                                                    |
|                         | PropertyLValue<std::string> operator [](const std::string& utf8name)                                                             |
|                         | PropertyLValue<uint32_t> operator [](uint32_t index)                                                                             |
|                         | Value operator [](const char* utf8name) const                                                                                    |
|                         | Value operator [](const std::string& utf8name) const                                                                             |
|                         | Value operator [](uint32_t index) const                                                                                          |
|                         | bool Has(uint32_t index) const                                                                                                   |
|                         | Value Get(uint32_t index) const                                                                                                  |
|                         | template <typename ValueType> void Set(uint32_t index, const ValueType& value)                                                   |
|                         | bool Delete(uint32_t index)                                                                                                      |
|                         | void DefineProperties(const std::vector<PropertyDescriptor>& properties)                                                         |
|                         | bool InstanceOf(const Function& constructor) const                                                                               |
| Object::PropertyLValue  | operator Value() const                                                                                                           |
|                         | template <typename ValueType> PropertyLValue& operator =(ValueType value)                                                        |
| Number                  | Number()                                                                                                                         |
|                         | Number(napi_env env, napi_value value)                                                                                           |
|                         | operator int32_t() const                                                                                                         |
|                         | operator uint32_t() const                                                                                                        |
|                         | operator int64_t() const                                                                                                         |
|                         | operator float() const                                                                                                           |
[...]

PR #354, opened 2018-09-24
Some updates on Boolean documentation and added 5 new test case

NickNaso:
In this PR I worked on `Napi::Boolean` and in the specific I modified the documentation in a way that it is consistent with the other parts of the docs and I added 5 new test case.
Please review and suggest some other test cases. See the issue **[ https://github.com/nodejs/node-addon-api/issues/332](https://github.com/nodejs/node-addon-api/issues/332)**.

### DIFF
--- a/doc/boolean.md
+++ b/doc/boolean.md
@@ -1,29 +1,64 @@
 # Boolean
 
-# Methods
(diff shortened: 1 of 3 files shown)


mhdawson:
Landed as 4f76262a1083fe856412bf2b3a4f0dba8b0ff965

PR #355, opened 2018-09-25
tests: add operator overloading tests for Number class

nadongguri:
Hi node-addon-api members,

in this PR, I added tests for checking operator overloading for Number class.

class | methods
-- | --
Number | operator int32_t() const
  | operator uint32_t() const
  | operator int64_t() const
  | operator float() const
  | operator double() const

Please review and suggest some other test cases. See the issue https://github.com/nodejs/node-addon-api/issues/332.

### DIFF
--- a/test/basic_types/number.cc
+++ b/test/basic_types/number.cc
@@ -41,6 +41,31 @@ Value MaxDouble(const CallbackInfo& info) {
   return Number::New(info.Env(), DBL_MAX);
 }
 
+Value OperatorInt32(const CallbackInfo& info) {
+  Number jsValue = info[0].As<Number>();
+  return Boolean::New(info.Env(), jsValue.Int32Value() == static_cast<int32_t>(jsValue));
+}
+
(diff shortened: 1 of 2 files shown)


NickNaso:
Hi @nadongguri,
I just taken a look at your tests and they seem good. Maybe you can add the following test:

-  Empty `Napi::Number` using `Napi::Number()` in this case you can use the method `Napi::Value::IsEmpty()` to check the object you created is empty or not.

- Create a new `Napi::Number` object from another using constructor that take as parameters `napi_env` and `napi_value` (`Napi::Number(napi_env env, napi_value value)`). Maybe you can pass a `Number` to a function and create a new `Napi::Number` then return it to JavaScript and check if it is equal or not to your initial value.

Hope you agree with my proposal. Keep me updated. Thanks for your contribution.

nadongguri:
Hi @NickNaso,
I added these tests based on your guide. :)

class | methods
-- | --
Number | Number()
  | Number(napi_env env, napi_value value)

PR #361, opened 2018-10-02
Use Call() instead of MakeCallback() in AsyncWorker

romandev:
Change `AsyncWorker::OnOK()` and `AsyncWorker::OnError()` callbacks to
**NOT** use `MakeCallback()`. An ordinary function call
(`_callback::Call()`) is now correct.

Refs: https://nodejs.org/api/n-api.html#n_api_napi_make_callback

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3444,11 +3444,11 @@ inline FunctionReference& AsyncWorker::Callback() {
 }
 
 inline void AsyncWorker::OnOK() {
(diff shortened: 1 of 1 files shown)


NickNaso:
I agree with your solution in the `AsyncWorker` we use `napi_create_async_work` and in the `napi_async_complete_callback` the async context is already set up, so it's not necessary use `napi_make_callback`.

mhdawson:
Landed as fa3a6150b3dd46395e67053cfcba5a4d060c93b7

PR #365, opened 2018-10-08
Fix one of the Napi::Reference link in Object reference doc

fraxken:
One of the link to the Napi::Reference is broken.


gabrielschulhof:
Landed in 015d95312f5e411537592755377b28004d078db4.

PR #371, opened 2018-10-16
test: create test objects in the stack instead of the heap

nadongguri:
Hi, I guess most systems will free any memory related to the program when it  is terminated but I changed code to create object in the heap in tests to avoid small memory leak. :)


mhdawson:
Landed as 2342415463818c1477155f165acf8aeaee2c08b8

PR #377, opened 2018-10-25
src: implement AsyncPromise class

arc6011:
I created a class to meet a need I have for an asynchronous task which fulfills a promise. Perhaps this would be of some help to others.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3633,8 +3633,7 @@ inline void AsyncWorker::OnWorkComplete(
 // AsyncPromise class
 ////////////////////////////////////////////////////////////////////////////////
 
-template <class T>
-inline AsyncPromise<T>::BaseTask::~BaseTask() {
+inline AsyncPromise::BaseTask::~BaseTask() {
   if (_work != nullptr) {
     napi_status status = napi_delete_async_work(_env, _work);
     if ((status) != napi_ok)
@@ -3643,23 +3642,19 @@ inline AsyncPromise<T>::BaseTask::~BaseTask() {
   }
 }
 
-template <class T>
-inline AsyncPromise<T>::BaseTask::operator napi_async_work() const {
+inline AsyncPromise::BaseTask::operator napi_async_work() const {
   return _work;
 }
 
-template <class T>
-inline AsyncPromise<T>::BaseTask::operator napi_deferred() const {
+inline AsyncPromise::BaseTask::operator napi_deferred() const {
   return _deferred;
 }
 
-template <class T>
-inline Napi::Env AsyncPromise<T>::BaseTask::Env() const {
+inline Napi::Env AsyncPromise::BaseTask::Env() const {
   return _env;
 }
 
-template <class T>
-inline void AsyncPromise<T>::BaseTask::Queue() {
+inline void AsyncPromise::BaseTask::Queue() {
   AssertNotQueued();
   AssertNotFulfilled();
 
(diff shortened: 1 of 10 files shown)


arc6011:
I suppose that if napi_throw fails, the app is probably already going down in flames, eh?
Done.

napi-inl.h
```diff
@@ -3629,6 +3629,293 @@ inline void AsyncWorker::OnWorkComplete(
...
   delete self;
 }
+////////////////////////////////////////////////////////////////////////////////
+// AsyncPromise class
+////////////////////////////////////////////////////////////////////////////////
+
+template <class T>
+inline AsyncPromise<T>::BaseTask::~BaseTask() {
+  if (_work != nullptr) {
+    napi_delete_async_work(_env, _work);
```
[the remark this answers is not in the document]
arc6011: I suppose that if napi_throw fails, the app is probably already going down in flames, eh?
Done.

arc6011:
Replacing public constructors with static factory methods to make the class signature more consistent with other classes in the API.
I'll commit it as soon as I finish updating the docs and tests.

test/asyncpromise.cc
```diff
@@ -0,0 +1,67 @@
...
+    Object resource = info[1].As<Object>();
+    std::string data = info[2].As<String>();
+
+    AsyncPromise<TestPromise> async_promise(info.Env(), "TestResource", resource);
+
+    async_promise.Task()->_data = data;
+
+    switch (test)
+    {
+    case Test::succeed_before_queue:
```
mhdawson: This should be indented I think to line up with the other cases.

mhdawson:
Overall looks good to me.  Since there is a reasonable amount of code I would like to take another look next week and possibly discuss with the rest of the team in terms of fit with what we have in the API.  Thanks for the contribution!

test/asyncpromise.cc
arc6011: Thanks, Michael. Now fixed.

arc6011:
AsyncPromise is no longer a template class. The factory methods are now the only templates.

DaAitch:
@arc6011 this is quite a huge PR :)

I like the idea of combining `AsyncWorker` + `Promise` which is a good match.
Nevertheless I always prefer a composition of features over implementing feature specializations, which means I think an implementation of `AsyncWorker` simply using a `Promise` and implement a getter for that promise is much easier to understand then to understand what specific problem an `AsyncPromise` may solve.
My gut feeling tells me, that this can be a new npm module, but is too specific for `node-addon-api`.

mhdawson:
Create this issue to discuss if it should be in scope or not: https://github.com/nodejs/abi-stable-node/issues/363

extmchristensen:
I would expect support for async promises to be  a natual part of this node-addon-api. The feature would be really great to have available soon.

KevinEady:
@mhdawson ,

Going back to the [contributing sources doc](https://github.com/nodejs/node-addon-api/blob/master/CONTRIBUTING.md#source-changes) ... this `AsyncPromise` is neither a NAPI nor NAN functionality, so I feel like this PR doesn't belong in node-addon-api?

This PR is also almost a year old 😄 If it _doesn't_ belong here, should we close it?

mhdawson:
@KevinEady thanks for the ping, I agree we should close it.  Based on our guidance and discussions we'd see this being implemented in a separate module as opposed to being part of node-addon-api. 
 We've chosen to limit node-add-api to be a wrapper of n-api functions with the exception that functionality in nan can also be included in order to support migration.

PR #397, opened 2018-11-25
src : implement Cleanup class

nadongguri:
This is a wrapper class to support the following N-APIs.

napi_add_env_cleanup_hook()
napi_remove_env_cleanup_hook()
Refs: https://nodejs.org/api/n-api.html#n_api_cleanup_on_exit_of_the_current_node_js_instance

I checked this pr with node v8.12.0 and v10.12.0.
The document for this class is not implemented yet.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3705,6 +3705,28 @@ inline const napi_node_version* VersionManagement::GetNodeVersion(Env env) {
   return result;
 }
 
+#if (NAPI_VERSION > 2)
+////////////////////////////////////////////////////////////////////////////////
+// Cleanup class
+////////////////////////////////////////////////////////////////////////////////
+
+inline CleanupHook::CleanupHook(Env env) : _env(env) {
+  napi_status status = napi_add_env_cleanup_hook(_env, NapiHook, static_cast<void*>(this));
+  NAPI_THROW_IF_FAILED_VOID(_env, status);
+}
+
+inline CleanupHook::~CleanupHook() {
+  napi_status status = napi_remove_env_cleanup_hook(_env, NapiHook, static_cast<void*>(this));
+  NAPI_THROW_IF_FAILED_VOID(_env, status);
+}
+
+inline void CleanupHook::NapiHook(void* arg) {
+  CleanupHook* self = static_cast<CleanupHook*>(arg);
+  self->Cleanup();
+  delete self;
+}
+#endif
+
 // These macros shouldn't be useful in user code.
 #undef NAPI_THROW
 #undef NAPI_THROW_IF_FAILED
--- a/napi.h
+++ b/napi.h
@@ -1773,6 +1773,18 @@ namespace Napi {
       static const napi_node_version* GetNodeVersion(Env env);
   };
 
+#if (NAPI_VERSION > 2)
+  class CleanupHook {
+    public:
+      CleanupHook(Env env);
+      virtual ~CleanupHook();
+      virtual void Cleanup() = 0;
+    private:
+      napi_env _env;
+      static void NapiHook(void* arg);
+  };
+#endif
+
 } // namespace Napi
 
 // Inline implementations of all the above class methods are included here.
--- a/test/cleanuphook.cc
+++ b/test/cleanuphook.cc
@@ -7,16 +7,16 @@ using namespace Napi;
 class TestCleanupHook : public CleanupHook {
   public:
     static void DoCleanup(const CallbackInfo& info) {
-	  static int secret = 42;
-	  static int wrongSecret = 17;
-	  TestCleanupHook* testCleanupHook = new TestCleanupHook(info, wrongSecret);
-	  delete testCleanupHook;
+      static int secret = 42;
+      static int wrongSecret = 17;
+      TestCleanupHook* testCleanupHook = new TestCleanupHook(info, wrongSecret);
+      delete testCleanupHook;
 
       testCleanupHook = new TestCleanupHook(info, secret);
-	}
-	void Cleanup() {
-	  printf("cleanup(%d)\n", _secret);
-	};
+    }
+    void Cleanup() {
+      printf("cleanup(%d)\n", _secret);
+    };
   private:
     TestCleanupHook(const CallbackInfo& info, int secret) : CleanupHook(info.Env()), _secret(secret) {}
     int _secret;
--- a/test/cleanuphook.cc
+++ b/test/cleanuphook.cc
@@ -1,0 +1,31 @@
+#include "napi.h"
+
+using namespace Napi;
+
+#if (NAPI_VERSION > 2)
+
+class TestCleanupHook : public CleanupHook {
+  public:
+    static void DoCleanup(const CallbackInfo& info) {
+	  static int secret = 42;
+	  static int wrongSecret = 17;
+	  TestCleanupHook* testCleanupHook = new TestCleanupHook(info, wrongSecret);
+	  delete testCleanupHook;
+
+      testCleanupHook = new TestCleanupHook(info, secret);
+	}
+	void Cleanup() {
+	  printf("cleanup(%d)\n", _secret);
+	};
+  private:
+    TestCleanupHook(const CallbackInfo& info, int secret) : CleanupHook(info.Env()), _secret(secret) {}
+    int _secret;
+};
+
+Object InitCleanupHook(Env env) {
+  Object exports = Object::New(env);
+  exports["doCleanup"] = Function::New(env, TestCleanupHook::DoCleanup);
+  return exports;
+}
+
+#endif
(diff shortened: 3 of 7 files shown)


addaleax:
I’m not sure, but I think it might make sense to expose this as a class that other classes can inherit from, and which automatically cleans up once destroyed?

```c++
class CleanupHook {
 public:
  CleanupHook();  // calls napi_add_env_cleanup_hook with (this + NapiHook)
  ~CleanupHook();  // calls napi_remove_env_cleanup_hook with (this + NapiHook)
  virtual void Cleanup() = 0;

 private:
  void NapiHook(void* arg);  // calls this->Cleanup();
};
```

Otherwise this doesn’t provide a lot of extra value?

mhdawson:
I'm not sure forcing the creation of an object makes sense in this case.  I can see that in many cases a cleanup hook could be added but never removed since those hooks run when the environment exists and after that point you can't remove the hook anyway.

DaAitch:
Can you give a use case, when a cleanup hook makes sense?
I don't have a use-case, where a cleanup should not be implemented as dtor, but in a hook, except for those leaking global objects or data structures living on the heap which might be more common in C world. Is it possible that cleanup hooks are called, but no dtors?

@addaleax @mhdawson: I agree with both of you; maybe we can have both options?

1. `static EnvCleanupHook::Add(...)`, `static EnvCleanupHook::Remove(...)` (@nadongguri maybe we can use templated callbacks instead of C-functions, like e.g. `Callable` in `Napi::Function::New(...)`)
2. `class EnvCleanupHook` like @addaleax described above

gabrielschulhof:
@DaAitch addons now get unloaded when a thread exits, so they need to clean up after themselves: https://github.com/nodejs/node/blob/master/doc/api/addons.md#worker-support

napi-inl.h
```diff
@@ -3661,6 +3661,24 @@ inline const napi_node_version* VersionManagement::GetNodeVersion(Env env) {
   return result;
 }
 
+#if (NAPI_VERSION > 2)
+////////////////////////////////////////////////////////////////////////////////
+// Cleanup class
+////////////////////////////////////////////////////////////////////////////////
+
+inline void Cleanup::AddCleanupHook(Env env, void (*fun)(void* arg), void* arg) {
```
gabrielschulhof: No strong feelings, but do we need to repeat the word `Cleanup` here? How about `Napi::Cleanup::AddHook` and `Napi::Cleanup::RemoveHook`?

gabrielschulhof:
TBH I've been thinking for a while about defining an `Addon` class that serves as sort of the global static context for an addon, but an instance of which is heap-allocated in the addon's entry point. A `Cleanup` class, as @addaleax suggested would be a part of that, and a `Cleanup` class is perhaps a step in that direction.

nadongguri:
Thanks for good and kind reviews of this PR.
 At first time I thought users could know where and when hooks are called so simply implemented the cleanup class as static.
 In the meantime, I've checekd how BaseObject objects in the node use the cleanup hook. 
Based on you feedback, I will implement the cleanup class again.

nadongguri:
Hi, I have implemented it again based on @addaleax suggestion.

test/cleanuphook.cc
```diff
@@ -0,0 +1,31 @@
+#include "napi.h"
+
+using namespace Napi;
+
+#if (NAPI_VERSION > 2)
+
+class TestCleanupHook : public CleanupHook {
+  public:
+    static void DoCleanup(const CallbackInfo& info) {
+	  static int secret = 42;
```
romandev: nit: indentation

test/cleanuphook.cc
```diff
@@ -0,0 +1,31 @@
...
+	  static int wrongSecret = 17;
+	  TestCleanupHook* testCleanupHook = new TestCleanupHook(info, wrongSecret);
+	  delete testCleanupHook;
+
+      testCleanupHook = new TestCleanupHook(info, secret);
+	}
+	void Cleanup() {
```
romandev: You should check whether this Cleanup() method is called instead of printf().

napi-inl.h
```diff
@@ -3705,6 +3705,28 @@ inline const napi_node_version* VersionManagement::GetNodeVersion(Env env) {
...
+
+inline CleanupHook::~CleanupHook() {
+  napi_status status = napi_remove_env_cleanup_hook(_env, NapiHook, static_cast<void*>(this));
+  NAPI_THROW_IF_FAILED_VOID(_env, status);
+}
+
+inline void CleanupHook::NapiHook(void* arg) {
+  CleanupHook* self = static_cast<CleanupHook*>(arg);
+  self->Cleanup();
+  delete self;
```
romandev: This is unsafe. If the CleanupHooks is created on non-heap, this statement will cause a crash.

napi-inl.h
```diff
@@ -3705,6 +3705,28 @@ inline const napi_node_version* VersionManagement::GetNodeVersion(Env env) {
   return result;
 }
 
+#if (NAPI_VERSION > 2)
+////////////////////////////////////////////////////////////////////////////////
+// Cleanup class
+////////////////////////////////////////////////////////////////////////////////
+
+inline CleanupHook::CleanupHook(Env env) : _env(env) {
+  napi_status status = napi_add_env_cleanup_hook(_env, NapiHook, static_cast<void*>(this));
```
romandev: According to the document[1], `napi_add_env_cleanup_hook` can safely be specified multiple times with different `arg` values. But I guess your code doesn't allow it, right? Shouldn't we consider a situation that the method is called multiple times?

test/cleanuphook.cc
```diff
@@ -0,0 +1,31 @@
...
+
+#if (NAPI_VERSION > 2)
+
+class TestCleanupHook : public CleanupHook {
+  public:
+    static void DoCleanup(const CallbackInfo& info) {
+	  static int secret = 42;
+	  static int wrongSecret = 17;
+	  TestCleanupHook* testCleanupHook = new TestCleanupHook(info, wrongSecret);
+	  delete testCleanupHook;
```
romandev: If you want to delete the object here, do we really need to create the object on heap?

test/cleanuphook.cc
nadongguri: I will fix it.

napi-inl.h
nadongguri: As far as I know, hooks added by `napi_remove_env_cleanup_hook` API are called after the javascript code is done, so even if a subclass is created in the non-heap area such as stack, finally the destructor would be called so those hooks are removed. But honestly, I'm not sure 100% of the life cycle of the node so correct me if there's any wrong. :)

napi.h
```diff
@@ -1773,6 +1773,18 @@ namespace Napi {
       static const napi_node_version* GetNodeVersion(Env env);
   };
 
+#if (NAPI_VERSION > 2)
+  class CleanupHook {
```
nadongguri: As @gabrielschulhof reviewed before, class name EnvCleanup would be clear. I will rename it.

napi-inl.h
romandev: You could only allow on-heap object by using private constructor with static factory method.

addaleax: And `final`. :) I don’t understand why we’d want a heap-object-only restriction, though?

napi.h
addaleax: @nadongguri I don’t think referring to `Env` helps unless you’re a Node.js core developer (or embedder) and are familiar with what an `Environment` is.

I’d not include this in the name.

nadongguri: Okay. That makes sense.

napi-inl.h
romandev: > And final. :) I don’t understand why we’d want a heap-object-only restriction, though?

Am I missing something? I'm not sure if but this line always calls `delete self` whenever N-API Env hooks callback is called. Then, it will cause a crash. (e.g. if the object is created on static area) Therefore, we should remove this line and rewrite another way at least(if we don't want a heap-object-only).

mhdawson:
@nadongguri just thought I'd check on next steps.  I think this is waiting on some updates from your side right?

nadongguri:
Hi @mhdawson, I'm wondering about two things.
1. As @romandev pointed, shouldn't use printf function in the test code?
2. The case CleanupHook used is when BaseObject object is created in node.js. It registers it's base address(this pointer) as a cleanup argument for deleting itself. If my implementation is applied, the class object inherited from the Cleanup class will be automatically deleted when NapiHook method is called. 
In that sense, I wonder what the cleanup method can do.

KevinEady:
Is this still a valid wip PR...?

mhdawson:
@nadongguri are you still planning to work on this?

nadongguri:
Hi @mhdawson. :) 
Actually this PR is quite outdated. I was waiting for the feedback and I totally forgot about this. So ... is there anyone to commit this feature? Otherwise, I'll look into it.

mhdawson:
It's been quite a long time so I think I'll close this. @nadongguri please comment/reopen if you are still planning to work on it.

Issue #380, opened 2018-10-30
Object::GetPropertyNames() not const


AeromXundes:
The method `Object::GetPropertyNames()` is not declared `const`.

The method [`napi_get_property_names()`](https://nodejs.org/dist/latest/docs/api/n-api.html#n_api_napi_get_property_names) cannot not modify any member fields as the variables `_env` and `_value` are passed by value and `result` is a local variable. Thus, the function should be declared as `const`.

This will add the ability to declare a `const Object` and still get the property names associated with the object.

DaAitch:
@AeromXundes agree: PR created

PR #415, opened 2018-12-26
src: make `Object::GetPropertyNames()` const

DaAitch:
Closes: https://github.com/nodejs/node-addon-api/issues/380


mhdawson:
Landed as 020ac4a628debf7ce44429fd22c5bdff7c88a50a

PR #417, opened 2018-12-28
src: fix callbackData leaks on error napi status

DaAitch:


mhdawson:
Landed as 91eaa6f4cb970894ccbf7d9a0efb515d451191f2

Issue #419, opened 2018-12-28
noexcept control flow issues


DaAitch:
I think there are two noexcept issues with control flow.

#### 1. `NAPI_THROW` noexcept control flow
Like `NAPI_THROW_IF_FAILED*`, `NAPI_THROW` should also always abort control flow, but noexcept definition doesn't have a return so execution continues. All implementations using `NAPI_THROW` have a trailing `return` statement, so there is no bug, but I'd say it's unclean.

```cpp
// for noexcept
#define NAPI_THROW(e)  (e).ThrowAsJavaScriptException();
```

#### 2. noexcept `ThrowAsJavaScriptException` may produce infinite recursion
if `napi_throw` fails, it's likely that all further calls are failing and every time invoke another `ThrowAsJavaScriptException`, so I think this is a fatal case. Recursion is not a good idea, and also ignoring it, so it should fatal abort.

PR #420, opened 2018-12-28
src: fix noexcept control flow issues

DaAitch:
- adapt `NAPI_THROW`, create `NAPI_THROW_VOID`: either throw or
  return, never continue
- fix `Error::ThrowAsJavaScriptException`: if `napi_throw` fails, either
  explicitly throw or create fatal error

Closes: https://github.com/nodejs/node-addon-api/issues/419

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -19,11 +19,15 @@ namespace details {
 
 #ifdef NAPI_CPP_EXCEPTIONS
 
-#define NAPI_THROW(e)  throw e
-
 // When C++ exceptions are enabled, Errors are thrown directly. There is no need
-// to return anything after the throw statement. The variadic parameter is an
+// to return anything after the throw statements. The variadic parameter is an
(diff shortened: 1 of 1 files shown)


napi-inl.h
```diff
@@ -2031,8 +2044,20 @@ inline const std::string& Error::Message() const NAPI_NOEXCEPT {
 inline void Error::ThrowAsJavaScriptException() const {
   HandleScope scope(_env);
   if (!IsEmpty()) {
+
+    // We intentionally don't use `NAPI_THROW_*` macros here to point
```
gabrielschulhof: I don't understand "to point that there is no possible recursion". Perhaps, "to *ensure* that there is no possible recursion"?

DaAitch: agree

mhdawson:
Landed as 0b4027575216c3ceb5839d4699556b9a27814f88

PR #424, opened 2019-01-04
[WIP]: src: implement `NodeThreadScheduler`

DaAitch:
## This is a WIP PR.

Did some verbose commenting in the code as I was fixing some potential bugs/leaks from https://github.com/DaAitch/napi-experimental/commit/412a466ced8b6f4587db01a263acc21054bffdc7 as `napi_tsfn_abort` mode was used.

This code should work, but I think we can do better.

I have a new idea of how we can get rid of heap allocation of `NodeThreadScheduler`, having it copyable, movable on the stack, thus feels more lightweight.
I'll work on it soon.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3677,6 +3677,112 @@ inline void AsyncWorker::OnWorkComplete(
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+// Node Thread Scheduler class
+////////////////////////////////////////////////////////////////////////////////
+
+#if (NAPI_VERSION > 2147483646)
+
+namespace {
+  struct NodeThreadSchedulerCallbackData {
+    std::function<void(Env)> callback;
+    std::shared_ptr<NodeThreadScheduler> scheduler;
+  };
+
+  inline void NodeThreadSchedulerCallJsCb(napi_env env, napi_value /*js_callback*/,
+                                          void* /*context*/, void* data) {
+    using CallbackData = NodeThreadSchedulerCallbackData;
+
+    // will delete callback_data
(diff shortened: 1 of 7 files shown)


DaAitch:
So I thought about, why not implementing it in the napi itself like this [nodejs commit](https://github.com/DaAitch/node/commit/da1a97c149813d61764e369e9edb37421558e996)?
It's a more lightweight approach. I did some implementation, but there is a bug and I cannot find it, see failing tests. @gabrielschulhof I had a look at your `threadsafe_function` impl and tried to do it same way. Do you think you can have a look at it? Really appreciate.

gabrielschulhof:
@DaAitch AFAICT `napi_threadsafe_callback` doesn't use a queue like `napi_threadsafe_function` does, and `uv_async_t` does not guarantee that for every `uv_async_send` from a secondary thread it will wake up the main thread. 

OTOH it should be possible to break `v8impl::ThreadSafeFunction` into a base class of which `v8impl::ThreadSafeFunction` would be a subclass, and then create a second subclass which has the `ThreadsafeCallback` semantics.

TBH though the fact that we wish for this to be a single instance to be used by the entire addon for any number of reasons is beginning to turn this into a generic asynchronous queue implementation – which is perhaps what we really want, and which is perhaps what the base class will become.

KevinEady:
Hi @DaAitch ,

I think this can be closed since `ThreadSafeFunction` has been added to the core.

cc: @mhdawson

mhdawson:
Discussed in the N-API team meeting today.  Given that ThreadSafeFunction, function callback is optional and ThreadSafeFunction is available in all current LTS lines we believe we can close this.

Closing let us know if that is not the right thing to do.

Issue #368, opened 2018-10-09
Remove outdated Doxygen doc


mhdawson:
We have not kept it up to date and we have full coverage in our other doc, so remove it.

NickNaso:
Hi everyone,
I think that we think to take a decision about the **gh-pages** branch of this repo. Now it contains the old documentation  created by Docygen, but this documentation is not updated form the begin of July.
About that I have 3 proposals:
* Drop the old documentation and disable the **gh-pages**
* Continue to update the documentation generated by Doxygen
* Use **gh-pages** to better advertise **node-addon-api** and in general **N-API**. We could use tools like Gatsby and create a simple web site exposed by the **gh-pages**. In the  web site I would like to expose the new markdonwn documentation and other useful resources like materials of workshop and talks that we did.
I think that we can talk about this at Node.js Collaborator Summit.

NickNaso:
At Node.js Collaborator Summit we discussed about the idea to expose the documentation through the GitHub pages like a web site.
For the occasion I created a mock-up to better explain the idea. I attached the screenshot in this issue remembering that it’s a prototype and we need working on it to arrive at final version.
<img width="1286" alt="screenshot 2018-10-15 at 17 44 01" src="https://user-images.githubusercontent.com/387506/46986016-0d3a3d00-d0a2-11e8-8379-50c377f73112.png">
<img width="1287" alt="screenshot 2018-10-15 at 17 44 14" src="https://user-images.githubusercontent.com/387506/46986018-13c8b480-d0a2-11e8-9656-bc33336d7886.png">
<img width="1283" alt="screenshot 2018-10-15 at 17 44 32" src="https://user-images.githubusercontent.com/387506/46986020-16c3a500-d0a2-11e8-8588-9b1b1936144f.png">

I proposed to expose our markdown documentation through the web site together with some other resources that now we have to advertise if we want to increase the adoption of **node-addon-api** and **N-API**.
Now we have al the materials produced to create the **N-API workshop** and **slides** and **videos** of talks that members of the team did in this year. 
In the discussion it turns out that other than the markdown documentation we need to provide a **reference of the API** like V8, so we will have the documentation and a reference of the API that will cover all the developers’ needs.
From my perspective this will be a long term project now I will start on create what I proposed on my own then I will present to you the result and after some iterations we  will must to start thinking on how we want organize everything.
Please if someone have some other ideas I  will be happy to discuss on this issue or in the next N-API team meeting.

[...]

PR #426, opened 2019-01-10
Minor doc fixes

bmacnaughton:
As I was working through the docs I made edits when I noticed typos, misspellings, or in the case of object.md, incorrect information.

I find myself doing a bit more research than I wanted to in a few cases and wondered what your thoughts were on expanding some of the docs to include examples as opposed to (mostly just) API.

Here's a list of items that I found myself experimenting with to get something working because I couldn't deduce from the docs alone what I needed to do.

- external data
- DefineClass examples for Instance/Static methods/properties
- property_descriptors
- how to use class_property_descriptors

(actually I didn't get the last two working and just found a way around the issue but those were my first approaches.)

If you're open to providing more examples for the first two let me know where you think they belong and I'll work on some.

Thanks for the excellent doc. When I compare this to trying to get Nan working over a year ago it's night and day.

### DIFF
--- a/doc/class_property_descriptor.md
+++ b/doc/class_property_descriptor.md
@@ -8,7 +8,7 @@ This prevents using descriptors from a different class when defining a new class
 
 ## Methods
 
-### Contructor
+### Constructor
 
 Creates new instance of `Napi::ClassPropertyDescriptor` descriptor object.
 
--- a/doc/number.md
+++ b/doc/number.md
@@ -24,7 +24,7 @@ Creates a new instance of a `Napi::Number` object.
 Napi::Number(napi_env env, napi_value value);
 ```
 
- - `[in] env`: The `napi_env` environment in which to construct the `Napi::Nuber` object.
+ - `[in] env`: The `napi_env` environment in which to construct the `Napi::Number` object.
  - `[in] value`: The `napi_value` which is a handle for a JavaScript `Number`.
 
  Returns a non-empty `Napi::Number` object.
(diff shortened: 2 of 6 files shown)


mhdawson:
We are definitely open to adding more examples info to help clarify the concepts.  Will need a bit more time to take a look to see where they might go.  @NickNaso might also have some suggestions.

NickNaso:
Hi @bmacnaughton,
if you want add examples on documentation, you can refer on patterns the we follow in this part of the actual documentation:
https://github.com/nodejs/node-addon-api/blob/master/doc/async_worker.md#example
https://github.com/nodejs/node-addon-api/blob/master/doc/error_handling.md#examples-with-c-exceptions-enabled
https://github.com/nodejs/node-addon-api/blob/master/doc/async_context.md#example

In general we added the `Example` section after the API documentation. Feel free to be expressuve as you want. If your example are too complex maybe we need to consider the idea of creating an example to push on node-addon-example repo.

bmacnaughton:
Thanks. I'll take a look and think about it a bit.

NickNaso:
@bmacnaughton  please if it could help you feel free to open an issue about that so we can sharing all the ideas together.

DaAitch:
I hope I don't mix this, but as long as documentation is maintained twice, both should be up to date, but in the future we want to get rid of `/doc` (with #368), correct? So we should get this done soon?

mhdawson:
Landed as c629553cd7b08ea9e56a55176583219ee19a8a86.  @bmacnaughton thanks for the updates.

PR #442, opened 2019-02-05
Implement ThreadSafeFunction class

### LINKED ISSUE #312: Support for Asynchronous Thread-safe Function Calls
Hello, Thank you for the effort to create and maintain the C++ wrappers for N-API. The wrappers make it very much easier to create native addons.

N-API added capability for additional threads to invoke JavaScript functions based on the processing completed by those threads. Is there any timeline/plan to add wrappers for the same?

https://nodejs.org/docs/latest-v10.x/api/n-api.html#n_api_asynchronous_thread_safe_function_calls

Thank you

romandev:
This PR is implementing ThreadSafeFunction class wraps
napi_threadsafe_function features.

FYI, the test files that included in this PR have come from Node.js
repo[1]. They've been rewritten based on C++ and node-addon-api.

Fixes #312.

[1] https://github.com/nodejs/node/tree/master/test/node-api/test_threadsafe_function

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -125,6 +125,78 @@ struct FinalizeData {
   Hint* hint;
 };
 
+template <typename ContextType=void,
+          typename Finalizer=std::function<void(Env, void*, ContextType*)>,
+          typename FinalizerDataType=void>
+struct ThreadSafeFinalize {
+  static inline
+  void Wrapper(napi_env env, void* rawFinalizeData, void* /* rawContext */) {
+    if (rawFinalizeData == nullptr)
+      return;
+
+    ThreadSafeFinalize* finalizeData =
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env));
+    if (finalizeData->tsfn) {
+      *finalizeData->tsfn = nullptr;
+    }
+    delete finalizeData;
+  }
+
+  static inline
+  void FinalizeWrapperWithData(napi_env env,
+                               void* rawFinalizeData,
+                               void* /* rawContext */) {
+    if (rawFinalizeData == nullptr)
+      return;
+
+    ThreadSafeFinalize* finalizeData =
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env), finalizeData->data);
+    if (finalizeData->tsfn) {
+      *finalizeData->tsfn = nullptr;
+    }
+    delete finalizeData;
+  }
+
+  static inline
+  void FinalizeWrapperWithContext(napi_env env,
+                                  void* rawFinalizeData,
+                                  void* rawContext) {
+    if (rawFinalizeData == nullptr)
+      return;
+
+    ThreadSafeFinalize* finalizeData =
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env), static_cast<ContextType*>(rawContext));
+    if (finalizeData->tsfn) {
+      *finalizeData->tsfn = nullptr;
+    }
+    delete finalizeData;
+  }
+
+  static inline
+  void FinalizeFinalizeWrapperWithDataAndContext(napi_env env,
+                                         void* rawFinalizeData,
+                                         void* rawContext) {
+    if (rawFinalizeData == nullptr)
+      return;
+
+    ThreadSafeFinalize* finalizeData =
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env), finalizeData->data,
+        static_cast<ContextType*>(rawContext));
+    if (finalizeData->tsfn) {
+      *finalizeData->tsfn = nullptr;
+    }
+    delete finalizeData;
+  }
+
+  FinalizerDataType* data;
+  Finalizer callback;
+  napi_threadsafe_function* tsfn;
+};
+
 template <typename Getter, typename Setter>
 struct AccessorCallbackData {
   static inline
@@ -3628,6 +3700,325 @@ inline void AsyncWorker::OnWorkComplete(
   }
   if (!self->_suppress_destruct) {
     self->Destroy();
+  }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// ThreadSafeFunction class
+////////////////////////////////////////////////////////////////////////////////
+
+// static
+template <typename ResourceString>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount) {
+  return New(env, callback, Object(), resourceName, maxQueueSize,
+             initialThreadCount);
+}
+
+// static
+template <typename ResourceString, typename ContextType>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  ContextType* context) {
+  return New(env, callback, Object(), resourceName, maxQueueSize,
+             initialThreadCount, context);
+}
+
+// static
+template <typename ResourceString, typename Finalizer>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  Finalizer finalizeCallback) {
+  return New(env, callback, Object(), resourceName, maxQueueSize,
+             initialThreadCount, finalizeCallback);
+}
+
+// static
+template <typename ResourceString, typename Finalizer,
+          typename FinalizerDataType>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  Finalizer finalizeCallback,
+                                  FinalizerDataType* data) {
+  return New(env, callback, Object(), resourceName, maxQueueSize,
+             initialThreadCount, finalizeCallback, data);
+}
+
+// static
+template <typename ResourceString, typename ContextType, typename Finalizer>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  ContextType* context,
+                                  Finalizer finalizeCallback) {
+  return New(env, callback, Object(), resourceName, maxQueueSize,
+             initialThreadCount, context, finalizeCallback);
+}
+
+// static
+template <typename ResourceString, typename ContextType,
+          typename Finalizer, typename FinalizerDataType>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                                  const Function& callback,
+                                                  ResourceString resourceName,
+                                                  size_t maxQueueSize,
+                                                  size_t initialThreadCount,
+                                                  ContextType* context,
+                                                  Finalizer finalizeCallback,
+                                                  FinalizerDataType* data) {
+  return New(env, callback, Object(), resourceName, maxQueueSize,
+             initialThreadCount, context, finalizeCallback, data);
+}
+
+// static
+template <typename ResourceString>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  const Object& resource,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount) {
+  return New(env, callback, resource, resourceName, maxQueueSize,
+             initialThreadCount, static_cast<void*>(nullptr) /* context */);
+}
+
+// static
+template <typename ResourceString, typename ContextType>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  const Object& resource,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  ContextType* context) {
+  return New(env, callback, resource, resourceName, maxQueueSize,
+             initialThreadCount, context,
+             [](Env, ContextType*) {} /* empty finalizer */);
+}
+
+// static
+template <typename ResourceString, typename Finalizer>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  const Object& resource,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  Finalizer finalizeCallback) {
+  return New(env, callback, resource, resourceName, maxQueueSize,
+             initialThreadCount, static_cast<void*>(nullptr) /* context */,
+             finalizeCallback, static_cast<void*>(nullptr) /* data */,
+             details::ThreadSafeFinalize<void, Finalizer>::Wrapper);
+}
+
+// static
+template <typename ResourceString, typename Finalizer,
+          typename FinalizerDataType>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  const Object& resource,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  Finalizer finalizeCallback,
+                                  FinalizerDataType* data) {
+  return New(env, callback, resource, resourceName, maxQueueSize,
+             initialThreadCount, static_cast<void*>(nullptr) /* context */,
+             finalizeCallback, data,
+             details::ThreadSafeFinalize<
+                 void, Finalizer, FinalizerDataType>::FinalizeWrapperWithData);
+}
+
+// static
+template <typename ResourceString, typename ContextType, typename Finalizer>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                  const Function& callback,
+                                  const Object& resource,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  ContextType* context,
+                                  Finalizer finalizeCallback) {
+  return New(env, callback, resource, resourceName, maxQueueSize,
+             initialThreadCount, context, finalizeCallback,
+             static_cast<void*>(nullptr) /* data */,
+             details::ThreadSafeFinalize<
+                 ContextType, Finalizer>::FinalizeWrapperWithContext);
+}
+
+// static
+template <typename ResourceString, typename ContextType,
+          typename Finalizer, typename FinalizerDataType>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                                  const Function& callback,
+                                                  const Object& resource,
+                                                  ResourceString resourceName,
+                                                  size_t maxQueueSize,
+                                                  size_t initialThreadCount,
+                                                  ContextType* context,
+                                                  Finalizer finalizeCallback,
+                                                  FinalizerDataType* data) {
+  return New(env, callback, resource, resourceName, maxQueueSize,
+             initialThreadCount, context, finalizeCallback, data,
+             details::ThreadSafeFinalize<ContextType, Finalizer,
+                 FinalizerDataType>::FinalizeFinalizeWrapperWithDataAndContext);
+}
+
+inline ThreadSafeFunction::ThreadSafeFunction()
+  : _tsfn(new napi_threadsafe_function(nullptr)) {
+}
+
+inline ThreadSafeFunction::ThreadSafeFunction(
+    napi_threadsafe_function tsfn)
+  : _tsfn(new napi_threadsafe_function(tsfn)) {
+}
+
+inline ThreadSafeFunction::ThreadSafeFunction(ThreadSafeFunction&& other)
+  : _tsfn(std::move(other._tsfn)) {
+  other._tsfn.reset();
+}
+
+inline ThreadSafeFunction& ThreadSafeFunction::operator =(
+    ThreadSafeFunction&& other) {
+  if (*_tsfn != nullptr) {
+    Error::Fatal("ThreadSafeFunction::operator =",
+        "You cannot assign a new TSFN because existing one is still alive.");
+    return *this;
+  }
+  _tsfn = std::move(other._tsfn);
+  other._tsfn.reset();
+  return *this;
+}
+
+inline napi_status ThreadSafeFunction::BlockingCall() const {
+  return CallInternal(nullptr, napi_tsfn_blocking);
+}
+
+template <typename Callback>
+inline napi_status ThreadSafeFunction::BlockingCall(
+    Callback callback) const {
+  return CallInternal(new CallbackWrapper(callback), napi_tsfn_blocking);
+}
+
+template <typename DataType, typename Callback>
+inline napi_status ThreadSafeFunction::BlockingCall(
+    DataType* data, Callback callback) const {
+  auto wrapper = [data, callback](Env env, Function jsCallback) {
+    callback(env, jsCallback, data);
+  };
+  return CallInternal(new CallbackWrapper(wrapper), napi_tsfn_blocking);
+}
+
+inline napi_status ThreadSafeFunction::NonBlockingCall() const {
+  return CallInternal(nullptr, napi_tsfn_nonblocking);
+}
+
+template <typename Callback>
+inline napi_status ThreadSafeFunction::NonBlockingCall(
+    Callback callback) const {
+  return CallInternal(new CallbackWrapper(callback), napi_tsfn_nonblocking);
+}
+
+template <typename DataType, typename Callback>
+inline napi_status ThreadSafeFunction::NonBlockingCall(
+    DataType* data, Callback callback) const {
+  auto wrapper = [data, callback](Env env, Function jsCallback) {
+    callback(env, jsCallback, data);
+  };
+  return CallInternal(new CallbackWrapper(wrapper), napi_tsfn_nonblocking);
+}
+
+inline napi_status ThreadSafeFunction::Acquire() const {
+  return napi_acquire_threadsafe_function(*_tsfn);
+}
+
+inline napi_status ThreadSafeFunction::Release() {
+  return napi_release_threadsafe_function(*_tsfn, napi_tsfn_release);
+}
+
+inline napi_status ThreadSafeFunction::Abort() {
+  return napi_release_threadsafe_function(*_tsfn, napi_tsfn_abort);
+}
+
+inline ThreadSafeFunction::ConvertibleContext
+ThreadSafeFunction::GetContext() const {
+  void* context;
+  napi_get_threadsafe_function_context(*_tsfn, &context);
+  return ConvertibleContext({ context });
+}
+
+// static
+template <typename ResourceString, typename ContextType,
+          typename Finalizer, typename FinalizerDataType>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                                  const Function& callback,
+                                                  const Object& resource,
+                                                  ResourceString resourceName,
+                                                  size_t maxQueueSize,
+                                                  size_t initialThreadCount,
+                                                  ContextType* context,
+                                                  Finalizer finalizeCallback,
+                                                  FinalizerDataType* data,
+                                                  napi_finalize wrapper) {
+  static_assert(details::can_make_string<ResourceString>::value
+      || std::is_convertible<ResourceString, napi_value>::value,
+      "Resource name should be convertible to the string type");
+
+  ThreadSafeFunction tsfn;
+  auto* finalizeData = new details::ThreadSafeFinalize<ContextType, Finalizer,
+      FinalizerDataType>({ data, finalizeCallback, tsfn._tsfn.get() });
+  napi_status status = napi_create_threadsafe_function(env, callback, resource,
+      Value::From(env, resourceName), maxQueueSize, initialThreadCount,
+      finalizeData, wrapper, context, CallJS, tsfn._tsfn.get());
+  if (status != napi_ok) {
+    delete finalizeData;
+    NAPI_THROW_IF_FAILED(env, status, ThreadSafeFunction());
+  }
+
+  return tsfn;
+}
+
+inline napi_status ThreadSafeFunction::CallInternal(
+    CallbackWrapper* callbackWrapper,
+    napi_threadsafe_function_call_mode mode) const {
+  napi_status status = napi_call_threadsafe_function(
+      *_tsfn, callbackWrapper, mode);
+  if (status != napi_ok && callbackWrapper != nullptr) {
+    delete callbackWrapper;
+  }
+
+  return status;
+}
+
+// static
+inline void ThreadSafeFunction::CallJS(napi_env env,
+                                       napi_value jsCallback,
+                                       void* /* context */,
+                                       void* data) {
+  if (env == nullptr && jsCallback == nullptr) {
+    return;
+  }
+
+  if (data != nullptr) {
+    auto* callbackWrapper = static_cast<CallbackWrapper*>(data);
+    (*callbackWrapper)(env, Function(env, jsCallback));
+    delete callbackWrapper;
+  } else if (jsCallback != nullptr) {
+    Function(env, jsCallback).Call({});
   }
 }
 
--- a/napi.h
+++ b/napi.h
@@ -4,6 +4,7 @@
 #include "node_api.h"
 #include <functional>
 #include <initializer_list>
+#include <memory>
 #include <string>
 #include <vector>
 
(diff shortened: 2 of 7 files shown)


mhdawson:
@gabrielschulhof would be good if you took a look at this one.

napi.h
```diff
@@ -1760,6 +1761,32 @@ namespace Napi {
     std::string _error;
   };
 
+  class ThreadSafeFunction {
+  public:
+    using Callback = std::function<void(Napi::Env, Napi::Function)>;
+
+    ThreadSafeFunction();
+    ThreadSafeFunction(napi_env env,
```
gabrielschulhof: We use the factory method `::New()` for all our types. We should use one here too. In fact, we should have several factory methods which all call the same constructor and which differ from one another in that `resource_name` can be a `const char*`, a `std::string`, or a `napi_value`.

test/threadsafe_function/threadsafe_function.cc
```diff
@@ -0,0 +1,85 @@
...
+
+int sum = 0;
+
+static void data_source_thread(void* data) {
+  int args = 15;
+  ts_fn_wrap.Call([args](Napi::Env env, Napi::Function jsCallback) {
+    jsCallback.Call({ Number::New(env, args) });
+  }, napi_tsfn_blocking);
+  sum += 5;
+  if (data) {}
```
gabrielschulhof: Warnings about unused parameters should be prevented by doing
```C++
(void) data;
```
at the top of the function.

test/threadsafe_function/threadsafe_function.cc
```diff
@@ -0,0 +1,85 @@
...
+static Value StartThreadNoNative(const CallbackInfo& info) {
+  return StartThreadInternal(info, true);
+}
+
+static Value StartThreadNonblocking(const CallbackInfo& info) {
+  return StartThreadInternal(info, false);
+}
+
+static Value StopThread(const CallbackInfo& info) {
+  return Object::New(info.Env());
```
gabrielschulhof: These bindings and the ones below are not implemented.

test/threadsafe_function/threadsafe_function.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
gabrielschulhof: The test needs assertions.

test/threadsafe_function/threadsafe_function.cc
```diff
@@ -0,0 +1,85 @@
+#define NAPI_EXPERIMENTAL
```
legendecas: Does `NAPI_EXPERIMENTAL` still required? Maybe a more specific one `#define NAPI_VERSION 4` is preferred?

romandev:
@gabrielschulhof, @mhdawson PTAL
(If you don't mind, I'd like to push docs update into separated PR because this PR is enough large)

napi-inl.h
```diff
@@ -3618,6 +3618,159 @@ inline void AsyncWorker::OnWorkComplete(
...
+      || std::is_convertible<ResourceString, napi_value>::value,
+      "Resource name should be string convertible type");
+
+  napi_threadsafe_function tsFunctionValue;
+  auto* finalizeData = new details::FinalizeData<DataType, Finalizer>({
+      finalizeCallback, context });
+  napi_status status = napi_create_threadsafe_function(env, callback, resource,
+      Value::From(env, resourceName), maxQueueSize, initialThreadCount, data,
+      details::FinalizeData<DataType, Finalizer, Context>::WrapperWithHint,
+      finalizeData, CallJS, &tsFunctionValue);
```
KevinEady: Are these arguments in the right order?

```
napi_create_threadsafe_function(
  napi_env env,
  napi_value func,
  napi_value async_resource,
  napi_value async_resource_name,
  size_t max_queue_size,
  size_t initial_thread_count,
  void* thread_finalize_data,
  napi_finalize thread_finalize_cb,
  void* context,
  napi_threadsafe_function_call_js call_js_cb,
  napi_threadsafe_function* result);
```

I think there is a mixup after `initial_thread_count` (it looks like the thread finalize data and context are switched?)

KevinEady: Disregard, took a closer look at how `details::FinalizeData` works

mhdawson:
@romandev I'm ok with the docs being in a different PR, but I'd prefer if the two were landed together as opposed to landing one without the other.  That might make it difficult to do a release as I think we want to have both in place before having the functionality go out in a release.

KevinEady:
Is there anything I can do to help with this PR? Maybe docs or something? I have taken a look, and have been successfully using @romandev 's fork in my own project now.

NickNaso:
@KevinEady maybe you can start writing a draft to document this new api. What do you 
think @romandev @mhdawson @gabrielschulhof ?

KevinEady:
Hi @NickNaso , that sounds good. I have started on the docs, and I noticed this piece of code in the threadsafe_function.cc:

https://github.com/romandev/node-addon-api/blob/78b68bea4845413b9d06c584f8e21aa75e377058/test/threadsafe_function/threadsafe_function.cc#L39

This got me thinking about the context handling, and I noticed that we pass to `napi_create_threadsafe_function` a `FinalizeData` for the context. This means that the `call_js` callback will have this FinalizeData object passed as the context. Is this acceptable behavior?

KevinEady:
> This got me thinking about the context handling, and I noticed that we pass to `napi_create_threadsafe_function` a `FinalizeData` for the context. This means that the `call_js` callback will have this FinalizeData object passed as the context. Is this acceptable behavior?

I'm working on templating the ThreadSafeFunction class with the Context, and a no-Context tsfn will have `nullptr_t` as the Context type.  With this, I can also implement the (currently unimplemented) `GetContext()`, as well as correctly passing the `Context` to `CallJs` callback.

I am also adding overloaded methods for `ThreadSafeFunction<Context>::New()` for the optional finalizer, data, and context. 

If these API changes make sense, then that'll be what I'll document.

KevinEady:
@romandev So I did the work on templating the `Context`, take a look here: https://github.com/romandev/node-addon-api/compare/threadfunction...KevinEady:threadfunction-context?expand=1 

And it's used in the test class like https://github.com/KevinEady/node-addon-api/blob/8fc51e5cbeee30c6cdeac3fa0bc48c7cad9ef6d0/test/threadsafe_function/threadsafe_function.cc 

What are your thoughts?

KevinEady:
I'm currently working on the documentation https://github.com/KevinEady/node-addon-api/blob/threadfunction-doc/doc/threadsafe_function.md

napi-inl.h
```diff
@@ -3618,6 +3618,275 @@ inline void AsyncWorker::OnWorkComplete(
...
+
+// static
+template <typename ResourceString>
+inline ThreadSafeFunction ThreadSafeFunction::New(napi_env env,
+                                                  const Function& callback,
+                                                  ResourceString resourceName,
+                                                  size_t maxQueueSize,
+                                                  size_t initialThreadCount) {
+  return New(env, callback, Object(), resourceName, maxQueueSize,
+             initialThreadCount, nullptr);
```
DuBistKomisch: `nullptr` used here (and elsewhere) is ambiguous so shows up as `mismatched types ‘Context*’ and ‘std::nullptr_t’` when the intended overload is considered as a candidate

apparently need an explicitly overload with `std::nullptr_t` to resolve it

DuBistKomisch:
Also, the whole thing probably needs to be wrapped in a NAPI_VERSION check, explodes on 10.6 without declaring NAPI_EXPERIMENTAL.

mhdawson:
@DuBistKomisch it probably needs a NAPI version 1.4 check instead as I believe the N-API methods it uses are  non-experimental and that would check would exclude it from the earlier versions until we have the backports PRs landed.

mhdawson:
Although maybe that is what you suggested :)

DuBistKomisch:
yeah I'm no n-api expert but the node docs page doesn't seem to list a version under these thread safe function methods

mhdawson:
hmm, we seem to be missing that info for those methods.  I'll look at adding it  as they should show as version 4.

KevinEady:
@romandev will you be implementing a `GetContext()` function...?

napi-inl.h
```diff
@@ -3618,6 +3618,275 @@ inline void AsyncWorker::OnWorkComplete(
...
+                                                  Context* context) {
+  static_assert(details::can_make_string<ResourceString>::value
+      || std::is_convertible<ResourceString, napi_value>::value,
+      "Resource name should be string convertible type");
+
+  napi_threadsafe_function tsFunctionValue;
+  napi_status status = napi_create_threadsafe_function(env, callback, resource,
+      Value::From(env, resourceName), maxQueueSize, initialThreadCount, nullptr,
+      nullptr, context, CallJS, &tsFunctionValue);
+  if (status != napi_ok) {
```
gabrielschulhof: You don't need the `if`-statement around the `NAPI_THROW_IF_FAILED`.

napi-inl.h
```diff
@@ -3618,6 +3618,275 @@ inline void AsyncWorker::OnWorkComplete(
...
+inline ThreadSafeFunction::Status ThreadSafeFunction::NonBlockingCall(
+    DataType* data, Callback callback) const {
+  auto wrapper = [data, callback](Env env, Function jsCallback) {
+    callback(env, jsCallback, data);
+  };
+  return CallInternal(new CallbackWrapper(wrapper), napi_tsfn_nonblocking);
+}
+
+inline bool ThreadSafeFunction::Acquire() const {
+  return !IsAborted() && napi_acquire_threadsafe_function(
```
gabrielschulhof: This should handle all possible `napi_status`es returned by `napi_acquire_threadsafe_function()`.

napi-inl.h
```diff
@@ -3618,6 +3618,275 @@ inline void AsyncWorker::OnWorkComplete(
...
+      _tsFunctionValue) == napi_ok;
+}
+
+inline bool ThreadSafeFunction::Release() {
+  return !IsAborted() && napi_release_threadsafe_function(
```
gabrielschulhof: This also needs to handle all possible `napi_status`es.

napi-inl.h
```diff
@@ -3618,6 +3618,275 @@ inline void AsyncWorker::OnWorkComplete(
...
+    return false;
+  }
+
+  napi_status status = napi_release_threadsafe_function(
+      _tsFunctionValue, napi_tsfn_abort);
+
+  _tsFunctionValue = nullptr;
+  _env = nullptr;
+
+  return status == napi_ok;
```
gabrielschulhof: This should specifically handle `napi_ok`, `napi_generic_failure`, and `napi_invalid_arg`.

romandev: Sorry, I don't understand what you want exactly. Do you want this function to return three status instead of boolean? (like napi_release_threadsafe_function) If so, could you please explain the use cases in more details?

romandev:
@KevinEady If you're already implementing GetContext(), I'll leave a FIXME comment and then you can land your patch in follow-up PR. If you'd like to include the GetContext() implementation in this patch, I'll continue the work. Basically, I wanted to implement GetContext without explicit template. (But I'm not sure if it is possible or not.)

KevinEady:
@romandev alright yes, I will implement the `GetContext()` method. There was discussion in the N-API Group about context-aware addons, so we definitely need to support the ability to have context-aware tsfns. I don't think it's possible without templating the context since the type needs to exist across the `TSFN::New` static method and `TSFN::GetContext` instance method calls, but we shouldn't need to template the Finalizer (as we it was only used inside the native CallJs wrapper iirc, which we could have two of) 

I'll work on it hopefully over the weekend using your current implementation from https://github.com/nodejs/node-addon-api/commit/5fc684df2e41d23a441a8a3f22511b1ca0d1a426

KevinEady:
Hi @romandev ... So I took a stab at it again yesterday from your latest changes.. and yeah, I'm not sure how to complete the `GetContext()` API _without_ templating both the `Context` _and_ the finalizer `DataType` ... because you're passing to the napi_create_threadsafe_function the context of the `new details::FinalizeData<DataType, Finalizer>` and the finalize function as that `FinalizeData::WrapperWithHint` callback ... I'm not sure how to cast the response of `napi_get_threadsafe_function_context` to the proper FinalizeData without templating the `Context` _and_ the finalizer `DataType` (as then the Finalizer has type `std::function<void(napi_env, DataType*, Context* )>` allowing us to cast the napi context to the proper FinalizeData<DataType, Finalizer> ... that's what I ended up having to do in my first iteration (with the single TSFN::New method) ... https://github.com/KevinEady/node-addon-api/blob/threadfunction-context/napi-inl.h#L3847

Do you see any other way?

cc: @gabrielschulhof

extmchristensen:
@romandev Will this change allow calls into JS from another 3rd party thread similar to the napi-thread-safe-callback module (https://www.npmjs.com/package/napi-thread-safe-callback) ?

romandev:
@KevinEady, I implemented my idea for `GetContext() without class template` into latest patch. Could you please review the patch? If you mind my idea, I'll get rid of it from my patch. Thank you.

romandev:
@extmchristensen Sorry for belated response.
> Will this change allow calls into JS from another 3rd party thread similar to the napi-thread-safe-callback module?

Yes, it is. this patch wraps the N-APIs for thread safe function calls.
https://nodejs.org/dist/latest-v11.x/docs/api/n-api.html#n_api_asynchronous_thread_safe_function_calls

napi-inl.h
romandev: @gabrielschulhof Gentle ping

gabrielschulhof: @romandev sorry about the delay! Normally, on the main thread, we'd throw an exception if the status is anything other than `napi_ok`. Since this may or may not be the main thread, I think we should pass the `napi_status` through. In fact, we should probably not check `IsAborted()`, because that duplicates what `napi_threadsafe_function` already does. `napi_threadsafe_function` will return `napi_invalid_arg` if there are no more threads associated with it. Additionally, `napi_release_threadsafe_function` internally initiates a `uv_async_send` which may result in a `napi_generic_failure`. This status should not be swallowed.

I guess what I'm saying is that I don't believe that in this case we should do anything more than pass through what the underlying C API produces, because we do not have the facilities to throw errors on behalf of the user like we do on the main thread.

napi.h
```diff
@@ -1820,6 +1820,205 @@ namespace Napi {
...
+                                  const Object& resource,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  ContextType* context,
+                                  Finalizer finalizeCallback,
+                                  FinalizerDataType* data,
+                                  napi_finalize wrapper);
+
+    ThreadSafeFunction(napi_env env, napi_threadsafe_function tsFunctionValue);
```
gabrielschulhof: This should maybe not be private, so we can allow users to mix C-API and C++-API calls. We have similar constructors for things like `Napi::Object`, which are also public.

napi-inl.h
```diff
@@ -3627,6 +3686,357 @@ inline void AsyncWorker::OnWorkComplete(
...
+}
+
+inline bool ThreadSafeFunction::IsAborted() const {
```
gabrielschulhof: I don't believe we can accommodate such an API, because the checks below are not thread-safe. The thread-safe way to check if a `napi_threadsafe_function` has been aborted is to attempt to release it and to receive `napi_invalid_arg`. I understand that if you abort the thread-safe function in one thread you need not release it in other threads, but the flag informing you that you need not release it cannot itself be stored in the object representing the thread-safe function, because that is not thread-safe.

The behaviour whereby a thread receives `napi_closing` when attempting to place a call to indicate that it need not release the function has to be implemented in the application.

KevinEady:
Hi @romandev , 

I have been looking at your latest version with GetContext() and everything looks good. However, in response to @gabrielschulhof comments above, have you had any time to take a look at his two points? Furthermore, specifically regarding the `IsAborted()` API: this also means that you'd need to remove it from the TSFN destructor.

Do you have any time this week to work on this? Maybe we can divide and conquer somehow?

Thanks, Kevin

romandev:
@gabrielschulhof, @KevinEady 

Sorry for delayed update. From now on, I'll respond quickly.
I addressed all your comments. PTAL.

napi-inl.h
```diff
@@ -3627,6 +3686,343 @@ inline void AsyncWorker::OnWorkComplete(
...
+
+inline ThreadSafeFunction::ThreadSafeFunction(
+    napi_env env, napi_threadsafe_function tsFunctionValue)
+  : _env(env),
+    _tsFunctionValue(tsFunctionValue) {
+}
+
+inline ThreadSafeFunction::~ThreadSafeFunction() {
+  if (_env != nullptr && _tsFunctionValue != nullptr) {
+    Abort();
```
KevinEady: I do not think aborting on TSFN destruction is correct. If I create a TSFN with `initialThreadCount == 1`, and then `Release()` it -- since the threadcount is now 0, we shouldn't access it anymore... However, the TSFN's `_env` and `_tsFunctionValue` are still non-nullptr (as `Release()` does _not_ clear them, only `Abort()` does)... In this scenario, it would try to Abort() my previously Release()'d thread (if i am understanding correctly)

napi-inl.h
```diff
@@ -3627,6 +3686,343 @@ inline void AsyncWorker::OnWorkComplete(
...
+inline ThreadSafeFunction::ThreadSafeFunction(ThreadSafeFunction&& other)
+  : _env(other._env), _tsFunctionValue(other._tsFunctionValue) {
+  other._env = nullptr;
+  other._tsFunctionValue = nullptr;
+}
+
+inline ThreadSafeFunction& ThreadSafeFunction::operator =(
+    ThreadSafeFunction&& other) {
+  if (_env != nullptr && _tsFunctionValue != nullptr) {
+    Abort();
```
KevinEady: Same as above... What if `this` has been Release()'d , would it attempt to Abort()'d again...?

test/threadsafe_function/threadsafe_function.cc
```diff
@@ -0,0 +1,185 @@
...
+
+static Value StartThreadInternal(const CallbackInfo& info,
+    ThreadSafeFunctionInfo::CallType type) {
+  tsfnInfo.type = type;
+  tsfnInfo.abort = info[1].As<Boolean>();
+  tsfnInfo.startSecondary = info[2].As<Boolean>();
+  tsfnInfo.maxQueueSize = info[3].As<Number>().Uint32Value();
+
+  tsfn = ThreadSafeFunction::New(info.Env(), info[0].As<Function>(),
+      "Test", tsfnInfo.maxQueueSize, 2, &tsfnInfo, JoinTheThreads, uvThreads);
```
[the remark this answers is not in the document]
KevinEady: Hi @romandev ,

I think based off of @gabrielschulhof 's comments, it is not feasible for the object itself to keep track of whether it needs to be aborted. The underlying napi tsfn is sort of "refcounted" itself through the use of `Acquire`, `Release` and -- unfortunately for ease of this wrapping implementation -- `Abort`. 

So, yeah, you are correct.. In the above scenario, the first tsfn will always exist and never die. 

My thought is that, since the tsfn is refcounted, I would have the `ThreadSafeFunction`s move and assignment operators similar to `Reference`'s, where it just sets the its instance members accordingly, and not necessarily worrying about `Abort()`ing. There is no napi "how many threads are using the tsfn" call unfortunately, so we cannot be certain whether we need to do anything to the `ThreadSafeFunction` when it is moved/assigned. Unless you implement the counter yourself, which I don't think is right at all... 

Just spitballing here... Could we do something where a `ThreadSafeFunction` object can only hold one valid `_tsFunctionValue` at a time? So in the assignment operator, you would throw an error if `this->_tsFunctionValue` is non-nullptr? Is that valid...?

Do you have thoughts @gabrielschulhof ?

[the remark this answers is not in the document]
KevinEady: Hey @romandev ,

We talked about this in the NAPI meeting today See [here](https://www.youtube.com/watch?v=jmokE8uvc7Y&feature=youtu.be&t=2942) for the full transcript. So...

The node-addon-api wrapper itself should really be a convenience wrapper, and since the wrapper object cannot know the lifespan, it should not store instance members on itself. For example, if you have a ThreadSafeFunction that you Release() to the point where it is no longer valid (and the engine has cleaned it up), but you could still have valid `ThreadSafeFunction` object around that now has an invalid `_tsFunctionValue` reference. 

However...  It was brought up that napi tsfn does have the `Finalize` callback. So, the call to `TSFN::New()` is done in the main thread as well as the `Finalize()` callback, and these are the only places that "state" can be checked and modified (so my understanding would be the `_tsFunctionValue` member). I also remember there being an `_env` but I don't think this is necessary?)

Does this help any?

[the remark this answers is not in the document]
romandev: @KevinEady,

Thank you for your input!
My new commit includes the following things:
  - Remove _env member from TSFN
  - Reset internal _tsfn pointer in Finalize callback
  - Add a check whether the internal _tsfn pointer is valid in move assign operator.

It works well in my local env.  Could you please review it again?

romandev:
Gentle ping

KevinEady:
Hi @romandev , I hope to look at your changes within the next few days. Don't have any time right now.

Kevin

napi-inl.h
```diff
@@ -125,6 +125,78 @@ struct FinalizeData {
...
+                          void* rawFinalizeData,
+                          void* rawContext) {
+    if (rawFinalizeData == nullptr)
+      return;
+
+    ThreadSafeFinalize* finalizeData =
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env), static_cast<ContextType*>(rawContext));
+    if (finalizeData->tsfn) {
+      *finalizeData->tsfn = nullptr;
```
gabrielschulhof: Why are we dereferencing here? Shouldn't it be `finalizeData->tsfn = nullptr;`?

KevinEady:
Hi @romandev ,

I have updated my project with your latest changes and I see no problems.  However, I'm not doing anything too "crazy" and not really using all your PRs features to the fullest, eg. I am NOT:
1.  Using `Acquire()` / `Release()` after the TSFN creation
2. Using `Abort()`
3. Anything with `Finalizer`, `Context`, or async hooks `resource` object.

But I see you do have some of those in your threadsafe_function test cases.

I am updating the documentation now to reflect these changes, and hope to push it out after the N-API meeting.

Thanks, Kevin

KevinEady:
Hi @romandev ,

I'm not sure how we want to proceed... I don't know if I target _this_ nodejs repo instead but I created a PR against your fork for now. We can figure out the right approach later, but take a look at https://github.com/romandev/node-addon-api/pull/17 / https://github.com/romandev/node-addon-api/blob/94e11070d61fe48f8961ce0b466f6c59cd2d72c6/doc/threadsafe_function.md  

Thanks, Kevin

romandev:
@KevinEady Thank you for your update. 
> I'm not doing anything too "crazy" and not really using all your PRs features to the fullest

I think that we can fix the problem in follow-up PR before release if we found some problem :)

> I'm not sure how we want to proceed... I don't know if I target this nodejs repo instead but I created a PR against your fork for now. We can figure out the right approach later,

I recommend that you forks node/node-addon-api and then send a new PR apart from my PR. It's not problem because your `.md` file change is not depending on my PR. (It will make your change easier to review.)

> but take a look at ...

Looks overall good to me. BTW, you might have to match the format with existing documents as follows (e.g. [AsyncContext](https://github.com/romandev/node-addon-api/blob/94e11070d61fe48f8961ce0b466f6c59cd2d72c6/doc/async_context.md)):
```
ThreadSafeFunction
  ...
Methods
  ...
Operator
  ...
Example
  ...
```

KevinEady:
Hi @romandev ,

Thanks for the feedback. I was using the normal [Function](https://github.com/nodejs/node-addon-api/blob/master/doc/function.md) document which has the Example first, which I did find odd... I'll switch the order though to make it more understandable. 

Secondly, I didn't write _all_ the constructors out since there's, what, eight or so? Hope what I did makes sense.

I'll make the changes to the order, and create a new PR directly into node-addon-api, and I'll just keep the docs PR up-to-date with your src PR.

Thanks, Kevin

napi-inl.h
```diff
@@ -125,6 +125,78 @@ struct FinalizeData {
...
+    }
+    delete finalizeData;
+  }
+
+  static inline
+  void WrapperWithDataAndContext(napi_env env,
```
gabrielschulhof: Let's make it `FinalizeWrapperWithDataAndContext`.

napi-inl.h
```diff
@@ -3627,6 +3699,324 @@ inline void AsyncWorker::OnWorkComplete(
...
+                                                  ResourceString resourceName,
+                                                  size_t maxQueueSize,
+                                                  size_t initialThreadCount,
+                                                  ContextType* context,
+                                                  Finalizer finalizeCallback,
+                                                  FinalizerDataType* data,
+                                                  napi_finalize wrapper) {
+  static_assert(details::can_make_string<ResourceString>::value
+      || std::is_convertible<ResourceString, napi_value>::value,
+      "Resource name should be string convertible type");
```
gabrielschulhof: Can we change this to "... should convertible to the string type"?

napi-inl.h
```diff
@@ -3627,6 +3699,324 @@ inline void AsyncWorker::OnWorkComplete(
...
+  return status;
+}
+
+// static
+inline void ThreadSafeFunction::CallJS(napi_env env,
+                                       napi_value jsCallback,
+                                       void* /* context */,
+                                       void* data) {
+  if (env == nullptr && jsCallback == nullptr)
+    return;
```
gabrielschulhof: There should be braces around `return`.

napi-inl.h
```diff
@@ -3627,6 +3699,324 @@ inline void AsyncWorker::OnWorkComplete(
...
+
+  if (data != nullptr) {
+    auto* callbackWrapper = static_cast<CallbackWrapper*>(data);
+    (*callbackWrapper)(env, Function(env, jsCallback));
+    delete callbackWrapper;
+  } else {
```
gabrielschulhof: We should change this to `else if (jsCallback != nullptr)`, because soon it will be possible to have a non-null `env` with a null `jsCallback`.

napi.h
```diff
@@ -1820,6 +1821,193 @@ namespace Napi {
     bool _suppress_destruct;
   };
 
+  class ThreadSafeFunction {
+  public:
+    // Default set
```
gabrielschulhof: What does "Default set" mean?

Also, we should have comments that split the API into functions which can only be called from the main thread, and those that are thread-safe.

napi.h
```diff
@@ -1820,6 +1821,193 @@ namespace Napi {
...
+              typename Finalizer, typename FinalizerDataType>
+    static ThreadSafeFunction New(napi_env env,
+                                  const Function& callback,
+                                  const Object& resource,
+                                  ResourceString resourceName,
+                                  size_t maxQueueSize,
+                                  size_t initialThreadCount,
+                                  ContextType* context,
+                                  Finalizer finalizeCallback,
+                                  FinalizerDataType* data);
```
gabrielschulhof: For consistency we should change `napi_env env` to `Napi::Env env` in the factory functions above.

napi-inl.h
```diff
@@ -125,6 +125,78 @@ struct FinalizeData {
...
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env));
+    if (finalizeData->tsfn) {
+      *finalizeData->tsfn = nullptr;
+    }
+    delete finalizeData;
+  }
+
+  static inline
+  void WrapperWithData(napi_env env,
```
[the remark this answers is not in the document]
romandev: Done

napi-inl.h
```diff
@@ -125,6 +125,78 @@ struct FinalizeData {
...
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env), finalizeData->data);
+    if (finalizeData->tsfn) {
+      *finalizeData->tsfn = nullptr;
+    }
+    delete finalizeData;
+  }
+
+  static inline
+  void WrapperWithContext(napi_env env,
```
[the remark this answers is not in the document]
romandev: Done

napi-inl.h
romandev: Done

napi-inl.h
romandev: Done

napi-inl.h
romandev: Done

napi-inl.h
romandev: Done

napi.h
romandev: > What does "Default set" mean?

Unmeaningful. There are many overloaded methods. So, I was so confusing. To clarify it during development, it's just used for my convenience.

> Also, we should have comments that split the API into functions which can only be called from the main thread, and those that are thread-safe.

Done

napi.h
romandev: Could you let me know more details about the consistency you mentioned?
I'm not sure but the most of New() methods are already using `napi_env` instead of `Napi::Env`.

FYI, please see the following `grep` result.
```
$ grep -Enr "New\((Napi::Env|Env|napi_env)" napi.h

693:    static External New(napi_env env, T* data);
697:    static External New(napi_env env,
702:    static External New(napi_env env,
715:    static Array New(napi_env env);
716:    static Array New(napi_env env, size_t length);
908:    static DataView New(napi_env env,
910:    static DataView New(napi_env env,
913:    static DataView New(napi_env env,
961:    static Function New(napi_env env,
968:    static Function New(napi_env env,
1005:      static Deferred New(napi_env env);
1026:    static Buffer<T> New(napi_env env, size_t length);
1027:    static Buffer<T> New(napi_env env, T* data, size_t length);
1031:    static Buffer<T> New(napi_env env, T* data,
1036:    static Buffer<T> New(napi_env env, T* data,
1285:    static Error New(napi_env env);
1286:    static Error New(napi_env env, const char* message);
1287:    static Error New(napi_env env, const std::string& message);
1312:    static TError New(napi_env env,
1324:    static TypeError New(napi_env env, const char* message);
1325:    static TypeError New(napi_env env, const std::string& message);
1333:    static RangeError New(napi_env env, const char* message);
1334:    static RangeError New(napi_env env, const std::string& message);
1829:    static ThreadSafeFunction New(napi_env env,
1837:    static ThreadSafeFunction New(napi_env env,
1846:    static ThreadSafeFunction New(napi_env env,
1856:    static ThreadSafeFunction New(napi_env env,
1866:    static ThreadSafeFunction New(napi_env env,
1877:    static ThreadSafeFunction New(napi_env env,
1888:    static ThreadSafeFunction New(napi_env env,
1897:    static ThreadSafeFunction New(napi_env env,
1907:    static ThreadSafeF1897:    static ThreadSafeFunction New(napi_env env,
1907:    static ThreadSafeFunction New(napi_env env,
1918:    static ThreadSafeFunction New(napi_env env,
1929:    static ThreadSafeFunction New(napi_env env,
1941:    static ThreadSafeFunction New(napi_env env,
2003:    static ThreadSafeFunction New(napi_env env,
```

gabrielschulhof:
@romandev hmmm ... looks like the code is leaning towards `napi_env` instead of `Napi::Env`. Alright, let's leave it at `napi_env` for now.

napi-inl.h
gabrielschulhof: I mean "... should be convertible to the string type"?

fholzer:
Hi gents,

I'm looking forward to using this TSFN wrapper in my project. While testing this code I encountered an issue in. Sometimes I would get a Segmentation fault (SIGSEGV) or just a core dump (SIGABRT) some time after calling `Release()`. I was able to increase the failure rate by exposing the GC (v8 `--expose-gc` flag), then triggering GC a lot.
I managed to reproduce this with a stripped down version of my code: https://github.com/fholzer/node-addon-api-tsfn
Though with this version the issue only occurred sometimes, usually less then 150 executions are required to trigger it. (My Makefile builds the project, then runs it up to 1000 times until it crashes)
I'm using node v10.16.0 (latest LTS), and yesterday's force push from @romandev.
Finally I ran the same test on the latest version v12.4.0 but I was unable to reproduce it using that version.

Could you please have a look and let me know if this is a potential issue with my code, this TSFN wrapper, or the LTS node version?

Thanks,
Ferdinand

fholzer:
I found that even if node doesn't core dump there's something wrong. That is with both of the mentioned node versions. Here's valgrind output showing invalid writes to free'd blocks: https://gist.github.com/fholzer/03d44449fefb6a9631724a9b4b59b513

Looks like `ThreadSafeFinalize` is trying to write to `std::unique_ptr<napi_threadsafe_function> _tsfn`, even though it has been free'd by the destructor of `ThreadSafeFunction`, which has been GC'd.

gabrielschulhof:
@fholzer thanks for the test case!

napi-inl.h
```diff
@@ -125,6 +125,78 @@ struct FinalizeData {
...
+struct ThreadSafeFinalize {
+  static inline
+  void Wrapper(napi_env env, void* rawFinalizeData, void* /* rawContext */) {
+    if (rawFinalizeData == nullptr)
+      return;
+
+    ThreadSafeFinalize* finalizeData =
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env));
+    if (finalizeData->tsfn) {
```
gabrielschulhof: You're not actually using `finalizeData->tsfn` anywhere in `ThreadSafeFinalize`, so you don't need it at all. Setting it like this is causing the segfault @fholzer found, AFAICT.

napi-inl.h
```diff
@@ -125,6 +125,78 @@ struct FinalizeData {
...
+  void FinalizeWrapperWithData(napi_env env,
+                               void* rawFinalizeData,
+                               void* /* rawContext */) {
+    if (rawFinalizeData == nullptr)
+      return;
+
+    ThreadSafeFinalize* finalizeData =
+        static_cast<ThreadSafeFinalize*>(rawFinalizeData);
+    finalizeData->callback(Env(env), finalizeData->data);
+    if (finalizeData->tsfn) {
```
gabrielschulhof: Ditto. This if-statement can be removed.

napi-inl.h
gabrielschulhof: The structure `ThreadSafeFinalize` makes no use of the thread-safe function, so it doesn't need to store it.

napi-inl.h
```diff
@@ -125,6 +125,78 @@ struct FinalizeData {
...
+        static_cast<ContextType*>(rawContext));
+    if (finalizeData->tsfn) {
+      *finalizeData->tsfn = nullptr;
+    }
+    delete finalizeData;
+  }
+
+  FinalizerDataType* data;
+  Finalizer callback;
+  napi_threadsafe_function* tsfn;
```
gabrielschulhof: You can remove this member.

romandev:
@fholzer, Thank you for finding the bug :) I addressed the bug in latest patch. Could you please check it again?

@gabrielschulhof,
As I mentioned in my previous comment[1], this code is necessary to  reset the internal `tsfn_` value of ThreadSafeFunction automatically.
[1] https://github.com/nodejs/node-addon-api/pull/442#discussion_r285868822

Otherwise, ThreadSafeFunction wrapper should provide `Reset()` to reset the internal value and then users should explicitly call the method in finalize callback.

The problem is caused that ThreadSafeFunction is already destroyed but finalize callback tries to write `nullptr` value to the internal value of destroyed TSFN. I used raw pointer in the FinalizeCallback, so, there was no way to detect whether TSFN is already destroyed. So, I used `std::weak_ptr` instead of raw pointer in latest patch set. It looks to work well.

fholzer:
@romandev  No more use after free :+1:

test/threadsafe_function/threadsafe_function.cc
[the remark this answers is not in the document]
gabrielschulhof: @romandev I also think that, as @KevinEady mentioned above, it's acceptable to throw if one attempts to overwrite a non-empty `ThreadSafeFunction`. In fact, it's probably a good idea, because it'll help users identify leaks.

gabrielschulhof:
@romandev I think if we should throw when one attempts to assign to a non-empty `ThreadSafeFunction`.

romandev:
@gabrielschulhof, In fact, I already did that here: https://github.com/nodejs/node-addon-api/pull/442/files#diff-f546e4cc35b454813e5264fe9c9e6fc8R3898

However, I think we still need a way to reset the `ThreadSafeFunction` to empty in `ThreadSafeFinalizeCallback` if the `ThreadSafeFunction` is aborted or released.

For example, in the current test, declare `ThreadSafeFunction` as a static global variable, and assign a new `ThreadSafeFunction` whenever `StartThreadInternal` is called. If it's non-empty at this time, it would throw an exception. However, if it's empty (by `ThreadSafeFinalizeCallback`), we can re-assign a new object again.

If there was no such logic, we would have to have a static `ThreadSafeFunction` array to create a new `ThreadSafeFunction` whenever `StartThreadInternal` is called. (because there would be no way to re-assign a new object) Or, we might need to make a wrapper object for `ThreadSafeFunction` on heap to maintain creating/destroying the object.

gabrielschulhof:
@romandev a policy of "You cannot assign to a `ThreadSafeFunction` until the existing one has finished operating" makes sense to me.

test/threadsafe_function/threadsafe_function.cc
```diff
@@ -0,0 +1,185 @@
+#include <uv.h>
```
mhdawson: I think we want a big comment here that says this using uv is not guaranteed to be ABI stable.  That if you want to use the stability guarantee from N-API you need to use other means to implement threads.  This is the first example where we have used uv in the testing and we want it to be clear that it is not what we recommend and that we only do it because we know the test suite will be build/run with Node.js as opposed to being built into a binary that will run across different Node.js versions.

gabrielschulhof: Hmmm ... we could use `std::thread` instead, right? I think all our supported compilers now have C++11, right? In the test in core I used libuv because, well, we maintain core ourselves, and because in core I seek to keep N-API tests written in C as much as possible. @romandev could you convert this to `std::thread`?

KevinEady:
FWIW, I used `std::thread` in the example in the docs, so yeah makes sense
to standardize.

gabrielschulhof:
... and I'm actually thinking of using C11's `thrd_create()` in the core N-API test – wanna test though whether it's available on all our supported platforms.

test/threadsafe_function/threadsafe_function.cc
mhdawson: That would be better.

gabrielschulhof:
Well, it [looks like](https://ci.nodejs.org/job/node-test-pull-request/24157/) `#include <threads.h>` doesn't work on all our platforms. Dunno if the STL is in a similar state. Maybe we should hold off on this for now and just bring in the comment from https://github.com/nodejs/node/blob/master/test/node-api/test_threadsafe_function/binding.c#L1-L5

gabrielschulhof:
@addaleax:
> [...] I think C++ `std::thread` is the easier way to go. That’s in C++11 and should be included everywhere.

@romandev can you please move the test over to `std::thread`?

romandev:
> @romandev can you please move the test over to std::thread?

@gabrielschulhof Done. Thanks.

gabrielschulhof:
| Version | Job | Status |
| --- | --- | --- |
| 13 | https://ci.nodejs.org/job/node-test-node-addon-api-new/139/ | ✓ |
| 12 | https://ci.nodejs.org/job/node-test-node-addon-api-new/140/ | ✓ |
| 11 | https://ci.nodejs.org/job/node-test-node-addon-api-new/141/ |  |
| 10 | https://ci.nodejs.org/job/node-test-node-addon-api-new/142/ |  |
| 9 | https://ci.nodejs.org/job/node-test-node-addon-api-new/143/ |  |
| 8 | https://ci.nodejs.org/job/node-test-node-addon-api-new/144/ |  |

romandev:
FYI, I got rid of std::make_unique from latest patch because the make_unique() is C++ 14 feature..

gabrielschulhof:
New CI:

| Version | Job | Status |
|--- |--- |--- |
| 13 | https://ci.nodejs.org/job/node-test-node-addon-api-new/194/ | ✓ |
| 12 | https://ci.nodejs.org/job/node-test-node-addon-api-new/195/ | ✓ |
| 11 | https://ci.nodejs.org/job/node-test-node-addon-api-new/196/ | ✗ |
| 10 | https://ci.nodejs.org/job/node-test-node-addon-api-new/197/ | ✓ |
| 9 | https://ci.nodejs.org/job/node-test-node-addon-api-new/198/ | ✗ |
| 8 | https://ci.nodejs.org/job/node-test-node-addon-api-new/199/ | ✓ |

KevinEady:
Hi @gabrielschulhof , 

Looks like the node-9 tests are failing because the napi tsfn functions don't exist. Were tsfns backported to node 9? Don't see them listed on the [n-api docs](https://nodejs.org/docs/latest-v9.x/api/n-api.html).

gabrielschulhof:
Landed in 0a90df2fcb511c56fdee0d193fa7c8a5b6212e33. Thanks @romandev!

PR #446, opened 2019-02-14
Removed redundant information in basic_types.md

lukechu10:
Removed duplicate information


gabrielschulhof:
@NickNaso I believe you are right. This is not duplicate information.

PR #452, opened 2019-02-25
package: add npm search keywords

sam-github:
It was hard to find this package on npm, even though I knew it was there. With keywords, it should be possible to find it with, for example, https://www.npmjs.com/search?q=napi%20addon


mhdawson:
Landed as b409a2f9877ceeed8f09b783d9bd5e6f45c1baea

PR #458, opened 2019-03-11
doc: fix links to the Property Descriptor docs

okuryu:
I think correct URL is as follows.
https://github.com/nodejs/node-addon-api/blob/master/doc/property_descriptor.md


mhdawson:
Landed as 72b1975cfffbc55a04c64047467d7dd0bfa2934e

PR #459, opened 2019-03-20
Correct return type of Int32Value to int32_t

gallafent:
It currently reads uint32_t, which would be very strange, and is also untrue (I checked napi.h) ;)


mhdawson:
Thanks ! Landed as 1ed7ad87696a54fab9edc91e329cc9cd7ef842b9

PR #460, opened 2019-03-20
Document adding -fvisibility=hidden flag for macOS users

NickNaso:
Added section to remember the macOS user to add the `-fvisibility=hidden` flag.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -55,14 +55,14 @@ To use **N-API** in a native module:
 ```gyp
   'defines': [ 'NAPI_DISABLE_CPP_EXCEPTIONS' ],
 ```
-  4. **Only** for **macOS** user please don't forget to insert these additional 
-  settings in the `binding.gyp` file:
-  
+  4. If you would like your native addon to support OSX, please also add the
+  following settings in the `binding.gyp` file:
+
   ```gyp
   ['OS=="mac"', {
       'cflags+': ['-fvisibility=hidden'],
       'xcode_settings': {
-        'OTHER_CFLAGS': ['-fvisibility=hidden']
+        'GCC_SYMBOLS_PRIVATE_EXTERN': 'YES', # -fvisibility=hidden
       }
   }]
   ```
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -55,8 +55,19 @@ To use **N-API** in a native module:
 ```gyp
   'defines': [ 'NAPI_DISABLE_CPP_EXCEPTIONS' ],
 ```
+  4. **Only** for **macOS** user please don't forget to insert these additional 
+  settings in the `binding.gyp` file:
+  
+  ```gyp
+  ['OS=="mac"', {
+      'cflags+': ['-fvisibility=hidden'],
+      'xcode_settings': {
+        'OTHER_CFLAGS': ['-fvisibility=hidden']
+      }
+  }]
+  ```
 
-  4. Include `napi.h` in the native module code.
+  5. Include `napi.h` in the native module code.
      To ensure only ABI-stable APIs are used, DO NOT include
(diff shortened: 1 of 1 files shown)


doc/setup.md
```diff
@@ -55,8 +55,19 @@ To use **N-API** in a native module:
 ```gyp
   'defines': [ 'NAPI_DISABLE_CPP_EXCEPTIONS' ],
 ```
-
-  4. Include `napi.h` in the native module code.
+  4. **Only** for **macOS** user please don't forget to insert these additional
```
gabrielschulhof: Perhaps you should rephrase this as "If you would like your native addon to support OSX, please also add the following:"

doc/setup.md
```diff
@@ -55,8 +55,19 @@ To use **N-API** in a native module:
...
-
-  4. Include `napi.h` in the native module code.
+  4. **Only** for **macOS** user please don't forget to insert these additional 
+  settings in the `binding.gyp` file:
+  
+  ```gyp
+  ['OS=="mac"', {
+      'cflags+': ['-fvisibility=hidden'],
+      'xcode_settings': {
+        'OTHER_CFLAGS': ['-fvisibility=hidden']
```
gabrielschulhof: Instead of `'OTHER_CFLAGS': ['-fvisibility=hidden']`, you can use `'GCC_SYMBOLS_PRIVATE_EXTERN': 'YES', # -fvisibility=hidden`

vweevers:
@NickNaso @gabrielschulhof Could you summarize why you recommend `-fvisibility=hidden`? Are there any risks / downsides? It seems to stem from https://github.com/nodejs/node-addon-api/pull/456, is that right?

I'm asking because `leveldown` has a similar issue (https://github.com/Level/leveldown/pull/687) and  `-fvisibility=hidden` might solve it but I'm not sure what the implications are for a native addon.

Thanks!

NickNaso:
Hi @vweevers,
what I remember is that on OSX symbols are exported by default, and they **overlap** if two different copies of the same symbol appear in a process. This means that if you build two different native add-ons and export the same symbol on OSX there is a risk to call a function on the wrong native add-on. In our case https://github.com/nodejs/node-addon-api/pull/456 we built two different native add-ons one that support the exception (`binding.node`) and another not (`binding_noexcept.node`) and we discovered that on OSX  loading the native add-ons on the same process it could happens that a function within the wrong native add-on will be called.
@gabrielschulhof reported this to node-gyp maintainers you can follow the status of the duplicate symbol check here on this PR https://github.com/nodejs/node-gyp/pull/1891
I hope that this answer help to you.

mvduin:
Note that `-fvisibility=hidden` should be enabled not just on macOS but also on Linux and every other platform that supports it, which I'm guessing is probably every platform except Windows where this behaviour is default anyway (more or less), not just for node addons but for shared libraries in general.

Even if it happens to not crash, you obviously don't want symbols from one module to unintentionally resolve against the same symbol defined in a different module, that's a recipe for horrifyingly weird problems that only occur when certain combinations of modules are loaded. This is especially true for libraries like node-addon-api which is partially parameterized through preprocessor macros.

Note that the disturbing default behaviour for shared libraries on ELF-based targets (such as Linux) is that every symbol with external linkage is resolved at runtime by the dynamic linker even if that symbol is defined in the same file, i.e. if you write:
```c
void foo() {}
void bar() { foo(); }
```
then `bar` may in fact call `foo` imported from elsewhere rather than the one defined right above it. This also applies to functions declared as `inline` if the compiler chose to not actually inline it, and it prevents inlining of functions that aren't explicitly declared as `inline` (such as inlining `foo` into `bar` above).

As a result, `-fvisibility=hidden` also tends to significiantly reduce the size of and improve the performance of shared libraries.

PR #475, opened 2019-04-29
src: fix ObjectWrap destruction

addaleax:
##### src: call `napi_remove_wrap()` in `ObjectWrap` dtor

Currently, when the `ObjectWrap` constructor runs, it calls
`napi_wrap()`, adding a finalize callback to the freshly created
JS object.

However, if the `ObjectWrap` instance is prematurely deleted,
for example because a subclass constructor throws – which seems
like a reasonable scenario – that finalize callback was not removed,
possibly leading to a use-after-free crash.

This commit adds a call `napi_remove_wrap()` from the `ObjectWrap`
destructor,  and a test for that scenario.

Fixes: https://github.com/node-ffi-napi/weak-napi/issues/16

~~(As you can see, there is a fixup commit attached. I’m not strictly *sure*
whether it’s necessary, but just from looking at the code, it seems that
calling `napi_get_reference_value()` from the finalize callback is invalid;
that seems like a bug in itself, but one that needs to be addressed in
nodejs/node.)~~ See nodejs/node#27470

##### src: make ObjectWrap dtor virtual

Otherwise, subclasses of `ObjectWrap` would not be deleted
correctly by the `delete instance;` line in `FinalizeCallback()`.

(This is also just the right thing to do for classes from which
subclasses are supposed to be created.)

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2702,37 +2702,6 @@ inline Object FunctionReference::New(const std::vector<napi_value>& args) const 
 // CallbackInfo class
 ////////////////////////////////////////////////////////////////////////////////
 
-class ObjectWrapConstructionContext {
- public:
-  ObjectWrapConstructionContext(CallbackInfo* info) {
-    info->_objectWrapConstructionContext = this;
-  }
-
-  static inline void SetObjectWrapped(const CallbackInfo& info) {
-    if (info._objectWrapConstructionContext == nullptr) {
-      Napi::Error::Fatal("ObjectWrapConstructionContext::SetObjectWrapped",
-                         "_objectWrapConstructionContext is NULL");
-    }
-    info._objectWrapConstructionContext->_objectWrapped = true;
-  }
-
-  inline void Cleanup(const CallbackInfo& info) {
-    if (_objectWrapped) {
-      napi_status status = napi_remove_wrap(info.Env(), info.This(), nullptr);
-
-      // There's already a pending exception if we are at this point, so we have
-      // no choice but to fatally fail here.
-      NAPI_FATAL_IF_FAILED(status,
-                           "ObjectWrapConstructionContext::Cleanup",
-                           "Failed to remove wrap from unsuccessfully "
-                           "constructed ObjectWrap instance");
-    }
-  }
-
- private:
-  bool _objectWrapped = false;
-};
-
 inline CallbackInfo::CallbackInfo(napi_env env, napi_callback_info info)
     : _env(env), _info(info), _this(nullptr), _dynamicArgs(nullptr), _data(nullptr) {
   _argc = _staticArgCount;
@@ -3140,13 +3109,22 @@ inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
   status = napi_wrap(env, wrapper, this, FinalizeCallback, nullptr, &ref);
   NAPI_THROW_IF_FAILED_VOID(env, status);
 
-  ObjectWrapConstructionContext::SetObjectWrapped(callbackInfo);
   Reference<Object>* instanceRef = this;
   *instanceRef = Reference<Object>(env, ref);
 }
 
-template<typename T>
-inline ObjectWrap<T>::~ObjectWrap() {}
+template <typename T>
+inline ObjectWrap<T>::~ObjectWrap() {
+  // If the JS object still exists at this point, remove the finalizer added
+  // through `napi_wrap()`.
+  if (!IsEmpty()) {
+    Object object = Value();
+    // It is not valid to call `napi_remove_wrap()` with an empty `object`.
+    // This happens e.g. during garbage collection.
+    if (!object.IsEmpty())
+      napi_remove_wrap(Env(), object, nullptr);
+  }
+}
 
 template<typename T>
 inline T* ObjectWrap<T>::Unwrap(Object wrapper) {
@@ -3716,23 +3694,15 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
 
   napi_value wrapper = details::WrapCallback([&] {
     CallbackInfo callbackInfo(env, info);
-    ObjectWrapConstructionContext constructionContext(&callbackInfo);
 #ifdef NAPI_CPP_EXCEPTIONS
-    try {
-      new T(callbackInfo);
-    } catch (const Error& e) {
-      // Re-throw the error after removing the failed wrap.
-      constructionContext.Cleanup(callbackInfo);
-      throw e;
-    }
+    new T(callbackInfo);
 #else
     T* instance = new T(callbackInfo);
     if (callbackInfo.Env().IsExceptionPending()) {
       // We need to clear the exception so that removing the wrap might work.
       Error e = callbackInfo.Env().GetAndClearPendingException();
-      constructionContext.Cleanup(callbackInfo);
-      e.ThrowAsJavaScriptException();
       delete instance;
+      e.ThrowAsJavaScriptException();
     }
 # endif  // NAPI_CPP_EXCEPTIONS
     return callbackInfo.This();
@@ -3859,7 +3829,7 @@ inline napi_value ObjectWrap<T>::InstanceSetterCallbackWrapper(
 
 template <typename T>
 inline void ObjectWrap<T>::FinalizeCallback(napi_env env, void* data, void* /*hint*/) {
-  T* instance = reinterpret_cast<T*>(data);
+  ObjectWrap<T>* instance = static_cast<ObjectWrap<T>*>(data);
   instance->Finalize(Napi::Env(env));
   delete instance;
 }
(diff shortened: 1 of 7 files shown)


napi-inl.h
```diff
@@ -2806,14 +2806,22 @@ inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
...
-  Reference<Object>* instanceRef = instance;
+  Reference<Object>* instanceRef = this;
   *instanceRef = Reference<Object>(env, ref);
 }
 
+template <typename T>
+inline ObjectWrap<T>::~ObjectWrap() {
+  if (!IsEmpty()) {
+    Object object = Value();
+    if (!object.IsEmpty())
```
digitalinfinity: This code could probably benefit from a comment

addaleax: @digitalinfinity Yup, added a comment!

addaleax:
Looks like this may cause issues in Node.js versions that don’t have https://github.com/nodejs/node/pull/24494 … I’ll try and see if there’s some way to detect/work around that.

mhdawson:
Added `do not land` label so that we don't land until @addaleax figures out the issues mentioned.

Veetaha:
Hey guys, funny enough I encountered the issue anticipated right here. When I throw `Napi::Error` in `ObjectWrap<Subclass>` I experience double free (when the next instance is created right at the same address where the previous one was, but threw an error), munmup() errors and segmentation fault.
When will this be fixed? What workarounds do you propose?
I also noticed that even when `ObjectWrap<Sublcass>` construction finishes successfully it is not guaranteed that the destructor of a created instance will be called before the end of the program.
Any comments on this behavior?

napi-inl.h
```diff
@@ -2806,14 +2806,22 @@ inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
...
   napi_status status;
   napi_ref ref;
-  T* instance = static_cast<T*>(this);
-  status = napi_wrap(env, wrapper, instance, FinalizeCallback, nullptr, &ref);
+  status = napi_wrap(env, wrapper, this, FinalizeCallback, nullptr, &ref);
   NAPI_THROW_IF_FAILED_VOID(env, status);
 
-  Reference<Object>* instanceRef = instance;
+  Reference<Object>* instanceRef = this;
   *instanceRef = Reference<Object>(env, ref);
```
Veetaha: These two statements though work, as intended but seem to be a little bit misleading. Having a fast peek over the code you could think that we re-move-assign the whole *this object.
I suggest making an explicit call to parent move-assignment operator, namely:
```cpp
Reference::operator=({env, ref});
```
It is more laconic and clear from my point of view.

napi-inl.h
```diff
@@ -3329,7 +3337,7 @@ inline napi_value ObjectWrap<T>::InstanceSetterCallbackWrapper(
 
 template <typename T>
 inline void ObjectWrap<T>::FinalizeCallback(napi_env /*env*/, void* data, void* /*hint*/) {
-  T* instance = reinterpret_cast<T*>(data);
+  ObjectWrap<T>* instance = static_cast<ObjectWrap<T>*>(data);
```
Veetaha: I propose using the `auto` keyword here in order to decrease verbosity. Or you could simply rename argument `data` to `instance` and do it in one statement:
```cpp
inline void ObjectWrap<T>::FinalizeCallback(napi_env /*env*/, void * instance, void * /*hint*/) {
    delete static_cast<ObjectWrap<T>*>(instance);
}
```
Any of two options are alright from my viewpoint, though it may just be a matter of style.

davedoesdev:
Also seeing this issue (double call to FinalizeCallback), and this PR fixed the SEGV.

gabrielschulhof:
We addressed the portion whereby an exception in the `ObjectWrap<T>` constructor throws an error in #600.

gabrielschulhof:
@addaleax since we dropped support for v8.x, I believe this PR can go ahead. It may actually be a more generic solution than #600.

addaleax:
@gabrielschulhof I’ll look into rebasing this :+1:

addaleax:
@gabrielschulhof I’ve rebased this, but I don’t quite understand #600 and the added complexity there – the destructor should undo what the constructor did in case that there’s an exception, and I don’t really see why that’s enough.

This also doesn’t pass tests yet.

gabrielschulhof:
@addaleax I think you're right.

gabrielschulhof:
Sorry, wrong button 😳

gabrielschulhof:
@addaleax

```diff
diff --git a/napi-inl.h b/napi-inl.h
index fde0a66..2855e80 100644
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2702,37 +2702,6 @@ inline Object FunctionReference::New(const std::vector<napi_value>& args) const
 // CallbackInfo class
 ////////////////////////////////////////////////////////////////////////////////
 
-class ObjectWrapConstructionContext {
- public:
-  ObjectWrapConstructionContext(CallbackInfo* info) {
-    info->_objectWrapConstructionContext = this;
-  }
-
-  static inline void SetObjectWrapped(const CallbackInfo& info) {
-    if (info._objectWrapConstructionContext == nullptr) {
-      Napi::Error::Fatal("ObjectWrapConstructionContext::SetObjectWrapped",
-                         "_objectWrapConstructionContext is NULL");
-    }
-    info._objectWrapConstructionContext->_objectWrapped = true;
-  }
-
-  inline void Cleanup(const CallbackInfo& info) {
-    if (_objectWrapped) {
-      napi_status status = napi_remove_wrap(info.Env(), info.This(), nullptr);
-
-      // There's already a pending exception if we are at this point, so we have
-      // no choice but to fatally fail here.
-      NAPI_FATAL_IF_FAILED(status,
-                           "ObjectWrapConstructionContext::Cleanup",
-                           "Failed to remove wrap from unsuccessfully "
-                           "constructed ObjectWrap instance");
-    }
-  }
-
- private:
-  bool _objectWrapped = false;
-};
-
 inline CallbackInfo::CallbackInfo(napi_env env, napi_callback_info info)
     : _env(env), _info(info), _this(nullptr), _dynamicArgs(nullptr), _data(nullptr) {
   _argc = _staticArgCount;
@@ -3140,7 +3109,6 @@ inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
   status = napi_wrap(env, wrapper, this, FinalizeCallback, nullptr, &ref);
   NAPI_THROW_IF_FAILED_VOID(env, status);
 
-  ObjectWrapConstructionContext::SetObjectWrapped(callbackInfo);
   Reference<Object>* instanceRef = this;
   *instanceRef = Reference<Object>(env, ref);
 }
@@ -3726,23 +3694,15 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
 
   napi_value wrapper = details::WrapCallback([&] {
     CallbackInfo callbackInfo(env, info);
-    ObjectWrapConstructionContext constructionContext(&callbackInfo);
 #ifdef NAPI_CPP_EXCEPTIONS
-    try {
-      new T(callbackInfo);
-    } catch (const Error& e) {
-      // Re-throw the error after removing the failed wrap.
-      constructionContext.Cleanup(callbackInfo);
-      throw e;
-    }
+    new T(callbackInfo);
 #else
     T* instance = new T(callbackInfo);
     if (callbackInfo.Env().IsExceptionPending()) {
       // We need to clear the exception so that removing the wrap might work.
       Error e = callbackInfo.Env().GetAndClearPendingException();
-      constructionContext.Cleanup(callbackInfo);
-      e.ThrowAsJavaScriptException();
       delete instance;
+      e.ThrowAsJavaScriptException();
     }
 # endif  // NAPI_CPP_EXCEPTIONS
     return callbackInfo.This();
```

this makes the tests pass by basically reverting most of #600, but with one important caveat: The last `+` in the above diff moves the re-throwing of the exception in the non-C++-exception case to **after** deleting the instance, otherwise `napi_remove_wrap()` will encounter a pending exception and fail. Since an exception is already pending, node-addon-api will handle that failure as fatal.

Please feel free to use the diff!

addaleax:
@gabrielschulhof Thanks, done! PTAL

gabrielschulhof:
CI:

| Version | Job | Status |
|---:|---|:---:|
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1637/ |  |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1638/ |  |
| v10.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1639/ |  |

gabrielschulhof:
@addaleax I think you need to rebase so that https://github.com/nodejs/node-addon-api/pull/649 might apply and fix the BigInt tests.

addaleax:
@gabrielschulhof Thanks, done!

gabrielschulhof:
CI:

| Version | Job | Status |
|---:|---|:---:|
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1640/ |  |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1641/ |  |
| v10.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1642/ |  |

gabrielschulhof:
Landed in 4e885069f1721dec1705c898383625dd2907f9e6.

PR #478, opened 2019-04-29
Use IsEmpty to check for empty value

NickNaso:
Use `Value::IsEmpty` instead check for empty value using the following code:
```cpp
if (_value == nullptr) {
   // Do something
}
```


mhdawson:
Landed as `e1cf9a35a1b6edc404c1b465ec94f8a821641059`

Issue #231, opened 2018-03-05
AsyncWorker structural defects and limitations


ebickle:
The current design of napi::AsyncWorker has a number of defects and limitations with it's current design

### Unsafe Self-Destructing Behavior
Managing object lifetimes across multiple threads is a challenging problem. In the case of AsyncWorker, the caller that instantiates a new AsyncWorker instance has no control over when the work will be completed. To work around this problem, the AsyncWorker class currently has a "self-destructing" behavior - the last step of the OnWorkComplete callback function causes the object to delete itself. (`delete self;`). This behavior is unsafe and causes a number of problems:

1. The AsyncWorker class does not know how the AsyncWorker memory was allocated. Imagine the following code as an example:
`
void runWorkOnThread() {
  AsyncWorkerSubclass worker;
  worker.Queue();
}
`
In this case, `new` was never called and the AsyncWorker (unfortunately) lives on the stack.

2. the work_complete callback executes asynchronously from the caller's lifetime of the object. If the caller holds a reference to the AsyncWorker and later calls Cancel() asynchronously from the main node.js event loop, a race condition will occur. Cancellation is always an asynchronous operation and must not fail in a dangerous manner when called after execution of the asynchronous work has already completed. Imagine this scenario:

    * The AsyncWorker is instantiated and queued, a reference to the AsyncWorker pointer is retained.
    * Call flow returns out of the addon and back to node.js.
    * Node.js executes other tasks on the event loop, one of which is the work_complete callback.
    * The AsyncWorker is deleted.
    * User code is executed on the Node.js event loop that calls an operation on the addon that holds the reference to the AsyncWorker and calls Cancel.
    * Invalid memory is referenced - boom. 

### Hard-coded Javascript Function Callbacks
The AsyncWorker currently takes a callback napi::Function and optional napi::Object receiver as required parameters. The intent behind the callback function is to provide an easy mechanism to automatically call a user-defined callback function outside of the addon, but this makes an incorrect assumption that the implementing addon will use a callback design pattern. 

Consider a "legacy-free" addon that wishes to return a Promise for direct use with async and await. The addon would need to implement a Javascript function as a callback, wrap C++ code inside the Javascript Function that marshals their return value and/or errors, then have that C++ code signal a deferred. 

The requirement for a "Javascript round-trip" to execute C++ code or implement Promises is a major headache.

### No Suppose for Promises
As discussed above, there is no way to directly bridge an AsyncWorker to a Promise.

### No Support for Error objects
[...]

PR #489, opened 2019-06-07
AsyncWorker: make callback optional

KevinEady:
`AsyncWorker` assumed that after work is complete,  a JavaScript  callback would need to execute.

This change removes the restriction of specifying a `Function` callback, and instead replaces it with an `Env` parameter. Since the purpose of `receiver` was for the `this` context for the callback, it has also been removed from the constructors.

Re: https://github.com/nodejs/node-addon-api/issues/231#issuecomment-483356091

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3522,6 +3522,32 @@ inline AsyncWorker::AsyncWorker(const Object& receiver,
   NAPI_THROW_IF_FAILED_VOID(_env, status);
 }
 
+inline AsyncWorker::AsyncWorker(Napi::Env env)
+  : AsyncWorker(env, "generic") {
+}
+
+inline AsyncWorker::AsyncWorker(Napi::Env env,
+                                const char* resource_name)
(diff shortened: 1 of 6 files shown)


napi-inl.h
```diff
@@ -3522,6 +3522,32 @@ inline AsyncWorker::AsyncWorker(const Object& receiver,
   NAPI_THROW_IF_FAILED_VOID(_env, status);
 }
 
+inline AsyncWorker::AsyncWorker(napi_env env)
```
mhdawson: These should be `Napi::Env env` as opposed to `napi_env env` for consistency with the rest of the API.  Same goes for other places env is being added.

KevinEady: Thanks for the tip. I was a little confused because several places use `napi_env` but maybe those are older pieces of code. Changed now to `Napi::Env`

mhdawson:
Landed as b2b08122ea72ef4bb42e0dc99c0e50b7f4cf2f91

PR #493, opened 2019-06-13
Use curly brackets to  include node_api.h

NickNaso:
In this PR I include `node_api.h` file using curly brackets because when native addon are building we have `node_api.h` in our include path.


mhdawson:
Landed as 1fb540eeb52db660fbafd8d064754708c1569b0b

PR #494, opened 2019-06-13
doc: document ThreadSafeFunction

KevinEady:
Re: https://github.com/nodejs/node-addon-api/pull/442

### DIFF
--- a/doc/threadsafe_function.md
+++ b/doc/threadsafe_function.md
@@ -1,0 +1,219 @@
+# ThreadSafeFunction
+
+JavaScript functions can normally only be called from a native addon's main thread. If an addon creates additional threads, then node-addon-api functions that require a `Env`, `Value`, or `Ref` must not be called from those threads.
+
+When an addon has additional threads and JavaScript functions need to be invoked based on the processing completed by those threads, those threads must communicate with the addon's main thread so that the main thread can invoke the JavaScript function on their behalf. The thread-safe function APIs provide an easy way to do this.
+
+These APIs provide the type `Napi::ThreadSafeFunction` as well as APIs to create, destroy, and call objects of this type. `Napi::ThreadSafeFunction::New()` creates a persistent reference that holds a JavaScript function which can be called from multiple threads. The calls happen asynchronously. This means that values with which the JavaScript callback is to be called will be placed in a queue, and, for each value in the queue, a call will eventually be made to the JavaScript function.
+
+`ThreadSafeFunction` objects are destroyed when every thread which uses the object has called `Release()` or has received a return status of `CLOSING` in response to a call to `BlockingCall()` or `NonBlockingCall()`. The queue is emptied before the `ThreadSafeFunction` is destroyed. It is important that `Release()` be the last API call made in conjunction with a given `ThreadSafeFunction`, because after the call completes, there is no guarantee that the `ThreadSafeFunction` is still allocated. For the same reason it is also important that no more use be made of a thread-safe function after receiving a return value of `CLOSING` in response to a call to `BlockingCall()` or `NonBlockingCall()`. Data associated with the `ThreadSafeFunction` can be freed in its `Finalizer` callback which was passed to `ThreadSafeFunction::New()`.
+
+Once the number of threads making use of a `ThreadSafeFunction` reaches zero, no further threads can start making use of it by calling `Acquire()`. In fact, all subsequent API calls associated with it, except `Release()`, will return an error value of napi_closing.
+
+## Methods
+
+### Constructor
+
+Creates a new empty instance of `Napi::ThreadSafeFunction`.
+
+```cpp
+Napi::Function::ThreadSafeFunction();
+```
+
+### Constructor
+
+Creates a new instance of the `Napi::ThreadSafeFunction` object.
+
+```cpp
+Napi::ThreadSafeFunction::ThreadSafeFunction(napi_threadsafe_function tsfn);
+```
+
+- `[in] value`: The `napi_threadsafe_function` which is a handle for an existing napi threadsafe function.
+
+Returns a non-empty `Napi::ThreadSafeFunction` instance.
+
+### New
+
+Creates a new instance of the `Napi::ThreadSafeFunction` object. The `New` function has several overloads for the various optional parameters: skip the optional parameter for that specific overload.
+
+```cpp
+New(napi_env env,
+    const Function& callback,
+    const Object& resource,
+    ResourceString resourceName,
+    size_t maxQueueSize,
+    size_t initialThreadCount,
+    DataType* data,
+    Finalizer finalizeCallback,
+    Context* context);
+```
+
+- `env`: The `napi_env` environment in which to construct the `Napi::ThreadSafeFunction` object.
+- `callback`: The `Function` to call from another thread.
+- `[optional] resource`: An object associated with the async work that will be passed to possible async_hooks init hooks.
+- `resourceName`: A JavaScript string to provide an identifier for the kind of resource that is being provided for diagnostic information exposed by the async_hooks API.
+- `maxQueueSize`: Maximum size of the queue. `0` for no limit.
+- `initialThreadCount`: The initial number of threads, including the main thread, which will be making use of this function.
+- `[optional] data`: Data to be passed to `finalizeCallback`.
+- `[optional] finalizeCallback`: Function to call when the `ThreadSafeFunction` is being destroyed.  This callback will be invoked on the main thread when the thread-safe function is about to be destroyed. It receives the context and the finalize data given during construction (if given), and provides an opportunity for cleaning up after the threads e.g. by calling `uv_thread_join()`. It is important that, aside from the main loop thread, there be no threads left using the thread-safe function after the finalize callback completes. Must implement `void operator()(Env env, DataType* data, Context* hint)`, skipping `data` or `hint` if they are not provided.
+- `[optional] context`: Data to attach to the resulting `ThreadSafeFunction`. Can be retreived via `GetContext()`.
+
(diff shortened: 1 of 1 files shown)


doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
...
+- `[in] value`: The `napi_threadsafe_function` which is a handle for an existing napi threadsafe function.
+
+Returns a non-empty `Napi::ThreadSafeFunction` instance.
+
+### New
+
+Creates a new instance of the `Napi::ThreadSafeFunction` object. The `New` function has several overloads for the various optional parameters: skip the optional parameter for that specific overload.
+
+```cpp
+New(napi_env env,
```
gabrielschulhof: We should change this to `Napi::Env env` - [commented in the PR](https://github.com/nodejs/node-addon-api/pull/442/files#r293508141) as well.

doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
...
+    const Object& resource,
+    ResourceString resourceName,
+    size_t maxQueueSize,
+    size_t initialThreadCount,
+    DataType* data,
+    Finalizer finalizeCallback,
+    Context* context);
+```
+
+- `env`: The `napi_env` environment in which to construct the `Napi::ThreadSafeFunction` object.
```
gabrielschulhof: Ditto wrt. changing to `Napi::Env`.

doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Acquire
+
+Add a thread to this thread-safe function object, indicating that a new thread will start making use of the thread-safe function. 
+
+```cpp
+bool Napi::ThreadSafeFunction::Acquire()
+```
+
+Returns `true` if the thread can successfully be added to the thread count.
```
gabrielschulhof: Maybe change this to `... if the thread has successfully acquired the thread-safe function for its use.` The implication that a successful addition to the thread count implies that the thread may henceforth use this thread-safe function is not necessarily obvious.

doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Release
+
+Indicate that an existing thread will stop making use of the thread-safe function. A thread should call this API when it stops making use of this thread-safe function. Using any thread-safe APIs after having called this API has undefined results in the current thread, as it may have been destroyed.
+
+```cpp
+bool Napi::ThreadSafeFunction::Release()
+```
+
+Returns `true` if the thread-safe function has successfully released.
```
gabrielschulhof: "... has been successfully released.",  maybe?

doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Abort
+
+"Abort" the thread-safe function. This will cause all subsequent APIs associated with the thread-safe function except `Release()` to return `napi_closing` even before its reference count reaches zero. In particular, `BlockingCall` and `NonBlockingCall()` will return `napi_closing`, thus informing the threads that it is no longer possible to make asynchronous calls to the thread-safe function. This can be used as a criterion for terminating the thread. Upon receiving a return value of `napi_closing` from a threadsafe function call a thread must make no further use of the thread-safe function because it is no longer guaranteed to be allocated.
```
gabrielschulhof: "&#96;napi_closing&#96;" should be "&#96;CLOSING&#96;".

doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
...
+
+```cpp
+napi_status Napi::ThreadSafeFunction::Abort()
+```
+
+Returns `napi_ok` if the thread-safe function has successfully aborted.
+
+### BlockingCall / NonBlockingCall
+
+Calls the Javascript function in a either a blocking or non-blocking fashion.
```
gabrielschulhof: "... in either a blocking ... "

doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
...
+- `BlockingCall()`: the API blocks until space becomes available in the queue. Will never block if the thread-safe function was created with a maximum queue size of `0`.
+- `NonBlockingCall()`: will return `napi_queue_full` if the queue was full, preventing data from being successfully added to the queue
```
gabrielschulhof: Period at the end of the sentence.

doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
+# ThreadSafeFunction
+
+JavaScript functions can normally only be called from a native addon's main thread. If an addon creates additional threads, then node-addon-api functions that require a `Env`, `Value`, or `Ref` must not be called from those threads.
```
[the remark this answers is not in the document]
KevinEady: Added namespace of `Napi`, changed `Ref` to `Reference`

doc/threadsafe_function.md
```diff
@@ -0,0 +1,219 @@
...
+
+`ThreadSafeFunction` objects are destroyed when every thread which uses the object has called `Release()` or has received a return status of `CLOSING` in response to a call to `BlockingCall()` or `NonBlockingCall()`. The queue is emptied before the `ThreadSafeFunction` is destroyed. It is important that `Release()` be the last API call made in conjunction with a given `ThreadSafeFunction`, because after the call completes, there is no guarantee that the `ThreadSafeFunction` is still allocated. For the same reason it is also important that no more use be made of a thread-safe function after receiving a return value of `CLOSING` in response to a call to `BlockingCall()` or `NonBlockingCall()`. Data associated with the `ThreadSafeFunction` can be freed in its `Finalizer` callback which was passed to `ThreadSafeFunction::New()`.
+
+Once the number of threads making use of a `ThreadSafeFunction` reaches zero, no further threads can start making use of it by calling `Acquire()`. In fact, all subsequent API calls associated with it, except `Release()`, will return an error value of napi_closing.
```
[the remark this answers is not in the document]
KevinEady: Removed all old references of the old implementation's `ThreadSafeFunction::Status` to the various `napi_status`es

doc/threadsafe_function.md
KevinEady: As discussed in the PR, we decided to stick with `napi_env`. Therefore, this was not changed.

doc/threadsafe_function.md
KevinEady: The signature for `Acquire()`, `Release()`, `Abort()`, and `BlockingCall()/NonBlockingCall()`  was changed to return an `napi_status`. This documentation has been updated for those return values.

doc/threadsafe_function.md
KevinEady: Addressed; likewise with the signature change mentioned above.

doc/threadsafe_function.md
KevinEady: `napi_closing` is correct as `CLOSING` was the previous `ThreadSafeFunction::Status` enum, which is no longer used.

doc/threadsafe_function.md
KevinEady: Addressed.

doc/threadsafe_function.md
KevinEady: Addressed.

doc/threadsafe_function.md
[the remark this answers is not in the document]
KevinEady: Lines have been wrapped to 80 characters.

KevinEady:
@gabrielschulhof / @romandev ,

Most recent commit addresses the issues brought up, with signatures now matching the approved #442  PR. Please review.

doc/threadsafe_function.md
```diff
@@ -0,0 +1,303 @@
...
+JavaScript function which can be called from multiple threads. The calls happen
+asynchronously. This means that values with which the JavaScript callback is to
+be called will be placed in a queue, and, for each value in the queue, a call
+will eventually be made to the JavaScript function.
+
+`Napi::ThreadSafeFunction` objects are destroyed when every thread which uses
+the object has called `Release()` or has received a return status of
+`napi_closing` in response to a call to `BlockingCall()` or `NonBlockingCall()`.
+The queue is emptied before the `Napi::ThreadSafeFunction` is destroyed. It is
+important that `Release()` be the last API call made in conjunction with a given
```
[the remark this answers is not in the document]
KevinEady: One thing (even with the current wording, generally taken from [n_api_asynchronous_thread_safe_function_calls](https://nodejs.org/api/n-api.html#n_api_asynchronous_thread_safe_function_calls)): I don't think the text is correct, because `Release()` [or `Abort()`] don't _necessarily_ have to be the last call. 

In a normal, non-aborted flow, `Release()` would be the last call, but in an aborted flow, it is possible that a `[Non]BlockingCall()` could be the last call, as it also decreases the thread count:

>  This can be used as a criterion for terminating the thread.

Should we better specify the abortable behavior here as well, or is it fine to keep in the Abort section?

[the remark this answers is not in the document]
gabrielschulhof: `Release()` or `Abort()` *absolutely* have to be the last call – on the thread making the call. I think the most important thing to point out, and the aspect to emphasize here is this: After you've called `Release()` or `Abort()`, make no further use of the TSFN *on that thread*, irrespective of how this will affect the other threads – namely that, in the case of `Abort()` it will cause them to receive `napi_closing`s.

This latter point is documented further down in the paragraph, and `BlockingCall()` and `NonBlockingCall()` are documented to have this feature such that the two calls are presented side-by-side as equal and alternative ways of being informed of the TSFN's impending closure.

The fact is that the documentation in core is almost certainly not phrased ideally. This is actually an opportunity to improve both places of documentation.

mhdawson:
Just noticed that I think this might need a link on the main page: https://github.com/nodejs/node-addon-api

Issue #485, opened 2019-05-22
Object.getOwnPropertyDescriptor throws on accessor properties of an ObjectWrap class' prototype object


argv-minus-one:
An error is thrown when trying to get the property descriptor of a `Napi::ObjectWrap::InstanceAccessor` from the class' prototype object.

Suppose you have a class named `DemoClass`, which was defined by `Napi::ObjectWrap::DefineClass`. Suppose it has a property named `hello`, which was defined by `Napi::ObjectWrap::InstanceAccessor`. If you then try to `Object.getOwnPropertyDescriptor(DemoClass.prototype, "hello")`, you get an error:

```
run.js:2
Object.getOwnPropertyDescriptor(DemoClass.prototype, "hello");
       ^

Error: Invalid argument
    at Function.getOwnPropertyDescriptor (<anonymous>)
    at Object.<anonymous> (run.js:2:8)
    at Module._compile (internal/modules/cjs/loader.js:738:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:749:10)
    at Module.load (internal/modules/cjs/loader.js:630:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:570:12)
    at Function.Module._load (internal/modules/cjs/loader.js:562:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:801:12)
    at internal/main/run_main_module.js:21:11
```

The reason is that `Napi::ObjectWrap::InstanceGetterCallbackWrapper` attempts to unwrap `this`, which doesn't work because `DemoClass.prototype` is not a wrapper!

But why is the callback even being called? Node checks whether native instance methods are called on a valid instance of the class, and doesn't even call the native function if not. Shouldn't it perform the same checks on native instance accessors?

[I have prepared a Gist demonstrating the problem.](https://gist.github.com/argv-minus-one/122179c71b1b0048c177f153c5fa434e) To use:
1. Download or `git clone` it.
2. `npm install`
3. `node run.js`

mhdawson:
To me it seems more that maybe InstanceGetterCallbackWrapper should handle being called when Unwrap fails or that it should check in that is has been called on a wrapper/

I don't think we thought anybody would ask for the property or descriptor on the prototype, it's just there so that it is inherited when instances are created for Wrapped objects.

I think we've effectively told the runtime that the InstnaceGEtterCallbackWrapper is a property on the prototype so unless I'm missing something it's reasonable for them to be called.

[...]

PR #495, opened 2019-06-13
src: fix objectwrap test case

mhdawson:
Refs: https://github.com/nodejs/node-addon-api/issues/485

The test case was relyingon the ordering of
"for in" which is not guarranteed as per
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

Update the testcase to check in an way that does
not depend on ordering.

### DIFF
--- a/test/objectwrap.js
+++ b/test/objectwrap.js
@@ -73,12 +73,11 @@ const test = (binding) => {
         keys.push(key);
       }
 
-      assert.deepEqual(keys, [
-        'testGetSet',
-        'testGetter',
-        'testValue',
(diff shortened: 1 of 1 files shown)


mhdawson:
@nodejs/abi-stable-node could I get an expedited review as it should land before we land https://github.com/nodejs/node/pull/27851

mhdawson:
sorry @nodejs/n-api

mhdawson:
Landed as ab7d8fcc48f5727961a0f09442d6ae4c5f9b5a17

PR #512, opened 2019-07-16
AsyncWorker: add GetResult() method

KevinEady:
Adds an overridable `GetResult()` method, providing arguments to the callback
invoked in `OnOK()`.

Re: https://github.com/nodejs/node-addon-api/issues/231#issuecomment-511504055

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3704,8 +3704,9 @@ inline std::vector<napi_value> AsyncWorker::GetResult(Napi::Env env) {
   return {};
 }
 // The OnExecute method receives an napi_env argument. However, do NOT
-// use it within this method, as it does not run on the main thread and cannot
-// access the napi_env.
+// use it within this method, as it does not run on the main thread and must
+// not run any method that would cause JavaScript to run. In practice, this
+// means that almost any use of napi_env will be incorrect.
 inline void AsyncWorker::OnExecute(napi_env /*DO_NOT_USE*/, void* this_pointer) {
   AsyncWorker* self = static_cast<AsyncWorker*>(this_pointer);
 #ifdef NAPI_CPP_EXCEPTIONS
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3686,7 +3686,7 @@ inline void AsyncWorker::SuppressDestruct() {
 
 inline void AsyncWorker::OnOK() {
   if (!_callback.IsEmpty()) {
-    _callback.Call(_receiver.Value(), std::initializer_list<napi_value>{});
+    _callback.Call(_receiver.Value(), GetResult(_callback.Env()));
   }
 }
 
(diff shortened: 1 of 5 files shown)


KevinEady:
Hey @mhdawson / @gabrielschulhof ,

This is the implementation of the `GetResult()` method we discussed in yesterday's meeting. There are two things that I did change
1. The method takes an `Napi::Env`. I thought it would be beneficial to provide this, so the implementer can create the arguments (since the callback and receiver are private, cannot get env from those)
2. Instead of an `std::initializer_list`, I used an `std::vector`. When using the initializer list, there were issues with scoping / binding the initializer list to a temporary array (if doing something eg `return { Boolean::New(env, true) }`). I think the vector implementation should be fine, since `Function::Call` already had a matching prototype. 

Let me know what you guys think :+1:

napi.h
```diff
@@ -1812,6 +1812,7 @@ namespace Napi {
     virtual void OnOK();
     virtual void OnError(const Error& e);
     virtual void Destroy();
+    virtual std::vector<napi_value> GetResult(Napi::Env env);
```
mhdawson: Adding Napi::Env here as you have done makes sense and keep it consistent with the OnError and OnOk methods.

napi-inl.h
```diff
@@ -3700,7 +3700,13 @@ inline void AsyncWorker::SetError(const std::string& error) {
   _error = error;
 }
 
-inline void AsyncWorker::OnExecute(napi_env /*env*/, void* this_pointer) {
+inline std::vector<napi_value> AsyncWorker::GetResult(Napi::Env env) {
+  return {};
+}
+// The OnExecute method receives an napi_env argument. However, do NOT
+// use it within this method, as it does not run on the main thread and cannot
+// access the napi_env.
```
mhdawson: Maybe. 

 as it does not run on the main thread and must not run any method that would cause
JavaScript to run. In practice, this means that almost any use of napi_env will be incorrect.

test/asyncworker.cc
```diff
@@ -9,14 +9,25 @@ class TestWorker : public AsyncWorker {
...
 
     TestWorker* worker = new TestWorker(cb, "TestResource", resource);
     worker->Receiver().Set("data", data);
     worker->_succeed = succeed;
+    worker->_setResult = setResult;
     worker->Queue();
   }
 
 protected:
+  std::vector<napi_value> GetResult(Napi::Env env) override {
```
mhdawson: I think it would be good to keep some tests that validate the default case.  Maybe we can add TestWorkerNoReturnValues and use that in 1 or more tests to validate the case were setResult is not overridden.

napi-inl.h
KevinEady: Updated.

test/asyncworker.cc
KevinEady: So I reverted back to the old test cases, and added an `AsyncWorkerWithResult` that does override the `GetResult()` method. I think this suffices your concern?

mhdawson: Perfect.

mhdawson:
@gabrielschulhof just want your input on the change to an array as well.

mhdawson:
Landed as 717c9ab163dba35dee3fe4bc0e6da769eedd91ae

PR #527, opened 2019-08-21
Update prebuild_tools.md

anurbol:
There are three tools, not two. And the list may be appended to, who knows


mhdawson:
Landed as 7b1ee96d5232d250e4f764a1ba2720576d357b59

Issue #473, opened 2019-04-20
Nan:AsyncProgressWorker equivalent?


JipBoesenkool:
@jasongin could you reply to @ptgolden's  question? I can find neither node-addon-api or n-api functions to implement `streaming worker` such as  [Nan::AsyncProgressWorker](https://github.com/nodejs/nan/blob/master/doc/asyncworker.md#api_nan_async_progress_worker)

There is example of usage from a book: [factorization example | https://github.com/freezer333/nodecpp-demo/blob/master/streaming/examples/factorization/factorization.js]. 

In my case I need to hook some system events and pass them to js callback.

_Originally posted by @gorshkov-leonid in https://github.com/nodejs/node-addon-api/issues/75#issuecomment-441468273_

JipBoesenkool:
I have exactly the same use case as @gorshkob-leanid

mhdawson:
@JipBoesenkool have you looked at the https://nodejs.org/api/n-api.html#n_api_asynchronous_thread_safe_function_calls as an option?  Not sure if that will work for your use case but worth a look.

gorshkov-leonid:
I tried napi and it, it works. But it looks redundantly. Nan API is much more convenient in this case.

gorshkov-leonid:
To clarify case, I use system hook with own callback. So I need to pass values from this callback to js side. The code might be not pretty, because I have to recall(=learn) c++ basis after ages of oblivion:

https://github.com/gorshkov-leonid/node-native-addons-example/blob/master/native-addon-get-processes/WindowListFunctions.cc

PR #529, opened 2019-08-27
Implement AsyncProgressWorker

legendecas:
Related: https://github.com/nodejs/node-addon-api/issues/473

- [x] npm test passed
- [x] test included
- [x] documents are added or changed

### DIFF
--- a/doc/async_progress_worker.md
+++ b/doc/async_progress_worker.md
@@ -1,6 +1,6 @@
 # AsyncProgressWorker
 
-`Napi::AsyncProgressWorker` is an abstract class, which implements `Napi::AsyncWorker`
+`Napi::AsyncProgressWorker` is an abstract class which implements `Napi::AsyncWorker`
 while extending `Napi::AsyncWorker` internally with `Napi::ThreadSafeFunction` for
 moving work progress reports from worker thread(s) to event loop threads.
 
@@ -19,7 +19,7 @@ For the most basic use, only the `Napi::AsyncProgressWorker::Execute` and
 
 ## Methods
 
-[`Napi::AsyncWorker`]() provides detailed descriptions for most methods.
+[`Napi::AsyncWorker`][] provides detailed descriptions for most methods.
 
 ### Execute
 
@@ -149,7 +149,7 @@ Creates a new `Napi::AsyncProgressWorker`.
 explicit Napi::AsyncProgressWorker(const Napi::Object& receiver, const Napi::Function& callback, const char* resource_name, const Napi::Object& resource);
 ```
 
-- `[in] receiver`: The `this` object passed to the called function.
+- `[in] receiver`: The `this` object to be passed to the called function.
 - `[in] callback`: The function which will be called when an asynchronous
 operations ends. The given function is called from the main event loop thread.
 - `[in] resource_name`:  Null-terminated string that represents the
@@ -253,11 +253,11 @@ void Napi::AsyncProgressWorker::ExecutionProcess::Send(const T* data, size_t cou
 
 The first step to use the `Napi::AsyncProgressWorker` class is to create a new class that
 inherits from it and implement the `Napi::AsyncProgressWorker::Execute` abstract method.
-Typically input to your worker will be saved within the class' fields generally
+Typically input to the worker will be saved within the class' fields generally
 passed in through its constructor.
 
 During the worker thread execution, the first argument of `Napi::AsyncProgressWorker::Execute`
-can be used to report the process of the execution.
+can be used to report the progress of the execution.
 
 When the `Napi::AsyncProgressWorker::Execute` method completes without errors the
 `Napi::AsyncProgressWorker::OnOK` function callback will be invoked. In this function the
@@ -269,7 +269,7 @@ function runs in the background out of the **event loop** thread and at the end
 the `Napi::AsyncProgressWorker::OnOK` or `Napi::AsyncProgressWorker::OnError` function will be
 called and are executed as part of the event loop.
 
-The code below shows a basic example of `Napi::AsyncProgressWorker` the implementation:
+The code below shows a basic example of the `Napi::AsyncProgressWorker` implementation:
 
 ```cpp
 #include<napi.h>
@@ -309,16 +309,16 @@ class EchoWorker : public AsyncProgressWorker<uint32_t> {
 };
 ```
 
-The `EchoWorker`'s contructor calls the base class' constructor to pass in the
+The `EchoWorker`'s constructor calls the base class' constructor to pass in the
 callback that the `Napi::AsyncProgressWorker` base class will store persistently. When
 the work on the `Napi::AsyncProgressWorker::Execute` method is done the
-`Napi::AsyncProgressWorker::OnOk` method is called and the results return back to
-JavaScript invoking the stored callback with its associated environment.
+`Napi::AsyncProgressWorker::OnOk` method is called and the results are return back to
+JavaScript when the stored callback is invoked with its associated environment.
 
 The following code shows an example of how to create and use an `Napi::AsyncProgressWorker`
 
 ```cpp
-#include<napi.h>
+#include <napi.h>
 
 // Include EchoWorker class
 // ..
@@ -326,7 +326,7 @@ The following code shows an example of how to create and use an `Napi::AsyncProg
 use namespace Napi;
 
 Value Echo(const CallbackInfo& info) {
-    // You need to validate the arguments here
+    // We need to validate the arguments here
     Function cb = info[1].As<Function>();
     std::string in = info[0].As<String>();
     EchoWorker* wk = new EchoWorker(cb, in);
@@ -335,10 +335,10 @@ Value Echo(const CallbackInfo& info) {
 }
 ```
 
-Using the implementation of a `Napi::AsyncProgressWorker` is straight forward. You only
-need to create a new instance and pass to its constructor the callback you want to
-execute when your asynchronous task ends and other data you need for your
-computation. Once created the only other action you have to do is to call the
-`Napi::AsyncProgressWorker::Queue` method that will queue the created worker for execution.
+The implementation of a `Napi::AsyncProgressWorker` can be used by creating a
+new instance and passing to its constructor the callback to execute when the
+asynchronous task ends and other data needed for the computation. Once created,
+the only other action needed is to call the `Napi::AsyncProgressWorker::Queue`
+method that will queue the created worker for execution.
 
 [`Napi::AsyncWorker`]: ./async_worker.md
--- a/doc/async_progress_worker.md
+++ b/doc/async_progress_worker.md
@@ -1,0 +1,344 @@
+# AsyncProgressWorker
+
+`Napi::AsyncProgressWorker` is an abstract class, which implements `Napi::AsyncWorker`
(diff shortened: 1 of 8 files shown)


napi.h
```diff
@@ -2032,6 +2033,40 @@ namespace Napi {
...
+        friend class AsyncProgressWorker;
+       public:
+        void Signal() const;
+        void Send(const T* data, size_t count) const;
+       private:
+        explicit ExecutionProgress(AsyncProgressWorker* that) : that_(that) {}
+        AsyncProgressWorker* const that_;
+     };
+     virtual void Execute(const ExecutionProgress& progress) = 0;
+     virtual void OnProgress(const T* data, size_t count) = 0;
```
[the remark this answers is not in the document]
legendecas: I was expecting to keep a consistent naming style with Napi::AsyncWorker like `OnOk` and `OnError`, which in Nan has name `HandleOKCallback` and `HandleErrorCallback`.

test/asyncprogressworker.cc
```diff
@@ -0,0 +1,69 @@
...
+    worker->Queue();
+  }
+
+protected:
+  void Execute(const ExecutionProgress& progress) override {
+    progress.Send(new ProgressData{0}, 1);
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    progress.Send(new ProgressData{50}, 1);
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    if (!_succeed) {
```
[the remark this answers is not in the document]
legendecas: It's used for testing errors. I haven't added test for it yet, like cases covering errored work with continued progress reporting.

legendecas:
> @legendecas can you take a look, seems to be failing in CI runs across platforms with:
> 
> ```
> All tests passed!
> terminate called after throwing an instance of 'Napi::Error'
>   what():  The expression evaluated to a falsy value:
> 
>   assert.ok(e instanceof Error)
> 
> Tests aborted with SIGABRT
> npm ERR! Test failed.  See above for more details.
> ```

Yes, it's crashed for case https://github.com/nodejs/node-addon-api/blob/master/test/asyncworker.js. It could be reproduced locally.

legendecas:
> One other thing that you'll need to add is docs in the "Async Operations" section of https://github.com/nodejs/node-addon-api

Added to checklist

napi-inl.h
```diff
@@ -4068,6 +4070,80 @@ inline void ThreadSafeFunction::CallJS(napi_env env,
...
+
+////////////////////////////////////////////////////////////////////////////////
+// Async Progress Worker class
+////////////////////////////////////////////////////////////////////////////////
+
+template<class T>
+inline AsyncProgressWorker<T>::AsyncProgressWorker(const Function& callback,
+                                                const char* resource_name,
+                                                const Object& resource)
+                           : AsyncWorker(callback, resource_name, resource), _asyncdata(NULL), _asyncsize(0) {
```
NickNaso: Maybe it' s better to use `nullptr` instead of `NULL`.

napi-inl.h
```diff
@@ -4068,6 +4070,80 @@ inline void ThreadSafeFunction::CallJS(napi_env env,
...
+inline void AsyncProgressWorker<T>::WorkProgress_(Napi::Env /* env */, Napi::Function /* jsCallback */, void* _data) {
+  AsyncProgressWorker* self = static_cast<AsyncProgressWorker*>(_data);
+
+  T* data;
+  size_t size;
+  {
+    std::lock_guard<std::mutex> lock(self->_mutex);
+    data = self->_asyncdata;
+    size = self->_asyncsize;
+    self->_asyncdata = NULL;
```
NickNaso: Maybe it' s better to use nullptr instead of NULL.

napi.h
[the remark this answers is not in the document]
NickNaso: My opinion is that we need to support equivalent features provided by `nan` but we need to be consistent with the style used  on `Napi::AsyncWorker.`

test/asyncprogressworker.js
```diff
@@ -0,0 +1,36 @@
...
+function success(binding) {
+  let progress = 0
+  binding.asyncprogressworker.doWork(true, {}, (err, _progress) => {
+    if (err) {
+      assert.fail(err);
+      return;
+    }
+    progress = _progress;
+  });
+  process.on('exit', () => {
```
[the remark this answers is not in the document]
legendecas: I was trying to add this case. But one problem is AsyncProgressWorker doesn't save every progress to call into JS. It could be overridden by design. So there is no guarantee on exact callback. 

Side note: AsyncProgressQueueWorker would save every progress to callback.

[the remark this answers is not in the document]
mhdawson: Ok, understood.  Is there any way to add to the test so that it validates that we were called with one of the expected values?  Right now we validate that the callback was called, but nothing to make sure the parameter being passed is called, right?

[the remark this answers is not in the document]
legendecas: We could synchronize between the main thread and the worker, make sure each callback was successfully invoked. I'd add this case ASAP.

legendecas:
#532 would blocking this PR from their routine work like CI/tests. (Just a note)

test/asyncprogressworker.js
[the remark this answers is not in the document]
legendecas: I've updated the test case to ensure each process report should be handled properly in JS land :)

mhdawson:
Was a bit worried that some of the locking features might have requuired later compilers.  This CI run seems to confirm it can build/run on Node.js https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/675/

mhdawson:
@legendecas sorry it's taking so long to get the review done. I'm out of time for today but I'll try to continue reviewing next week.

napi.h
```diff
@@ -2061,6 +2062,65 @@ namespace Napi {
...
+     virtual ~AsyncProgressWorker();
+
+     class ExecutionProgress {
+        friend class AsyncProgressWorker;
+       public:
+        void Signal() const;
+        void Send(const T* data, size_t count) const;
+       private:
+        explicit ExecutionProgress(AsyncProgressWorker* that) : _that(that) {}
+        AsyncProgressWorker* const _that;
```
mhdawson: Can you come up with a better name that `_that`

napi.h
```diff
@@ -2061,6 +2062,65 @@ namespace Napi {
     std::unique_ptr<napi_threadsafe_function, Deleter> _tsfn;
     Deleter _d;
   };
+
+  template<class T>
+  class AsyncProgressWorker : public AsyncWorker {
+    public:
+     virtual ~AsyncProgressWorker();
+
+     class ExecutionProgress {
```
mhdawson: I assume this separate class is used so that the Execute method only has access to the methods which are safe to call right?

napi-inl.h
```diff
@@ -4112,6 +4114,156 @@ inline void ThreadSafeFunction::CallJS(napi_env env,
...
+
+  self->OnProgress(data, size);
+  delete[] data;
+}
+
+template<class T>
+inline void AsyncProgressWorker<T>::SendProgress_(const T* data, size_t count) {
+    T* new_data = new T[count];
+    {
+      T* it = new_data;
```
mhdawson: Why do we need a copy of the pointer here versus just using new_data in the copy?

Looks like this does not add to existing data, just swaps in the new data and deletes the old data. From the doc earlier I might have thought it needed to add it to the existing data.

legendecas: There are a few cases of how the data pointer were managed:

1. the data was created on the stack, that is how we tested it. In this case, the data has to be copied to be used on the main thread.
2. the data was created on the heap, there is no test on that but since there are no finalizers to free the data pointer until the main thread usage is done, it would be safer to just copy the data and have it managed by the Worker itself. 

With acknowledged above two conditions, I'd keep the API consistent with the NAN one.

napi.h
legendecas: Yes, the ExecutionProgress was created on the stack, and all its methods were safe to call on the worker thread.

napi-inl.h
mhdawson: Sorry I think my question was not clear. I understand why we need to copy the data.  The question is why we need to use the T* it pointer instead of just using new_data in the copy instead of `it`?

napi-inl.h
mhdawson: +1 to nullptr

legendecas: I think there is no more `NULL` usage after the last update. Resolving conversation. Reopen if there are any new cases :)

napi-inl.h
legendecas: Ah, yes, I think it is unnecessary too. Just copied from NAN implementation. Submitted a PR to NAN: https://github.com/nodejs/nan/pull/870.

napi-inl.h
```diff
@@ -4112,6 +4114,153 @@ inline void ThreadSafeFunction::CallJS(napi_env env,
...
+}
+
+template<class T>
+inline AsyncProgressWorker<T>::AsyncProgressWorker(Napi::Env env,
+                                const char* resource_name,
+                                const Object& resource)
+  : AsyncWorker(env, resource_name, resource),
+    _asyncdata(nullptr),
+    _asyncsize(0) {
+  // FIXME: Napi::ThreadSafeFunction nullptr callback support?
```
mhdawson: @legendecas everything else looks good, thanks for making the updates.  The last thing is this //FIXME.  If this is about being able to improve once the callback is optional I think a comment along the following lines would be better:

```bash
// once the changes to make the callback optional for threadsafe functions are
// no longer optional we can remove the dummy Function here. 
```

or whatever makes sense, as I'm not 100% sure without checking that this is related to to making callbacks optional, only that I remember that is being done for at least one case.

doc/async_progress_worker.md
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    void OnOK() {
+        HandleScope scope(Env());
+        Callback().Call({Env().Null(), String::New(Env(), echo)});
+    }
+
+    void OnProgress(const uint32_t* data, size_t /* count */) {
+        HandleScope scope(Env());
+        Callback().Call({Env().Null(), Env().Null(), Number::New(Env(), data)});
```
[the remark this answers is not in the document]
legendecas: Yes, it's a pointer to an integer. I'll update the doc.

For the later question, as documented above, `progress.Send` will copy the data pointed to right on the call to the `progress.Send`. So it is safe to do so without any synchronizations.

napi-inl.h
```diff
@@ -4112,6 +4114,154 @@ inline void ThreadSafeFunction::CallJS(napi_env env,
...
+template<class T>
+inline AsyncProgressWorker<T>::AsyncProgressWorker(Napi::Env env)
+  : AsyncProgressWorker(env, "generic") {
+}
+
+template<class T>
+inline AsyncProgressWorker<T>::AsyncProgressWorker(Napi::Env env,
+                                const char* resource_name)
+  : AsyncProgressWorker(env, resource_name, Object::New(env)) {
+}
```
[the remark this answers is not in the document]
legendecas: I think the rule would not apply to AsyncWorker and its variants for most usage of these classes is to implement a new variant and override necessary methods. It might be not convenient to provide a `New` factory functions as other Napi Values.

mhdawson:
Landed as 650562cab9fade291169c6574c26d554465101b7

PR #534, opened 2019-09-04
ci: enable travis for fast PR check

legendecas:
##### test: remove unnecessary NAPI_EXPERIMENTAL
-----
As ci.nodejs.org doesn't automatically run for PRs, travis is still the easiest way to determine if the PR is test passed on multiple node variants and platforms.

##### travis: remove chakracore on travis
1. node-chakracore v8 doesn’t expose napi version
2. node-chakracore v10 doesn’t throws on handlescope double escape
3. set npm_config_NAPI_VERSION to override NAPI_EXPERIMENTAL expansion
  on Node.js version lower than 11 for napi_date_* not back ported to v10

- [x] npm test passed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,20 +12,15 @@ env:
     # https://github.com/jasongin/nvs/blob/master/doc/CI.md
     - NVS_VERSION=1.4.2
   matrix:
-    - NODEJS_VERSION=node/4
     - NODEJS_VERSION=node/6
     - NODEJS_VERSION=node/8
-    - NODEJS_VERSION=node/9
     - NODEJS_VERSION=node/10
-    - NODEJS_VERSION=chakracore/8
-    - NODEJS_VERSION=chakracore/10
+    - NODEJS_VERSION=node/12
     - NODEJS_VERSION=nightly
-    - NODEJS_VERSION=chakracore-nightly
 matrix:
   fast_finish: true
   allow_failures:
     - env: NODEJS_VERSION=nightly
-    - env: NODEJS_VERSION=chakracore-nightly
 sudo: false
 cache:
   directories:
@@ -59,7 +54,13 @@ install:
 script:
   # Travis CI sets NVM_NODEJS_ORG_MIRROR, but it makes node-gyp fail to download headers for nightly builds.
   - unset NVM_NODEJS_ORG_MIRROR
+  - NODEJS_MAJOR_VERSION=$(node -p "process.versions.node.match(/\d+/)[0]")
 
-  - npm test $NPMOPT
+  - |
+    if [ ${NODEJS_MAJOR_VERSION} -gt 11 ]; then
+      npm test
+    else
+      npm test $NPMOPT --NAPI_VERSION=$(node -p "process.versions.napi")
+    fi
 after_success:
   - cpp-coveralls --gcov-options '\-lp' --build-root test/build --exclude test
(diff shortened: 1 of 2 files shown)


test/bigint.cc
```diff
@@ -1,4 +1,3 @@
-#define NAPI_EXPERIMENTAL
```
mhdawson: I think this will cause the tests not to run due to this later check #if (NAPI_VERSION > 2147483646)

legendecas: 🤔 maybe an option to control whether `NAPI_EXPERIMENTAL` should be defined or not might be needed.

mhdawson: If I remember correctly we had it enabled by default so that we'd run the tests in the most common use case (running on master) and then then CI sets the env variable so that the tests don't run when they run on earlier node version (ex Node.js v8)

legendecas: But if `NAPI_EXPERIMENTAL` is defined the binding.node would mark napi_bigint related symbol as required, which should definitely fail on Node.js v8.

legendecas: I noticed `npm_config_NAPI_VERSION` in `test/index.js` which would disable bigint tests. AFAIKT, it doesn't affect compilation right? Then requiring binding.node would fail on Node versions that don't have bigint NAPIs.

legendecas: https://travis-ci.com/legendecas/node-addon-api/builds/125931273

Used node major version to determine if experimental tests should be ran.

In Node.js v8 bigint tests were ignored.
In Node.js v10 bigint tests were ran.

legendecas:
@mhdawson hi, just updated the PR to use `npm_config_NAPI_VERSION` with node-gyp on travis which should be aligned with ci.nodejs.org. Please take a look :)

test/index.js
```diff
@@ -45,8 +47,10 @@ let testModules = [
   'version_management'
 ];
 
-if ((process.env.npm_config_NAPI_VERSION !== undefined) &&
-    (process.env.npm_config_NAPI_VERSION < 50000)) {
+const napiVersion = Number(process.env.npm_config_NAPI_VERSION || process.versions.napi);
```
mhdawson: I think this will still disable testing the experimental tests by default right?  I think we want to make sure that if you just do a checkout that they will run unless you specifically do something to prevent that.  Otherwise It will be too easy for us to miss running the experimental tests when we land PRs.  Previously if npm_config_NAPI_VERSION was not set then all tests would run, afterwards the experimental tests will never run unless npm_config_NAPI_VERSION is set to run then....

legendecas: yeah, I noticed that with your mention.

Also I found out that napi_date related API was landed on Node.js v12 as experimental yet NAPI v5 on master. And bigint related API was landed on Node.js v8, v10, v12 and master as experimental. It might be not enough to control experimental features merely by NAPI_VERSION if we'd like to address failure of old Node.js version running node-addon-api tests locally with simple `npm test`.

So maybe NAPI_EXPERIMENTAL features should be filtered with Node.js major version.

Nevertheless, I'd like to submit a independent PR to address this issue. For this PR, we could enable travis for fast PR test check?

mhdawson: Seems reasonable to me.  Will wait until Monday to see if we have any other comments, otherwise will land.

mhdawson:
Landed as 5d6aeae7b58010b09ad4d7a71b202398417fc7cb

PR #545, opened 2019-09-18
test: conditional test experimental features with NODE_MAJOR_VERSION

legendecas:
##### test: conditionally test experimental features with NODE_MAJOR_VERSION
NAPI_EXPERIMENTAL would be expand to NAPI_VERSION=2147483647 if
NAPI_VERSION is not defined. It would be not compatible with various
Node.js versions if we use NAPI_VERSION > 2147483646 mixed with
definition of NAPI_EXPERIMENTAL. This fix introduced NODE_MAJOR_VERSION
to allow more precise control over test sets that which set should be
enabled on a Node.js release.

Refs: https://github.com/nodejs/node-addon-api/pull/534#discussion_r321590943

------

##### build: remove unnecessary NAPI_VERSION pre-definition on travis
Since experimental api tests were conditioned by NODE_MAJOR_VERSION, we could simply run tests with `npm test` on every Node.js LTS version.

- [x] npm test passed
  - [x] Node.js v8.16.1
  - [x] Node.js v10.16.3
  - [x] Node.js v12.10.0

------

##### src: enabling BigInt API with NAPI_EXPERIMENTAL
Strictly aligning with Node.js header js_native_api.h in which
napi_bigint related apis were nested in NAPI_EXPERIMENTAL.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -376,8 +376,9 @@ inline bool Value::IsNumber() const {
   return Type() == napi_number;
 }
 
-// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
-// released in once it is no longer experimental
+// Currently experimental guard with the definition of NAPI_EXPERIMENTAL.
+// Once it is no longer experimental guard with the NAPI_VERSION in which it is
+// released instead.
 #ifdef NAPI_EXPERIMENTAL
 inline bool Value::IsBigInt() const {
   return Type() == napi_bigint;
@@ -613,8 +614,9 @@ inline double Number::DoubleValue() const {
   return result;
 }
 
-// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
-// released in once it is no longer experimental
+// Currently experimental guard with the definition of NAPI_EXPERIMENTAL.
+// Once it is no longer experimental guard with the NAPI_VERSION in which it is
+// released instead.
 #ifdef NAPI_EXPERIMENTAL
 ////////////////////////////////////////////////////////////////////////////////
 // BigInt Class
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -376,9 +376,9 @@ inline bool Value::IsNumber() const {
   return Type() == napi_number;
 }
 
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
 // released in once it is no longer experimental
-#if (NAPI_VERSION > 2147483646)
+#ifdef NAPI_EXPERIMENTAL
 inline bool Value::IsBigInt() const {
   return Type() == napi_bigint;
 }
@@ -613,9 +613,9 @@ inline double Number::DoubleValue() const {
   return result;
 }
 
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
 // released in once it is no longer experimental
-#if (NAPI_VERSION > 2147483646)
+#ifdef NAPI_EXPERIMENTAL
 ////////////////////////////////////////////////////////////////////////////////
 // BigInt Class
 ////////////////////////////////////////////////////////////////////////////////
--- a/napi.h
+++ b/napi.h
@@ -111,8 +111,9 @@ namespace Napi {
   class Value;
   class Boolean;
   class Number;
-// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
-// released in once it is no longer experimental
+// Currently experimental guard with the definition of NAPI_EXPERIMENTAL.
+// Once it is no longer experimental guard with the NAPI_VERSION in which it is
+// released instead.
 #ifdef NAPI_EXPERIMENTAL
   class BigInt;
 #endif  // NAPI_EXPERIMENTAL
@@ -139,8 +140,9 @@ namespace Napi {
   typedef TypedArrayOf<uint32_t> Uint32Array; ///< Typed-array of unsigned 32-bit integers
   typedef TypedArrayOf<float> Float32Array;   ///< Typed-array of 32-bit floating-point values
   typedef TypedArrayOf<double> Float64Array;  ///< Typed-array of 64-bit floating-point values
-// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
-// released in once it is no longer experimental
+// Currently experimental guard with the definition of NAPI_EXPERIMENTAL.
+// Once it is no longer experimental guard with the NAPI_VERSION in which it is
+// released instead.
 #ifdef NAPI_EXPERIMENTAL
   typedef TypedArrayOf<int64_t> BigInt64Array;   ///< Typed array of signed 64-bit integers
   typedef TypedArrayOf<uint64_t> BigUint64Array; ///< Typed array of unsigned 64-bit integers
@@ -244,8 +246,9 @@ namespace Napi {
     bool IsNull() const;        ///< Tests if a value is a null JavaScript value.
     bool IsBoolean() const;     ///< Tests if a value is a JavaScript boolean.
     bool IsNumber() const;      ///< Tests if a value is a JavaScript number.
-// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
-// released in once it is no longer experimental
+// Currently experimental guard with the definition of NAPI_EXPERIMENTAL.
+// Once it is no longer experimental guard with the NAPI_VERSION in which it is
+// released instead.
 #ifdef NAPI_EXPERIMENTAL
     bool IsBigInt() const;      ///< Tests if a value is a JavaScript bigint.
 #endif  // NAPI_EXPERIMENTAL
@@ -321,8 +324,9 @@ namespace Napi {
     double DoubleValue() const;   ///< Converts a Number value to a 64-bit floating-point value.
   };
 
-// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
-// released in once it is no longer experimental
+// Currently experimental guard with the definition of NAPI_EXPERIMENTAL.
+// Once it is no longer experimental guard with the NAPI_VERSION in which it is
+// released instead.
 #ifdef NAPI_EXPERIMENTAL
   /// A JavaScript bigint value.
   class BigInt : public Value {
@@ -851,8 +855,9 @@ namespace Napi {
         : std::is_same<T, uint32_t>::value ? napi_uint32_array
         : std::is_same<T, float>::value ? napi_float32_array
         : std::is_same<T, double>::value ? napi_float64_array
-// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
-// released in once it is no longer experimental
+// Currently experimental guard with the definition of NAPI_EXPERIMENTAL.
+// Once it is no longer experimental guard with the NAPI_VERSION in which it is
+// released instead.
 #ifdef NAPI_EXPERIMENTAL
         : std::is_same<T, int64_t>::value ? napi_bigint64_array
         : std::is_same<T, uint64_t>::value ? napi_biguint64_array
--- a/napi.h
+++ b/napi.h
@@ -111,9 +111,9 @@ namespace Napi {
   class Value;
   class Boolean;
   class Number;
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
 // released in once it is no longer experimental
-#if (NAPI_VERSION > 2147483646)
+#ifdef NAPI_EXPERIMENTAL
   class BigInt;
 #endif  // NAPI_EXPERIMENTAL
 #if (NAPI_VERSION > 4)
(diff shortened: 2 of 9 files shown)


mhdawson:
I'll have to find time to look more carefully, but I don't think we should have checks based on NODE_MAJOR_VERSION.  It should be that either you need to enable experimental or guard against a specific N-API version.  I could be wrong but I think adding checks against the NODE_MAJOR_VERSION will go counter to us wanting to provide support across versions.

legendecas:
IMO, NODE_MAJOR_VERSION should only be used on the condition of experimental new apis like napi_bigint and previously napi_date which has different backport status so we could distinguish if the related test could be ran on that Node.js version hardly with NAPI_EXPERIMENTAL definition.

Out of experimental api should be guarded with NAPI_VERSION normally.

legendecas:
For the issue of test not been run, check out here https://travis-ci.com/nodejs/node-addon-api/jobs/248367006#L296, the tests of bigint and date have been run on Node.js v10 and Node.js v12.

gabrielschulhof:
@mhdawson I believe that using `NODE_MAJOR_VERSION` should be OK in the tests. Nevertheless, it must not make its way into `napi.h` or `napi-inl.h`.

legendecas:
Just updated the descriptions of the changelist to better indicates the NODE_MAJOR_VERSION was only used in tests.

@mhdawson please take another look :)

napi-inl.h
```diff
@@ -613,9 +613,9 @@ inline double Number::DoubleValue() const {
   return result;
 }
 
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
```
mhdawson:
```suggestion
// currently experimental guard with version of NAPI_VERSION that it is
```

napi-inl.h
```diff
@@ -376,9 +376,9 @@ inline bool Value::IsNumber() const {
   return Type() == napi_number;
 }
 
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
```
mhdawson:
```suggestion
// currently experimental guard with version of NAPI_VERSION that it is
```

napi.h
```diff
@@ -111,9 +111,9 @@ namespace Napi {
   class Value;
   class Boolean;
   class Number;
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
```
mhdawson:
```suggestion
// currently experimental guard with version of NAPI_VERSION that it is
```

napi.h
```diff
@@ -139,9 +139,9 @@ namespace Napi {
   typedef TypedArrayOf<uint32_t> Uint32Array; ///< Typed-array of unsigned 32-bit integers
   typedef TypedArrayOf<float> Float32Array;   ///< Typed-array of 32-bit floating-point values
   typedef TypedArrayOf<double> Float64Array;  ///< Typed-array of 64-bit floating-point values
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
```
mhdawson:
```suggestion
// currently experimental guard with version of NAPI_VERSION that it is
```

napi.h
```diff
@@ -244,9 +244,9 @@ namespace Napi {
     bool IsNull() const;        ///< Tests if a value is a null JavaScript value.
     bool IsBoolean() const;     ///< Tests if a value is a JavaScript boolean.
     bool IsNumber() const;      ///< Tests if a value is a JavaScript number.
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
```
mhdawson:
```suggestion
// currently experimental guard with version of NAPI_VERSION that it is
```

napi.h
```diff
@@ -321,9 +321,9 @@ namespace Napi {
     double DoubleValue() const;   ///< Converts a Number value to a 64-bit floating-point value.
   };
 
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
```
mhdawson:
```suggestion
// currently experimental guard with version of NAPI_VERSION that it is
```

napi.h
```diff
@@ -851,9 +851,9 @@ namespace Napi {
         : std::is_same<T, uint32_t>::value ? napi_uint32_array
         : std::is_same<T, float>::value ? napi_float32_array
         : std::is_same<T, double>::value ? napi_float64_array
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
```
mhdawson:
```suggestion
// currently experimental guard with version of NAPI_VERSION that it is
```

test/bigint.cc
```diff
@@ -1,11 +1,12 @@
+// currently experimental guard with version of NODE_MAJOR_VERSION that it is
```
mhdawson:
```suggestion
// currently experimental guard with version of NAPI_VERSION that it is
```

test/binding.cc
```diff
@@ -11,9 +10,9 @@ Object InitBasicTypesArray(Env env);
 Object InitBasicTypesBoolean(Env env);
 Object InitBasicTypesNumber(Env env);
 Object InitBasicTypesValue(Env env);
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with version of NODE_MAJOR_VERSION that it is
```
mhdawson:
```suggestion
// currently experimental, but tests can't tell which version of Node.js supports
// which experimental features.  Temporarily guard with NODE_MAJOR_VERISION.
// Guard with version of NAPI_VERSION that it is
```

test/binding.cc
```diff
@@ -56,9 +55,9 @@ Object Init(Env env, Object exports) {
   exports.Set("basic_types_boolean", InitBasicTypesBoolean(env));
   exports.Set("basic_types_number", InitBasicTypesNumber(env));
   exports.Set("basic_types_value", InitBasicTypesValue(env));
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with version of NODE_MAJOR_VERSION that it is
```
mhdawson:
```suggestion
// currently experimental, but tests can't tell which version of Node.js supports
// which experimental features.  Temporarily guard with NODE_MAJOR_VERISION.
// Guard with version of NAPI_VERSION that it is
```

test/typedarray.cc
```diff
@@ -1,4 +1,8 @@
+// currently experimental guard with version of NODE_MAJOR_VERSION that it is
```
mhdawson:
```suggestion
// currently experimental, but tests can't tell which version of Node.js supports
// which experimental features.  Temporarily guard with NODE_MAJOR_VERISION.
// Guard with version of NAPI_VERSION that it is
```

test/typedarray.cc
```diff
@@ -65,9 +69,9 @@ Value CreateTypedArray(const CallbackInfo& info) {
       NAPI_TYPEDARRAY_NEW(Float64Array, info.Env(), length, napi_float64_array) :
       NAPI_TYPEDARRAY_NEW_BUFFER(Float64Array, info.Env(), length, buffer, bufferOffset,
                                  napi_float64_array);
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with version of NODE_MAJOR_VERSION that it is
```
mhdawson:
```suggestion
// currently experimental, but tests can't tell which version of Node.js supports
// which experimental features.  Temporarily guard with NODE_MAJOR_VERISION.
// Guard with version of NAPI_VERSION that it is
```

test/typedarray.cc
```diff
@@ -101,9 +105,9 @@ Value GetTypedArrayType(const CallbackInfo& info) {
     case napi_uint32_array: return String::New(info.Env(), "uint32");
     case napi_float32_array: return String::New(info.Env(), "float32");
     case napi_float64_array: return String::New(info.Env(), "float64");
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with version of NODE_MAJOR_VERSION that it is
```
mhdawson:
```suggestion
// currently experimental, but tests can't tell which version of Node.js supports
// which experimental features.  Temporarily guard with NODE_MAJOR_VERISION.
// Guard with version of NAPI_VERSION that it is
```

test/typedarray.cc
```diff
@@ -143,9 +147,9 @@ Value GetTypedArrayElement(const CallbackInfo& info) {
       return Number::New(info.Env(), array.As<Float32Array>()[index]);
     case napi_float64_array:
       return Number::New(info.Env(), array.As<Float64Array>()[index]);
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with version of NODE_MAJOR_VERSION that it is
```
mhdawson:
```suggestion
// currently experimental, but tests can't tell which version of Node.js supports
// which experimental features.  Temporarily guard with NODE_MAJOR_VERISION.
// Guard with version of NAPI_VERSION that it is
```

test/typedarray.cc
```diff
@@ -189,9 +193,9 @@ void SetTypedArrayElement(const CallbackInfo& info) {
     case napi_float64_array:
       array.As<Float64Array>()[index] = value.DoubleValue();
       break;
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with version of NODE_MAJOR_VERSION that it is
```
mhdawson:
```suggestion
// currently experimental, but tests can't tell which version of Node.js supports
// which experimental features.  Temporarily guard with NODE_MAJOR_VERISION.
// Guard with version of NAPI_VERSION that it is
```

test/index.js
```diff
@@ -48,10 +48,12 @@ let testModules = [
...
-if ((process.env.npm_config_NAPI_VERSION !== undefined) &&
-    (process.env.npm_config_NAPI_VERSION < 50000)) {
-  // currently experimental only test if NAPI_VERSION
-  // is set to experimental. We can't use C max int
+const napiVersion = Number(process.versions.napi)
+const nodeMajorVersion = Number(process.versions.node.match(/\d+/)[0])
+
+if (nodeMajorVersion < 10) {
+  // currently experimental only test if node major version
+  // is set to experimental. We can't use napi_experimental here
```
mhdawson:
```suggestion
is >= were it exists. We can't use NAPI_EXPERIMENTAL here
as the tests can't tell which version of Node.js support which
experimental versions.
```

test/index.js
```diff
@@ -48,10 +48,12 @@ let testModules = [
...
   // as that is not supported as a number on earlier
```
mhdawson:
```suggestion
```

test/index.js
```diff
@@ -48,10 +48,12 @@ let testModules = [
...
   // Node.js versions. Once bigint is in a release
```
mhdawson:
```suggestion
Once bigint is in a release
 ```

mhdawson:
@legendecas having looked at this carefully I agree that we need to guard the tests with the NODE_MAJOR_VERSION but the comment need to say that when the feature is no longer experimental we'll guard with with the NAPI_VERSION instead.  I've made suggested comments.

If you think those make sense please accept them and let me know and then I'll run some CI tests to validate and we can get this landed.

napi-inl.h
legendecas: With commit `src: enabling BigInt API with NAPI_EXPERIMENTAL`, BigInt would only be enabled with the definition of `NAPI_EXPERIMENTAL` rather than a very big `NAPI_VERSION`, which is what has been done in `js_native_api.h`.

test/bigint.cc
legendecas: I think experimental feature tests do tell which Node.js versions it could be run on. And we should continuously update the test table on each backport.

And experimental features were always temporary status, IMO. It is not necessary to point out that an experimental guard flag is temporary, IMO.

legendecas:
There are many comments with similar content. Posting what I've replied as links:
- https://github.com/nodejs/node-addon-api/pull/545#discussion_r341236641
- https://github.com/nodejs/node-addon-api/pull/545#discussion_r341238208

Also, I've updated the comment in `test/index.js` to indicate when the feature is no longer experimental we'll guard with the `NAPI_VERSION` instead.

napi-inl.h
mhdawson: What the comment is trying to say is:

* Currently it is experimental
* When it becomes non-experimental is should be guarded with the NAPI_VERSION in which it became non-experimental (for example it might be version 6).

Maybe it comment should be

// currently experimental, for now guard with definition of NAPI_EXPERIMENTAL.  Once it is no
// longer experimental guard with the NAPI_VERSION in which it is released.

test/bigint.cc
mhdawson: I think it is important to point out that we should NOT use NODE_MAJOR_VERSION once it is not experimental.

I'd be ok with:

// currently experimental temporarily guard with NODE_MAJOR_VERISION.
// Once its no longer experimental guard with version of NAPI_VERSION that it is release in.

legendecas:
@mhdawson I've updated all the comments that mentioned experimental guard with your concern of note on future works. PTAL :)

mhdawson:
@legendecas thanks. Looking now.

mhdawson:
Landed as 295e560f5554c87d0a9dde6b73942ebd85fddb9d

legendecas:
@mhdawson @gabrielschulhof I think the test script of https://ci.nodejs.org/job/node-test-node-addon-api-new could be simplified as same as what's done to .travis.yml in this PR. How do we make changes to it?

Issue #524, opened 2019-08-15
Issues with the new ThreadSafeFunction api


extmchristensen:
@romandev I am enthusiastically trying to use the new ThreadSafeFunction Api as a replacement for the non-standard napi-thread-safe-callback which I was using previously. I have a few problems though in my (typical?) use case where I have async native C++ functions that return quickly to JS but continue to do the actual work in an async C++ worker or thread, calling back into JS with progress information and/or result(s) using ThreadSafeFunction.

**NB: Updated to show that issue 1 & 2 is now solved**

**1)** SOLVED: Most importantly, the ThreadSafeCallback class is lacking a copy-constructor so I can not safely and easily capture it into c++ threads, lambdas and AsyncWorker's with pass by value.  The provided example at the bottom of "https://github.com/nodejs/node-addon-api/blob/master/doc/threadsafe_function.md" solves this by having ThreadSafeCallback as a global mutable variable but this is dangerous (might break with repeated calls to the native function quickly after each other) and I also personally think global mutable variables is a bad thing. Instead the native method should initialize the ThreadSafeCallback object which is than copied into the thread/lambda/worker by value (before original object is destroyed as it goes out of scope when function returns etc).

**2)** FIXED BY ABOVE: In the absence of a copy-constructor, it is a problem that the ThreadSafeCallback class has only factory methods (for parameters) and lacks a constructor with the various parameters so new/delete could be used to control the life-cycle and hereby enable capturing of a pointer to ThreadSafeFunction in c++ threads, workers and lambdas (As a temporary fix, I have created by own simple wrapper class that holds a ThreadSafeFunction inside and has a constructor with all the arguments thus enabling the wrapper with the ThreadSafeFunction to be created on the heap).

**3)** It is not clear what initialThreadCount should be set to and how and when Acquire and Release should be called. The example sets initialThreadCount to 1 though used by two threads and does not call  Acquire ? From my own experiments, it seems important to call Release exactly once to make sure the node process terminates - it does not seem to matter what Thread I call Release from in my example as long as I don't call it twice (in which case an Assert error with a nice stack trace will be shown).

**4** There is no way for the C++ code to wait for the queue to be empty (all calls in queue to be processed). This would be useful to delay cleanup until a NonBlockingCall has been fully executed.

mhdawson:
@extmchristensen, @gabrielschulhof who it probably the best person to comment is away for another week so it might be after that when we engage in a discussion on this.

almogh52:
[...]

Issue #544, opened 2019-09-17
Add example/api to use ThreadSafeFunction with optional callback param


legendecas:
Since https://github.com/nodejs/node/pull/27791 has been landed, there might be some use cases on node-addon-api to use `Napi::ThreadSafeFunction` with optional callback parameter.

Approaches might be adding an example of how to use existing `Napi::ThreadSafeFunction::New` like:

```c++
Function callback;
auto tsfn = ThreadSafeFunction::New(env, callback, resource_name, 1, 1);
```

(is there a more intuitive one?)

Or add a new API which make the callback omitted:
```c++
auto tsfn = ThreadSafeFunction::New(env, resource_name, 1, 1);
```

KevinEady:
I think it would make more sense to add a new signature without the `Function` argument. I feel like that's more intuitive. We are currently working on a fix to add a copy constructor, so odds are we could add this as well.

KevinEady:
Sooo I think implementing a new constructor will blow up the number of constructors we have. We currently have _thirteen_ `::New` overloads, and if making the `Function`  optional by removing it, we would add thirteen more. I dunno, we will discuss in next napi meeting

KevinEady:
I would like to use this issue to highlight all issues with TSFN constructors. I point out the ambiguity between the TSFN::New with [`Finalizer + FinalizerDataType`](https://github.com/nodejs/node-addon-api/blob/1b42bfc62217edb20b093aadfd46a712e1d4a026/napi-inl.h#L3837) and [`ContextType + Finalizer`](https://github.com/nodejs/node-addon-api/blob/1b42bfc62217edb20b093aadfd46a712e1d4a026/napi-inl.h#L3938):

```cpp
struct TestData {
  // Native Promise returned to JavaScript
  Promise::Deferred deferred;

  // List of threads created for test. This list only ever accessed via main
  // thread.
  vector<thread> threads = {};

  ThreadSafeFunction tsfn = ThreadSafeFunction();
};

void FinalizerCallback(Napi::Env env, TestData* finalizeData){ 
  for (size_t i = 0; i < finalizeData->threads.size(); ++i) {   
    finalizeData->threads[i].join();                            
  }                                                             
  finalizeData->deferred.Resolve(Boolean::New(env,true));       
  delete finalizeData;
}

static Value TestAcquire(const CallbackInfo& info) { 
 // ....
  TestData *testData = new TestData({
    Promise::Deferred::New(env)
  });

[...]

PR #546, opened 2019-09-18
tsfn: Implement copy constructor

KevinEady:
Removes the `unique_ptr` from `ThreadSafeFunction`, thereby allowing
copies.

Ref: https://github.com/nodejs/node-addon-api/issues/524


KevinEady:
Hi @legendecas ,

So... IMO I don't think it should automatically acquire/release. I believe it adds an assumption that each copy must perform an `Acquire`, which I think is not true. In fact, I think it would actually cause problems in certain implementations. 

Also, if `Acquire()` returns `napi_closing`, I would not consider this a fatal error so we shouldn't throw in the copy ctor. How would we handle this? Storing the status on the newly constructed object for the developer to check after copy, eg. in the thread entry point? I dunno, seems clunky...

gabrielschulhof:
@legendecas IINM passing to another thread may not be the only reason for copying a `Napi::ThreadSafeFunction`. For example, if you used it as a key in a hash table ...

gabrielschulhof:
Also IINM it should be easy to subclass `Napi::ThreadSafeFunction` and add the automatic `Acquire()`/`Release()` for those that need it.

KevinEady:
So I have no idea about the ambiguous function calls for `New` as mentioned in https://github.com/nodejs/node-addon-api/issues/544#issuecomment-534175977 but I don't think it should block this. If so, I believe this is ready.

mhdawson:
I agree that the problem related to ambiguous function calls is independent and could be addressed in a follow on PR.

test/threadsafe_function/threadsafe_function_sum.cc
```diff
@@ -0,0 +1,155 @@
...
+    callback.Call( { Number::New(env, static_cast<double>(threadId))});
+  });
+  tsfn.Release();
+}
+
+static Value CreateThread(const CallbackInfo& info) {
+  TestData* testData = static_cast<TestData*>(info.Data());
+  ThreadSafeFunction tsfn = testData->tsfn;
+  int threadId = testData->threads.size();
+  testData->threads.push_back( thread(entryAcquire, tsfn, threadId) );
```
mhdawson: I wonder if it would be useful to highlight with a comment that this is where the copy constructor is being tested.  Since the main change is to add the copy constructor it would be good to make it obvious in the test that it is being used.

KevinEady: Added comment

KevinEady:
Hey @mhdawson, can we get this landed so I can continue work on Ref/Unref as well as the incorrect `[Non]BlockingCall` with `ThreadSafeFunction(napi_threadsafe_function)` constructor

mhdawson:
I'm in the process of landing. It did not seem to compile with master and the older 4.8.5 compiler.  That may just be because master need gcc 6, but I want to run a CI on 8.x before landing. 

Unfortunately, the CI job is having some issues getting the Node binaries so I'll have to figure that out first.

mhdawson:
This seems to fail to build on 8.x with the 4.8.5 compiler level. You can see the compile failures her: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/nodes=ubuntu1604-64/822/console

mhdawson:
@KevinEady can you take a look:

```
../threadsafe_function/threadsafe_function_sum.cc: In function ‘Napi::Value {anonymous}::TestWithTSFN(const Napi::CallbackInfo&)’:
../threadsafe_function/threadsafe_function_sum.cc:51:4: error: no matching function for call to ‘{anonymous}::TestData::TestData(<brace-enclosed initializer list>)’
   });
    ^
../threadsafe_function/threadsafe_function_sum.cc:12:8: note: candidate: {anonymous}::TestData::TestData(const {anonymous}::TestData&)
 struct TestData {
        ^
../threadsafe_function/threadsafe_function_sum.cc:12:8: note:   no known conversion for argument 1 from ‘<brace-enclosed initializer list>’ to ‘const {anonymous}::TestData&’

```

KevinEady:
Hey team,

So I was actually testing this on Windows, and it was failing because defining a move operator implicitly deletes the assignment operator... And, if you think about it, the `ThreadSafeFunction` wrapper only has for its member data this `napi_threadsafe_function` which behaves as a pointer: we can use the default compiler-generated functions anyways, right? 

```
test\threadsafe_function\threadsafe_function_sum.cc(90): error C2280: 'Napi::ThreadSafeFunction &Napi::ThreadSafeFunction::operator =(const Napi::ThreadSafeFunction &)': attempting to reference a deleted function [test\build\binding_noexcept.vcxproj]
  napi.h(2088): note: compiler has generated 'Napi::ThreadSafeFunction::operator =' here
  napi.h(2088): note: 'Napi::ThreadSafeFunction &Napi::ThreadSafeFunction::operator =(const Napi::ThreadSafeFunction &)': function was implicitly deleted because 'Napi::ThreadSafeFunction' has a user-defined move constructor (compiling source file ..\threadsafe_function\threadsafe_function_sum.cc)
```

I did not think it was worthwhile to explicitly create these constructors, nor state them as default in napi.h , so I just removed them. Since this is structural change, I've dismissed your review @mhdawson

mhdawson:
CI job to validate builds on Node 8.x: https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/854/

mhdawson:
@KevinEady seems to be failures on osx and aix: : https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/854/

KevinEady:
Hi @mhdawson ,

Fixed some issues in the tests that I know were problematic, but this one is really stumping me..

https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/nodes=osx1011/855/console : 

```
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/future:1170:1: error: 'future<Napi::ThreadSafeFunction>' is unavailable: introduced in macOS 10.8
future<_Rp>::get()
```

I don't know, just looking at the node name os `osx1011` ... but is this build targeting osx 10.8 ...?

KevinEady:
Hi @mhdawson / @gabrielschulhof ,

So I re-wrote my tests to use `std::condition_variable` and `std::mutex` instead of promises/futures. Let me know if this is suitable now.

KevinEady:
Hi @legendecas ,

> I noticed there are still usages on promises/futures in `test/threadsafe_function/threadsafe_function_sum.cc`. Are they intended?

So I noticed I did not remove the <future> include, which I now replaced with the correct ones for use with condition_variable/mutex.

However, the promise / deferred you see are `Napi::Promise`s, not std library. They are used in the test to ensure proper synchronization. Does this address your concern?

mhdawson:
CI run to see if it works on 8.x https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-new/906/

mhdawson:
@legendecas let us know if the latest changes address your concerns and then approve if appropriate. Will wait to hear from you before landing.

napi-inl.h
```diff
@@ -4083,34 +4070,34 @@ inline napi_status ThreadSafeFunction::NonBlockingCall(
...
 inline napi_status ThreadSafeFunction::Abort() {
-  return napi_release_threadsafe_function(*_tsfn, napi_tsfn_abort);
+  return napi_release_threadsafe_function(_tsfn, napi_tsfn_abort);
 }
 
 inline ThreadSafeFunction::ConvertibleContext
 ThreadSafeFunction::GetContext() const {
   void* context;
-  napi_get_threadsafe_function_context(*_tsfn, &context);
+  napi_get_threadsafe_function_context(_tsfn, &context);
```
legendecas: Abort if failed here?

legendecas: Since the change doesn't change the behavior here, we could address this issue in another PR. Opened issue here https://github.com/nodejs/node-addon-api/issues/581.

mhdawson: @legendecas I agree, can you open an issue to make sure we fix it later?

KevinEady: @mhdawson the issue is above, and PR for fix (which is also conflicting with this PR) is #583

mhdawson: @KevinEady - oops should have read more carefullly. Thanks.

mhdawson:
Landed as 2e71842f633b4138baddc2982079f4590e991fa6

PR #547, opened 2019-09-22
Conditional enabling BigInt API with NAPI_EXPERIMENTAL

legendecas:
Strictly aligning with Node.js header js_native_api.h in which
napi_bigint related apis were nested in NAPI_EXPERIMENTAL.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -371,9 +371,9 @@ inline bool Value::IsNumber() const {
   return Type() == napi_number;
 }
 
-// currently experimental guard with version of NAPI_VERSION that it is
+// currently experimental guard with definition of NAPI_EXPERIMENTAL that it is
 // released in once it is no longer experimental
-#if (NAPI_VERSION > 2147483646)
+#ifdef NAPI_EXPERIMENTAL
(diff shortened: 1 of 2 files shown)


mhdawson:
This may change the ways that the testing works.  

This is in js_native_api.h

#define NAPI_VERSION_EXPERIMENTAL 2147483647
#ifndef NAPI_VERSION
#ifdef NAPI_EXPERIMENTAL
#define NAPI_VERSION NAPI_VERSION_EXPERIMENTAL
#else
// The baseline version for N-API.
// The NAPI_VERSION controls which version will be used by default when
// compilling a native addon. If the addon developer specifically wants to use
// functions available in a new version of N-API that is not yet ported in all
// LTS versions, they can set NAPI_VERSION knowing that they have specifically
// depended on that version.
#define NAPI_VERSION 5
#endif
#endif

I believe the CI jobs for the node-addon-api set NAPI_VERSION.

So using the define versus comparing againts NAPI_VERSION might result in less testing or failed tests.  We'd need to run against all of the versions and double check we get the same tests run.

legendecas:
Closed in favor of #545.

PR #552, opened 2019-10-01
doc: fix return type of ArrayBuffer::Data

tniessen:
`ArrayBuffer` itself is not typed and `Data()` returns a `void*`.


mhdawson:
Landed as: b513d1aa7a06fc0cae6d1483831536ce0524d36a

PR #565, opened 2019-10-15
scope: disallow copying scopes to prevent double close of scopes

legendecas:
Early detect and prevent unexpected crashing on closing an already closed handle scope.

- [x] `npm test` passed

### DIFF
--- a/napi.h
+++ b/napi.h
@@ -1729,6 +1729,10 @@ namespace Napi {
     explicit HandleScope(Napi::Env env);
     ~HandleScope();
 
+    // Disallow copying to prevent double close of napi_handle_scope
+    HandleScope(HandleScope const &) = delete;
+    void operator=(HandleScope const &) = delete;
+
     operator napi_handle_scope() const;
 
     Napi::Env Env() const;
@@ -1743,6 +1747,10 @@ namespace Napi {
     EscapableHandleScope(napi_env env, napi_escapable_handle_scope scope);
     explicit EscapableHandleScope(Napi::Env env);
     ~EscapableHandleScope();
+
+    // Disallow copying to prevent double close of napi_escapable_handle_scope
+    EscapableHandleScope(EscapableHandleScope const &) = delete;
+    void operator=(EscapableHandleScope const &) = delete;
 
     operator napi_escapable_handle_scope() const;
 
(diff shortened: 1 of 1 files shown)


legendecas:
> I think that this should prevent runtime error in case of double close. @legendecas Am I right?

Yes, a `napi_handlescope_mismatch` would be returned if double closing a handle scope. 

Though I've found that return statuses of napi_close_*_scope have been ignored 🤔 https://github.com/nodejs/node-addon-api/blob/master/napi-inl.h#L3475. Are these behavior expected?

gabrielschulhof:
@legendecas I believe the reason for that is that, if C++ exceptions are turned on, one would be throwing a C++ exception from a destructor which by my understanding is a big no-no.

legendecas:
Since no exceptions were expected to be thrown in the destructor, I'd think to disallow copying of scopes should be an intuitive approach. What do you think? @gabrielschulhof

mhdawson:
This would be a breaking change.  While we've not said that will never happen for node-addon-api, we have chosen to avoid those in all cases before.

I think we'd want to have a really good reason to introduce a breaking change and I'm not sure it's going to be worth it in this case.

legendecas:
>  @mhdawson: This would be a breaking change.

Technically it would depend on platforms and *possibly crash* on double closing a handle scope -- since in node core `napi_handle_scope` is just a pointer to be deleted on `napi_close_handle_scope`.

So I'd wonder there would hardly be possible any use case in the wild. But yes, it is a breaking change.

legendecas:
Append: node core has a pre-condition of deleting the `HandleScopeWrapper*`, so a `napi_handle_scope_mismatch` would be returned but not crashing immediately, which is what #566 was expected to address.

legendecas:
PPS: still, I could produce the crash with the following snippet:

```c++

Value doubleCloseScopeNapi(const CallbackInfo& info) {
  Env env = info.Env();
  HandleScope scope(env);
  HandleScope scope2(env);
  HandleScope scope3 = scope2;
  return env.Undefined();
}
```

mhdawson:
Discussed in meeting today:

1) team is comfortable that it is extremely unlikely this will break any existing code, as the code would already be crashing.
2) V8 and NaN don't have the copy constructors as they don't make sense.

Consensus is we should we remove them.

mhdawson:
We did discuss and this should land as a SemVer major even though it should not affect any existing code.

mhdawson:
Landed as 34c11cf0a4f238654b011e3fd60e583a222d4498

PR #566, opened 2019-10-15
scopes: make failure of closing scopes fatal

legendecas:
But not ignoring these failures.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3472,7 +3472,10 @@ inline HandleScope::HandleScope(Napi::Env env) : _env(env) {
 }
 
 inline HandleScope::~HandleScope() {
-  napi_close_handle_scope(_env, _scope);
+  napi_status status = napi_close_handle_scope(_env, _scope);
+  NAPI_FATAL_IF_FAILED(status,
+                       "HandleScope::~HandleScope",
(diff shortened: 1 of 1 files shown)


legendecas:
I don't think it would affect the existing working code. Yet it would be helpful to debug possibly misusage of `Napi::HandleScope`s like:

```c++
Value doubleCloseScope(const CallbackInfo& info) {
  Env env = info.Env();
  HandleScope scope(env);
  HandleScope scope2 = scope; // this one would be closed with status `napi_handle_scope_mismatch` silently
  return env.Undefined();
}
```

mhdawson:
@gabrielschulhof if you agree it's unlikely to affect working code I'm ok with it.

mhdawson:
Discussion in N-API team meeting today was that this is directly related to https://github.com/nodejs/node-addon-api/pull/565.  We don't believe that the close should fail except for the case where it had been copied which will no longer be possible.

So we don't believe this will not break any existing code and its ok to land as a semver Patch fixing an bug.

mhdawson:
We did discuss and this should land as a SemVer major even though it should not affect any existing code.

mhdawson:
Landed as ce139a05e867e5b46cd8971324263823560a89f9

Issue #581, opened 2019-10-31
Error was ignored if `ThreadSafeFunction::GetContext()` failed


legendecas:
Currently `napi_status` was ignored in `ThreadSafeFunction::GetContext()`. No `napi_status` shall be ignored as it would unstablize the behavior we'd expect.

Since the function signature doesn't support error returning, an alternative might be aborting on non-`napi_ok` statuses.

_Originally posted by @legendecas in https://github.com/nodejs/node-addon-api/pull/546_

KevinEady:
Hi team,

I can start work on this if nobody has? (direct ping to @legendecas since he reported this)

legendecas:
@KevinEady You have it. :P

legendecas:
I think aborting would be just fine. Since we don't expect any error on this method. Neither do we have any recover solution on the error situation. Try `NAPI_FATAL_IF_FAILED`.

gabrielschulhof:
Looking at the implementation, `napi_get_threadsafe_function_context()` will only fail if one of the parameters passed in is `NULL`. So, since both parameters are under our control, it should never fail – unless, in the future, we introduce some new failure mode, in which case the abort should be perfect.

PR #580, opened 2019-10-29
tsfn: support direct calls to underlying napi_tsfn

KevinEady:
Fixes: https://github.com/nodejs/node-addon-api/issues/556

PR #583, opened 2019-10-31
tsfn: add error checking on GetContext

KevinEady:
Fixes: https://github.com/nodejs/node-addon-api/issues/581

### DIFF
--- a/test/threadsafe_function/threadsafe_function_ctx.cc
+++ b/test/threadsafe_function/threadsafe_function_ctx.cc
@@ -1,0 +1,64 @@
+#include "napi.h"
+
+#if (NAPI_VERSION > 3)
+
+using namespace Napi;
+
+namespace {
+
+class TSFNWrap : public ObjectWrap<TSFNWrap> {
+public:
+  static Object Init(Napi::Env env, Object exports);
+  TSFNWrap(const CallbackInfo &info);
+
+  Napi::Value GetContext(const CallbackInfo & /*info*/) {
+    Reference<Napi::Value> *ctx = _tsfn.GetContext();
+    return ctx->Value();
+  };
+
+  Napi::Value Release(const CallbackInfo &info) {
+    Napi::Env env = info.Env();
+    _deferred = std::unique_ptr<Promise::Deferred>(new Promise::Deferred(env));
+    _tsfn.Release();
+    return _deferred->Promise();
+  };
+
+private:
+  ThreadSafeFunction _tsfn;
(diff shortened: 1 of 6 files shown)


KevinEady:
https://github.com/nodejs/node-addon-api/blob/e142a1ee57e5ebf8030fbd174ae2d481b727494b/napi-inl.h#L4113-L4114

Will need to review if FATAL error is correct. If not, how to properly handle this error

test/threadsafe_function/threadsafe_function_ctx.cc
```diff
@@ -0,0 +1,64 @@
...
+
+  Reference<Napi::Value> *_ctx = new Reference<Napi::Value>;
+  *_ctx = Persistent(info[0]);
+
+  _tsfn = ThreadSafeFunction::New(
+      info.Env(), Function::New(env, [](const CallbackInfo & /*info*/) {}),
+      Object::New(env), "Test", 1, 1, _ctx,
+      [this](Napi::Env env, Reference<Napi::Value> *ctx) {
+        _deferred->Resolve(env.Undefined());
+        _deferred.reset();
```
legendecas: I don't think reset is necessary here as the unique ptr would call the deleter on destruct.

KevinEady: :+1: will update in later push, when master is updated with #583 since this PR conflicts

KevinEady:
Hey @mhdawson  / @gabrielschulhof ,

On the same basis of #566 (see https://github.com/nodejs/node-addon-api/pull/566#issuecomment-544637060) , would this also be considered SemVer major? If so, should we put this in to the 2.0.0 release (the scope seems to be ever increasing...)

In this case, "should not affect any code": If you were using `GetContext()` after an error and you still used the resulting `Context*` pointer, you'd probably crash with an error anyway?

This PR also directly conflicts with #580 because of the modifications of the test files, and as 580 is higher prio, I can't update this until that PR is merged.

mhdawson:
@KevinEady do you know when you will be able to push the change to address the outstanding comment?

test/threadsafe_function/threadsafe_function_ctx.cc
KevinEady: Updated

KevinEady:
Hi @mhdawson , just pushed!

mhdawson:
Landed as c881168d4951e833fb8275b233fa181ec0b60b43

Issue #582, opened 2019-10-31
Allow overriding AsyncWorker::OnWorkComplete etc


legendecas:
As for now, `AsyncWorker::OnWorkComplete` and `AsyncWorker::OnExecute` was declared as a static private method. We could make a global static function to accept an arbitrary AsyncWorker variant and invoke its own `OnWorkComplete` and `OnExecute` to make override possible.

This would be a blocking issue to implement AsyncProgressQueueWorker, which expects a queue cleanup before invoking either `OnOk` and `OnError` in `OnWorkComplete`.

legendecas:
Closed with #589

PR #585, opened 2019-11-02
Implement AsyncProgressQueueWorker

legendecas:
Although this patch comes along with a lot of changes on either `Napi::AsyncWorker` and `Napi::AsyncProgressWorker`, there are no changes in the existing code expected. (i.e. tests of `Napi::AsyncWorker` and `Napi::AsyncProgressWorker` shall pass without any modification)

Fixes: #582 

- [x] await TSFN queue clean up to call OnWorkComplete (https://github.com/nodejs/node/issues/31531)
- [x] npm test passed
- [x] test included
- [x] document changed

### DIFF
--- a/doc/async_worker_variants.md
+++ b/doc/async_worker_variants.md
@@ -344,10 +344,10 @@ method that will queue the created worker for execution.
 # AsyncProgressQueueWorker
 
 `Napi::AsyncProgressQueueWorker` acts exactly like `Napi::AsyncProgressWorker`
-except that each progress commited by `Napi::AsyncProgressQueueWorker::ExecutionProgress::Send`
+except that each progress committed by `Napi::AsyncProgressQueueWorker::ExecutionProgress::Send`
 during `Napi::AsyncProgressQueueWorker::Execute` is guaranteed to be
 processed by `Napi::AsyncProgressQueueWorker::OnProgress` on JavaScript thread
-by committing order order.
+by committing order.
 
 For the most basic use, only the `Napi::AsyncProgressQueueWorker::Execute` and
 `Napi::AsyncProgressQueueWorker::OnProgress` method must be implemented in a subclass.
@@ -368,9 +368,9 @@ The data pointed to will be copied to internal slots of `Napi::AsyncProgressQueu
 after the call to `Napi::AsyncProgressQueueWorker::ExecutionProcess::Send` the data can
 be safely released.
 
-`Napi::AsyncProgressQueueWorker::ExecutionProcess::Send` guarantees **eventual**
-invocation of `Napi::AsyncProgressQueueWorker::OnProgress`, which means
-multiple send will be cast to orderly invocation of `Napi::AsyncProgressQueueWorker::OnProgress`
+`Napi::AsyncProgressQueueWorker::ExecutionProcess::Send` guarantees invocation
+of `Napi::AsyncProgressQueueWorker::OnProgress`, which means multiple send will
+be cast to orderly invocation of `Napi::AsyncProgressQueueWorker::OnProgress`
 with each data.
 
 ```cpp
--- a/doc/async_worker_variants.md
+++ b/doc/async_worker_variants.md
@@ -346,8 +346,8 @@ method that will queue the created worker for execution.
 `Napi::AsyncProgressQueueWorker` acts exactly like `Napi::AsyncProgressWorker`
 except that each progress committed by `Napi::AsyncProgressQueueWorker::ExecutionProgress::Send`
 during `Napi::AsyncProgressQueueWorker::Execute` is guaranteed to be
-processed by `Napi::AsyncProgressQueueWorker::OnProgress` on JavaScript thread
-by committing order.
+processed by `Napi::AsyncProgressQueueWorker::OnProgress` on the JavaScript
+thread in the order it was committed.
 
 For the most basic use, only the `Napi::AsyncProgressQueueWorker::Execute` and
 `Napi::AsyncProgressQueueWorker::OnProgress` method must be implemented in a subclass.
@@ -369,9 +369,9 @@ after the call to `Napi::AsyncProgressQueueWorker::ExecutionProcess::Send` the d
 be safely released.
 
 `Napi::AsyncProgressQueueWorker::ExecutionProcess::Send` guarantees invocation
-of `Napi::AsyncProgressQueueWorker::OnProgress`, which means multiple send will
-be cast to orderly invocation of `Napi::AsyncProgressQueueWorker::OnProgress`
-with each data.
+of `Napi::AsyncProgressQueueWorker::OnProgress`, which means multiple `Send`
+call will result in the in-order invocation of `Napi::AsyncProgressQueueWorker::OnProgress`
+with each data item.
 
 ```cpp
 void Napi::AsyncProgressQueueWorker::ExecutionProcess::Send(const T* data, size_t count) const;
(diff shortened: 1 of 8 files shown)


legendecas:
Shall hold until we landed #589.

mhdawson:
@legendecas https://github.com/nodejs/node-addon-api/pull/589 has now landed so you can rebase/proceed.

mhdawson:
Discussed in last N-API team meeting, still one aspect to be completed, but @legendecas would like feedback on what is there so far.

mhdawson:
@nodejs/addon-api this is ready for review.

test/asyncprogressqueueworker.cc
```diff
@@ -0,0 +1,90 @@
...
+
+class TestWorker : public AsyncProgressQueueWorker<ProgressData> {
+public:
+  static Napi::Value CreateWork(const CallbackInfo& info) {
+    int32_t times = info[0].As<Number>().Int32Value();
+    Function cb = info[1].As<Function>();
+    Function progress = info[2].As<Function>();
+
+    TestWorker* worker = new TestWorker(cb, progress, "TestResource", Object::New(info.Env()));
+    worker->_times = times;
```
gabrielschulhof: Can this not be passed to the constructor instead of being assigned afterwards?

test/asyncprogressqueueworker.cc
```diff
@@ -0,0 +1,90 @@
...
+
+    return Napi::External<TestWorker>::New(info.Env(), worker);
+  }
+
+  static Napi::Value QueueWork(const CallbackInfo& info) {
+    auto wrap = info[0].As<Napi::External<TestWorker>>();
+    auto worker = wrap.Data();
+    worker->Queue();
+    return Napi::Boolean::New(info.Env(), true);
```
gabrielschulhof: Why return `true` here? There's no possibility that this will return `false`. Therefore the `assert` on the JS side is superfluous. Might as well make the function void.

test/asyncprogressqueueworker.cc
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  static Napi::Value CancelWork(const CallbackInfo& info) {
```
gabrielschulhof: This function can also be void. No need to return a value.

gabrielschulhof: OTOH, if you wish to test that cancelling work works, then `Queue()` and/or `Cancel()` must return a value which will be passed through to JS here.

test/asyncprogressqueueworker.cc
```diff
@@ -0,0 +1,90 @@
...
+    using namespace std::chrono_literals;
+    std::this_thread::sleep_for(1s);
+
+    if (_times < 0) {
+      SetError("test error");
+    }
+    ProgressData data{0};
+    for (int32_t idx = 0; idx < _times; idx++) {
+      data.progress = idx;
+      progress.Send(&data, 1);
```
gabrielschulhof: I think we should copy here and free in the JS thread.

test/asyncprogressqueueworker.cc
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  void OnProgress(const ProgressData* data, size_t /* count */) override {
+    Napi::Env env = Env();
+    if (!_progress.IsEmpty()) {
+      Number progress = Number::New(env, data->progress);
+      _progress.MakeCallback(Receiver().Value(), { progress });
```
gabrielschulhof: You shouldn't need to `MakeCallback` here because the TSFN does this for you.

test/asyncprogressqueueworker.cc
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+private:
+  TestWorker(Function cb, Function progress, const char* resource_name, const Object& resource)
+    : AsyncProgressQueueWorker(cb, resource_name, resource) {
+    _progress.Reset(progress, 1);
+  }
+
+  int32_t _times;
+  FunctionReference _progress;
```
gabrielschulhof: This variable could be named better, like, maybe `_js_progress_cb`.

mhdawson:
I'm not sure putting the doc for both classes into the same file make sense. I think it would stick to having one file to doc each class as we have for most classes. @NickNaso what do you think?

napi-inl.h
```diff
@@ -4587,9 +4587,73 @@ inline void ThreadSafeFunction::CallJS(napi_env env,
...
+template <typename DataType>
+inline void AsyncProgressWorkerBase<DataType>::OnWorkComplete(Napi::Env /* env */, napi_status status) {
+  _work_completed = true;
+  _complete_status = status;
+  _tsfn.Release();
+}
+
+template <typename DataType>
+inline void AsyncProgressWorkerBase<DataType>::OnThreadSafeFunctionFinalize(Napi::Env env, void* /* data */, AsyncProgressWorkerBase* context) {
+  if (context->_work_completed) {
```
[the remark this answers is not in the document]
legendecas: There might be cases that the worker is been released but doesn't been queued and executed. So `CHECK` will not be appropriate here IMO.

napi-inl.h
```diff
@@ -4684,20 +4739,18 @@ inline void AsyncProgressWorker<T>::Execute() {
...
-    std::lock_guard<std::mutex> lock(self->_mutex);
-    data = self->_asyncdata;
-    size = self->_asyncsize;
-    self->_asyncdata = nullptr;
-    self->_asyncsize = 0;
+    std::lock_guard<std::mutex> lock(this->_mutex);
+    data = this->_asyncdata;
+    size = this->_asyncsize;
+    this->_asyncdata = nullptr;
+    this->_asyncsize = 0;
```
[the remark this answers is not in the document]
legendecas: Since this is previous work on AsyncProgressWorker, I opened an issue https://github.com/nodejs/node-addon-api/issues/677 to tracking it down.

legendecas:
> @mhdawson:  I'm not sure putting the doc for both classes into the same file make sense. I think it would stick to having one file to doc each class as we have for most classes. @NickNaso what do you think?

I put them in a single doc file because they shared much similar behavior. It can be verbose if we separate those docs into different files.

doc/async_worker_variants.md
```diff
@@ -341,4 +341,114 @@ asynchronous task ends and other data needed for the computation. Once created,
...
+    public:
+        EchoWorker(Function& callback, std::string& echo)
+        : AsyncProgressQueueWorker(callback), echo(echo) {}
+
+        ~EchoWorker() {}
+    // This code will be executed on the worker thread
+    void Execute(const ExecutionProgress& progress) {
+        // Need to simulate cpu heavy task
+        for (uint32_t i = 0; i < 100; ++i) {
+          progress.Send(&i, 1)
```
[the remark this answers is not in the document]
legendecas: Internally, the passed in data items are copied to internal slots and these allocations are managed by AsyncProgressQueueWorker.
Exposing passing raw pointers may have better control over the `Send` operation, but it also made the lifetime management of these data mandatory. Users have to manually delete the data pointers after the progress been handled.

I'm leaning on the current behavior (copy the data) and this is what NAN provides. WDYT?

napi.h
```diff
@@ -2242,8 +2249,55 @@ namespace Napi {
...
+  template <typename DataType>
+  class AsyncProgressWorkerBase : public AsyncWorker {
+    public:
+     virtual void OnWorkProgress(DataType* data) = 0;
+     class ThreadSafeData {
+       public:
+        ThreadSafeData(AsyncProgressWorkerBase* asyncprogressworker, DataType* data)
+          : _asyncprogressworker(asyncprogressworker), _data(data) {}
+
+        AsyncProgressWorkerBase* asyncprogressworker() { return _asyncprogressworker; };
```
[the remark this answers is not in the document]
legendecas: These methods are very simple and naïve field exposing. I'm noticed there are previous works that also inlined these methods for readability. Since there are no acutal logics in the block, I'd believe it's ok to put them here?

mhdawson:
Landed as ab018444aef47795771854d983b1367629696e15

gabrielschulhof:
According to git bisect, this causes #695. @legendecas, could you PTAL?

PR #590, opened 2019-11-06
src: explicitly disallow assign and copy

legendecas:
Use macros to explicitly disallow assign and copy.

- [x] `npm test` passed.

### DIFF
--- a/napi.h
+++ b/napi.h
@@ -91,6 +91,13 @@ static_assert(sizeof(char16_t) == sizeof(wchar_t), "Size mismatch between char16
 
 #endif // NAPI_CPP_EXCEPTIONS
 
+# define NAPI_DISALLOW_ASSIGN(CLASS) void operator=(const CLASS&) = delete;
+# define NAPI_DISALLOW_COPY(CLASS) CLASS(const CLASS&) = delete;
+
+#define NAPI_DISALLOW_ASSIGN_COPY(CLASS)  \
+    NAPI_DISALLOW_ASSIGN(CLASS)           \
+    NAPI_DISALLOW_COPY(CLASS)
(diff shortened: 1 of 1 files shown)


NickNaso:
Hi @legendecas,
the code it's ok and works, but is it really necessary disallow the copy constructor using a macro?  
In this way se avoid to manually disallow the copy constructor in 9 classes, but we will lose in code readability. 
What do you think?

legendecas:
@NickNaso I found that this is what's done in Node.js core, v8, and NAN, lots of C++ projects.

It can not be denied that with macros we could not get first sight of what's done after the macro expansion. Yet since disallowing assign/copy/move operations are so commonly used, and I cannot assume it would be easy to write such things every time without missing something. So a macro with a straightforward name describing what it would do will be helpful. And I'd assume it would not be confusing very much on this use case.

gabrielschulhof:
@legendecas could you please rebase?

legendecas:
@gabrielschulhof rebased :D

gabrielschulhof:
CI:

| Version | Job | Status |
|---:|---|---|
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1446/ | ✓ |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1447/ |  |
| v10.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1448/ |  |
| v8.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1449/ |  |

gabrielschulhof:
Landed in 79deefb6f3f83f2ad453da531824ad76fbb02509.

PR #604, opened 2019-11-22
Issue 602: Add template methods to ObjectWrap

### LINKED ISSUE #602: Avoid extra memory allocation with templates
I suppose that it is possible to add methods in the following way to avoid extra memory allocation: 

Example:

```cpp
template <class T> 
class ObjectWrap {
    using MethodType = Value (T::*)(const CallbackInfo &info);

    template <MethodType method>
    static auto ObjectWrap<T>::InstanceMethod2(const char* utf8name) -> ClassPropertyDescriptor<T>; 
};

template <typename T>
template <typename ObjectWrap<T>::MethodType method>
inline auto ObjectWrap<T>::InstanceMethod2(const char* utf8name) -> ClassPropertyDescriptor<T>  {
    napi_property_descriptor desc = {};
    desc.utf8name = utf8name;
    desc.method = [](napi_evn env, napi_callback_info args) {
        return details:;WrapCallback([&]{
            CallbackInfo cb_info(env, args);
            T* instance = Unwrap(cb_info.This().As<Object>());
            return (instance->*cb)(callbackInfo);
        });
    };
    return desc;
}

// How to use
class Test: public Napi::ObjectWrap<Test> {
public:
    static auto define_class(Napi::Env env) -> Napi::Function {
        return DefineClass(env, "Test", {
            InstanceMethod2<&Test::get_value>("get_value");
        });
    }

    auto get_value(const Napi::CallbackInfo &args) -> Napi::Value;
};
```

dmitryash:
ObjectWrap<T> was enhanced to support template methods for
defining properties and methods of JS class. Now C++ methods and
functions may be passed as template parameters for
ObjectWrap<T>::InstanceMethod, ObjectWrap<T>::StaticAccessor, etc.

There are several benefits:

  - no need to allocate extra memory for passing C++ function
    napi callback and use add_finalizer() to free memory;
  - a compiler can see whole chain of calls up to napi callback
    that may allow better optimisation.

Some examples:

```cpp
// Method
InstanceMethod<&MyClass::method>("method");

// Read-write property
InstanceAccessor<&MyClass::get, &MyClass::set>("rw_prop");

// Read-only property
InstanceAccessor<&MyClass::get>("ro_prop");
```

Closes #602.

Use squash method.

### DIFF
--- a/doc/class_property_descriptor.md
+++ b/doc/class_property_descriptor.md
@@ -26,9 +26,9 @@ class Example : public Napi::ObjectWrap<Example> {
 Napi::Object Example::Init(Napi::Env env, Napi::Object exports) {
     Napi::Function func = DefineClass(env, "Example", {
         // Register a class instance accessor with getter and setter functions.
-        InstanceAccessor("value", &Example::GetValue, &Example::SetValue),
-        // We can also register a readonly accessor by passing nullptr as the setter.
-        InstanceAccessor("readOnlyProp", &Example::GetValue, nullptr)
+        InstanceAccessor<&Example::GetValue, &Example::SetValue>("value"),
+        // We can also register a readonly accessor by omitting the setter.
+        InstanceAccessor<&Example::GetValue>("readOnlyProp")
     });
 
     constructor = Napi::Persistent(func);
--- a/doc/object_wrap.md
+++ b/doc/object_wrap.md
@@ -34,8 +34,8 @@ class Example : public Napi::ObjectWrap<Example> {
 Napi::Object Example::Init(Napi::Env env, Napi::Object exports) {
     // This method is used to hook the accessor and method callbacks
     Napi::Function func = DefineClass(env, "Example", {
-        InstanceMethod("GetValue", &Example::GetValue),
-        InstanceMethod("SetValue", &Example::SetValue)
+        InstanceMethod<&Example::GetValue>("GetValue"),
+        InstanceMethod<&Example::SetValue>("SetValue")
     });
 
     // Create a peristent reference to the class constructor. This will allow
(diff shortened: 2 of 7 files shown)


gabrielschulhof:
We should consider moving the existing methods to `napi-inl.deprecated.h` and mark them as deprecated. Or we should use a two-step process:
0. Mark them as deprecated in the docs
0. In 3.0.0, move them to `napi-inl.deprecated.h`
0. In 4.0.0, remove them

@nodejs/n-api what do y'all think?

legendecas:
IMO there are no really breaking changes landed on most recent major version v2. I'd prefer the latter less aggressive pattern for the common usage on `ObjectWrap`: no actual breaking changes would be landed on v2.

mhdawson:
I'm good with the move to napi-inl.deprecated.h, removing completely we can discuss later.

We talked about adding a new deprecation guard so that people who already have napi_deprecated are not affected, but people can opt-in to the the higher level of deprecation.

In this case we'll use:  napi_disable_deprecated_1

napi_disable_deprecated will map to napi_disabled_deprecated_0
napi_disable_deprecated_all will map to the current highest level

gabrielschulhof:
@mhdawson I think it would OK to land this PR as is, and then leave the move to deprecated to another PR.

gabrielschulhof:
CI against v8.x: https://ci.nodejs.org/job/node-test-node-addon-api-new/1216/

gabrielschulhof:
Here it is again:

CI against v8.x: https://ci.nodejs.org/job/node-test-node-addon-api-new/1218/

gabrielschulhof:
CI:

| Version | Job | Status |
|---:|---|---|
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1234/ |  |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1235/ |  |
| v10.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1236/ |  |
| v8.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1237/ |  |

gabrielschulhof:
Landed in ce91e14860ef6ee5c733e2a2e90263e30ac5038e.

PR #609, opened 2019-11-26
Correctly define copy assignment operators.

rolftimmermans:
Copy assignment operators should be defined as [`T operator=(const T& other)`](https://en.cppreference.com/w/cpp/language/copy_assignment). Failure to do so may render the type unusable in standard containers. Most copy assignment operators were deleted, but notably the `Error` class has one defined. As expected it leads to problems using it in some container types (I tried this with `std::optional<Napi::Error>` in clang). This fixes the definitions of all instances that I could find.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2120,7 +2120,7 @@ inline Error& Error::operator =(Error&& other) {
 inline Error::Error(const Error& other) : ObjectReference(other) {
 }
 
-inline Error& Error::operator =(Error& other) {
+inline Error& Error::operator =(const Error& other) {
   Reset();
 
   _env = other.Env();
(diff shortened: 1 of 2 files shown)


gabrielschulhof:
| version | job | status |
|---:|---|---|
| v8.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1206/ | .oOo. |
| v10.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1207/ | .oOo. |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1208/ | .oOo. |
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1209/ | .oOo. |

gabrielschulhof:
Landed in 734725e971146cb88dc75342d05c8d5a161ee41f.

PR #611, opened 2019-11-28
Disable caching in ArrayBuffer

tniessen:
Caching the data pointer and the byteLength in the `ArrayBuffer` class causes it to behave incorrectly when the buffer is detached, as demonstrated by the test case. The current behavior is dangerous since it depends on whether the `Napi::Value` was cast to an `Napi::ArrayBuffer` before or after being detached.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1330,7 +1330,7 @@ inline ArrayBuffer ArrayBuffer::New(napi_env env, size_t byteLength) {
   napi_status status = napi_create_arraybuffer(env, byteLength, &data, &value);
   NAPI_THROW_IF_FAILED(env, status, ArrayBuffer());
 
-  return ArrayBuffer(env, value, data, byteLength);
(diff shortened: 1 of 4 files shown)


tniessen:
@NickNaso Done. May I ask why that is necessary?

gabrielschulhof:
CI:

| Version | Job | Status |
|---:|---|---|
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1382/ | .oOo. |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1383/ | .oOo. |
| v10.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1384/ | .oOo. |
| v8.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1385/ | .oOo. |

gabrielschulhof:
Landed in b72f1d69786cf5092db0dd24352645a9652d4615.

Issue #612, opened 2019-11-30
[Docs] missing Array docs page.


zander:
If you go to https://github.com/nodejs/node-addon-api/blob/master/doc/object.md
you will see the sentence; 

> and extends [`Napi::Array`](array.md)

the 'array.md' is missing in the doc directory, though.

I can't find any docs on the normal arrays.

NickNaso:
Hi @zander,
the right link is https://github.com/nodejs/node-addon-api/blob/master/doc/basic_types.md#array sorry for the mistake.

mhdawson:
Believe this can be closed as PR #613 has laned.

PR #613, opened 2019-11-30
[doc] Fixed links to array documentation

NickNaso:
Fixed all the links to `Napi::Array` documentation.
Fixes issue https://github.com/nodejs/node-addon-api/issues/612

### DIFF
--- a/doc/object.md
+++ b/doc/object.md
@@ -2,7 +2,7 @@
 
 The `Napi::Object` class corresponds to a JavaScript object. It is extended by the following node-addon-api classes that you may use when working with more specific types:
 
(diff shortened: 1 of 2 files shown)


zander:
While you are in this file anyway.
The line you changed says "and extends". Which IMOHO is a bit confusing.

Maybe you can change that to say "is extended by".

> Napi::Value which is extended by Napi::Array

zander:
The dead link (to array.md) may be present in other docs pages as well.

I noticed it also in https://github.com/nodejs/node-addon-api/blob/master/doc/value.md

PR #617, opened 2019-12-01
Fix code format in tests

tniessen:
This fixes a few code formatting issues in `test/`.

### DIFF
--- a/test/basic_types/array.cc
+++ b/test/basic_types/array.cc
@@ -5,9 +5,9 @@ using namespace Napi;
 Value CreateArray(const CallbackInfo& info) {
   if (info.Length() > 0) {
     size_t length = info[0].As<Number>().Uint32Value();
-	  return Array::New(info.Env(), length);
+    return Array::New(info.Env(), length);
   } else {
-	  return Array::New(info.Env());
+    return Array::New(info.Env());
   }
 }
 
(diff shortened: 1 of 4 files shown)


legendecas:
Maybe linters can be introduced to the project?

tniessen:
That is generally a good idea!

NickNaso:
We could use [clang-format](https://www.npmjs.com/package/clang-format). We can install it like dev dependencies. What do you think?

tniessen:
I have never used clang-format. Most Node.js projects use ESLint for the JS parts, and in core we use [cpplint](https://github.com/nodejs/node/blob/master/tools/cpplint.py) for C++. Seems like we also have [a partial ruleset for clang-format](https://github.com/nodejs/node/blob/master/.clang-format), though.

mhdawson:
I think this can land while we figure out if we can add linters.

gabrielschulhof:
CI:

| Version | Job | Status |
|---:|---|---|
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1378/ | ✓ |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1379/ | ✓ |
| v10.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1380/ | ✓ |
| v8.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/1381/ | ✓ |

gabrielschulhof:
Landed in 0e7483eb7b926e67db83632a5a5baf9c9b8d525f.

Issue #635, opened 2019-12-16
Exceptions in constructors causes v8 GC to fail with access violations


blagoev:
Hey folks,
If an exception is thrown from inside a constructor it seems it causes v8 GC to fail with an access violation in this case below. I am starting to think that this is caused by ```ObjectWrap<T>::FinalizeCallback``` being called when it shouldn't be.

Here is the example code if you want to try it out quickly https://github.com/blagoev/node-addon-examples/commit/351752fab1b93068aaf84a9cc52d086170778f74

The highlights are these.

There is a static factory function on the exposed object

```C++
Napi::Value create(const Napi::CallbackInfo& info) {
	Napi::EscapableHandleScope scope(info.Env());
	Napi::Value val = MyObject::constructor.New({}); //this will throw an exception

	return scope.Escape(info.Env().Undefined());
}
```

then the ctor throws some error

```C++
MyObject::MyObject(const Napi::CallbackInfo& info)
	: Napi::ObjectWrap<MyObject>(info) {
	Napi::Env env = info.Env();
	Napi::HandleScope scope(env);
	throw Napi::Error::New(env, "TEST");
......
```

and we have this private field in the class
```C++
struct Data {
};

class MyObject : public Napi::ObjectWrap<MyObject> {
private:
	std::shared_ptr<Data> m_data;
....
```

and this is the JS code that tries to call that
```JavaScript
var addon = require('bindings')('addon');
try {
    var obj = addon.MyObject.create();
    }
    catch (e) {
        console.error("Incorrect expected exception message:" + e.message);
    }
  
    global.gc()
    global.gc()
    global.gc()
```

then there is an exception here 

```
>	addon.node!std::_Ref_count_base::_Decref() Line 845	C++
 	addon.node!std::_Ptr_base<Data>::_Decref() Line 1122	C++
 	addon.node!std::shared_ptr<Data>::~shared_ptr<Data>() Line 1403	C++
 	addon.node!MyObject::~MyObject()	C++
 	addon.node!MyObject::`scalar deleting destructor'(unsigned int)	C++
 	addon.node!Napi::ObjectWrap<MyObject>::FinalizeCallback(napi_env__ * __formal, void * data, void * __formal) Line 3433	C++
 	node.exe!`anonymous namespace'::v8impl::Reference::SecondPassCallback(const v8::WeakCallbackInfo<`anonymous namespace'::v8impl::Reference> & data) Line 496	C++
[...]

PR #638, opened 2019-12-20
fixes v8 GC access violation for zombie objects

blagoev:
fix v8 GC access violation after napi ObjectWrap instance is left in zombie state by an exception in its constructor

closes this https://github.com/nodejs/node-addon-api/issues/635

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3697,7 +3697,7 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
   }
 
   T* instance;
-  napi_value wrapper = details::WrapCallback([&] {
+  napi_value wrapper = details::WrapCallback([&] () -> napi_value {
     CallbackInfo callbackInfo(env, info);
     callbackInfo.zombie = new Zombie();
 #ifdef NAPI_CPP_EXCEPTIONS
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2711,6 +2711,27 @@ inline Object FunctionReference::New(const std::vector<napi_value>& args) const 
   return scope.Escape(Value().New(args)).As<Object>();
 }
 
+class FinalizerHint {
+  private:
+    FinalizerHint() {}
+  public:
+    bool shouldFinalize;
+    
+    static void Init(CallbackInfo& info, bool shouldFinalize = true) {
+      info.finalizerHint = new FinalizerHint(); 
+      info.finalizerHint->shouldFinalize = shouldFinalize;
+    }
+
+    static void Clear(CallbackInfo& info) {
+      info.finalizerHint = nullptr;
+    }
+    
+    static FinalizerHint* Get(const CallbackInfo& info) {
+      return info.finalizerHint;
+    }
+};
+
+
 ////////////////////////////////////////////////////////////////////////////////
 // CallbackInfo class
 ////////////////////////////////////////////////////////////////////////////////
@@ -3120,7 +3141,12 @@ inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
   napi_status status;
   napi_ref ref;
   T* instance = static_cast<T*>(this);
-  status = napi_wrap(env, wrapper, instance, FinalizeCallback, (void*)callbackInfo.zombie, &ref);
+  FinalizerHint* finalizerHint = FinalizerHint::Get(callbackInfo);
+  status = napi_wrap(env, wrapper, instance, FinalizeCallback, (void*)finalizerHint, &ref);
+  if (status != napi_ok && finalizerHint != nullptr) {
+      FinalizerHint::Clear(const_cast<Napi::CallbackInfo&>(callbackInfo));
+      delete finalizerHint;
+  }
   NAPI_THROW_IF_FAILED_VOID(env, status);
 
   Reference<Object>* instanceRef = instance;
@@ -3699,21 +3725,26 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
   T* instance;
   napi_value wrapper = details::WrapCallback([&] () -> napi_value {
     CallbackInfo callbackInfo(env, info);
-    callbackInfo.zombie = new Zombie();
+    FinalizerHint::Init(callbackInfo);
 #ifdef NAPI_CPP_EXCEPTIONS
     try {
       instance = new T(callbackInfo);
       return callbackInfo.This();
     }
     catch (...) {
-      callbackInfo.zombie->isZombie = true;
+      FinalizerHint* finalizerHint = FinalizerHint::Get(callbackInfo);
+      if (finalizerHint != nullptr) {
+         finalizerHint->shouldFinalize = false;
+      }
       throw;
     }
 #else
     instance = new T(callbackInfo);
     if (callbackInfo.Env().IsExceptionPending()) {
-      callbackInfo.zombie->isZombie = true;
-      return nullptr;
+      FinalizerHint* finalizerHint = FinalizerHint::Get(callbackInfo);
+      if (finalizerHint != nullptr) {
+         finalizerHint->shouldFinalize = false;
+      }
     }
     return callbackInfo.This();
 #endif
@@ -3841,10 +3872,10 @@ inline napi_value ObjectWrap<T>::InstanceSetterCallbackWrapper(
 template <typename T>
 inline void ObjectWrap<T>::FinalizeCallback(napi_env env, void* data, void* hint) {
  if (hint != nullptr) {
-    Zombie* zombie = (Zombie*)hint;
-    bool isZombie = zombie->isZombie;
-    delete zombie;
-    if (isZombie) {
+    FinalizerHint* finalizerHint = (FinalizerHint*)hint;
+    bool shouldFinalize = finalizerHint->shouldFinalize;
+    delete finalizerHint;
+    if (!shouldFinalize) {
       return;
     }
   }
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3700,6 +3700,7 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
   napi_value wrapper = details::WrapCallback([&] {
     CallbackInfo callbackInfo(env, info);
     callbackInfo.zombie = new Zombie();
+#ifdef NAPI_CPP_EXCEPTIONS
     try {
       instance = new T(callbackInfo);
       return callbackInfo.This();
@@ -3708,6 +3709,14 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
       callbackInfo.zombie->isZombie = true;
       throw;
     }
+#else
+    instance = new T(callbackInfo);
+    if (callbackInfo.Env().IsExceptionPending()) {
+      callbackInfo.zombie->isZombie = true;
+      return nullptr;
+    }
+    return callbackInfo.This();
+#endif
   });
 
   return wrapper;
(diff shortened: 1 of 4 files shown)


blagoev:
Naming is intentionally left colorful. 
Renaming to something like `FinalizerCallbackHint` or something like that will be more generic

gabrielschulhof:
@blagoev does https://github.com/nodejs/node-addon-api/pull/600 not also fix this?

gabrielschulhof:
@blagoev could you please add a test wherein the subclass throws an exception in the constructor?

blagoev:
I will try to add a test for this since the #600 discussion seems did not result in a conclusion. And my version seems simpler. Still I will leave it up for a discussion which version should be merged.

napi-inl.h
```diff
@@ -3823,7 +3839,16 @@ inline napi_value ObjectWrap<T>::InstanceSetterCallbackWrapper(
 }
 
 template <typename T>
-inline void ObjectWrap<T>::FinalizeCallback(napi_env env, void* data, void* /*hint*/) {
+inline void ObjectWrap<T>::FinalizeCallback(napi_env env, void* data, void* hint) {
```
gabrielschulhof: IINM, if `napi_wrap()` fails then this will never be called, and `zombie` will never be freed, resulting in a leak.

napi-inl.h
```diff
@@ -3697,10 +3697,26 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
...
+    }
+#else
     instance = new T(callbackInfo);
+    if (callbackInfo.Env().IsExceptionPending()) {
+      callbackInfo.zombie->isZombie = true;
+      return nullptr;
+    }
     return callbackInfo.This();
+#endif
   });
```
gabrielschulhof: I think it might be best to free `zombie` here and not pass it as the `hint` to the finalize callback, because, if `napi_wrap()` fails, that callback will not be called.

napi-inl.h
```diff
@@ -3120,7 +3120,7 @@ inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
   napi_status status;
   napi_ref ref;
   T* instance = static_cast<T*>(this);
-  status = napi_wrap(env, wrapper, instance, FinalizeCallback, nullptr, &ref);
+  status = napi_wrap(env, wrapper, instance, FinalizeCallback, (void*)callbackInfo.zombie, &ref);
   NAPI_THROW_IF_FAILED_VOID(env, status);
```
gabrielschulhof:
```suggestion
  if (status != napi_ok) {
    ping_pong->object_wrapping_failed = true;
  }
  NAPI_THROW_IF_FAILED_VOID(env, status);
```

napi-inl.h
```diff
@@ -3697,10 +3697,26 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
   }
 
   T* instance;
-  napi_value wrapper = details::WrapCallback([&] {
+  napi_value wrapper = details::WrapCallback([&] () -> napi_value {
```
gabrielschulhof:
```suggestion
  napi_value wrapper = details::WrapCallback([&] {
```

We don't need this change, because it's OK to return `callbackInfo.This()`. After all, when control returns to the engine it will throw anyway, so the actual instance will not be relevant, and there won't be a leak on the native side either.

napi-inl.h
```diff
@@ -3697,10 +3697,26 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
...
+    }
+    catch (...) {
+      callbackInfo.zombie->isZombie = true;
+      throw;
+    }
+#else
     instance = new T(callbackInfo);
+    if (callbackInfo.Env().IsExceptionPending()) {
+      callbackInfo.zombie->isZombie = true;
+      return nullptr;
```
gabrielschulhof:
```suggestion
```

It's OK to fall through to the return below.

test/objectwrap.cc
```diff
@@ -172,11 +172,30 @@ class Test : public Napi::ObjectWrap<Test> {
 
 std::string Test::s_staticMethodText;
 
+#ifdef NAPI_CPP_EXCEPTIONS
```
gabrielschulhof: We should perform this test in both cases – if C++ exceptions are enabled, *as well as* if they are not enabled.

napi-inl.h
```diff
@@ -3120,7 +3120,7 @@ inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
   napi_status status;
   napi_ref ref;
   T* instance = static_cast<T*>(this);
-  status = napi_wrap(env, wrapper, instance, FinalizeCallback, nullptr, &ref);
+  status = napi_wrap(env, wrapper, instance, FinalizeCallback, (void*)callbackInfo.zombie, &ref);
```
gabrielschulhof:
```suggestion
  ObjectWrapPingPong* ping_pong = static_cast<ObjectWrapPingPong*>(callbackInfo.Data());
  callbackInfo.SetData(ping_pong->data);
  status = napi_wrap(env, wrapper, instance, FinalizeCallback, nullptr, &ref);
```

gabrielschulhof: We can declare
```C++
struct ObjectWrapPingPong {
  void* data = nullptr;
  bool object_wrapping_failed = false;
};
```
in the private section of `ObjectWrap<T>`.

gabrielschulhof:
@dananderson @dmitryash WDYT about this approach?

gabrielschulhof:
@blagoev argh! GitHub shows the comments in a weird order 😕 Best to look at the "Files changed" tab.

blagoev:
@gabrielschulhof The napi_wrap failing was good catch. 

I was going to propose fixing this with 
```C++
status = napi_wrap(env, wrapper, instance, FinalizeCallback, (void*)callbackInfo.zombie, &ref);
  if (status != napi_ok && callbackInfo.zombie != nullptr) {
      auto zombie = callbackInfo.zombie;
      callbackInfo.zombie = nullptr; 
      delete zombie;
  }
```

I thought of doing it with SetData but then this is user API which might be already used for something else. So we would overwrite user data during the call. 

My proposal is dictated mostly with what the API already provides (ie the Finalizer hint field and the Callbackinfo passed as only argument) without introducing side effects to the user. The only thing I don't like (and you correctly pointed it out) is having this public field to handle something internal. But it's caused by this desire to make it work within the confines of the current API.

I am not at all attached to my approach and I am for finding the best approach for this. So maybe if we hide this public field and access it with a friend internal class or something we can do it better.

gabrielschulhof:
> @gabrielschulhof The napi_wrap failing was good catch.
> 
> I was going to propose fixing this with
> 
> ```c++
> status = napi_wrap(env, wrapper, instance, FinalizeCallback, (void*)callbackInfo.zombie, &ref);
>   if (status != napi_ok && callbackInfo.zombie != nullptr) {
>       auto zombie = callbackInfo.zombie;
>       callbackInfo.zombie = nullptr; 
>       delete zombie;
>   }
> ```

I still think we should avoid heap-allocating, because we can resolve the situation without needing a finalizer callback. Also, if we `napi_remove_wrap()` if the wrapping succeeds but the construction fails then we can avoid the need for adding to the finalizer callback.

> 
> I thought of doing it with SetData but then this is user API which might be already used for something else. So we would overwrite user data during the call.

AFAICT no user code gets executed between the ping of setting the `ObjectWrapPingPong` as the `callbackInfo` data and the pong of setting it back to the user data because the `ObjectWrap<T>` constructor runs before the user subclass' constructor.

> My proposal is dictated mostly with what the API already provides (ie the Finalizer hint field and the Callbackinfo passed as only argument) without introducing side effects to the user. The only thing I don't like (and you correctly pointed it out) is having this public field to handle something internal. But it's caused by this desire to make it work within the confines of the current API.
> 
> I am not at all attached to my approach and I am for finding the best approach for this. So maybe if we hide this public field and access it with a friend internal class or something we can do it better.

A solution based on friend classes would definitely be better than the ping pong, even if I'm right about there being no user code between the ping and the pong.

Your approach is good, and I added the elements from #600 (namely the `napi_remove_wrap()`) as suggestions. Let's go with it.

blagoev:
@gabrielschulhof I have refactored this a bit. Please share your thoughts.

gabrielschulhof:
@blagoev I updated the implementation in https://github.com/nodehs/node-addon-api/pull/600/files to use friend classes while avoiding heap allocation. I gave you co-authorship on the commit because I used friend classes the same way you used them in this PR. Please let me know what you think!

napi.h
```diff
@@ -1397,6 +1397,7 @@ namespace Napi {
   };
 
   class CallbackInfo {
+    friend class FinalizerHint;
```
gabrielschulhof: You need to forward-declare this class at the top of the file.

napi-inl.h
```diff
@@ -3120,7 +3141,12 @@ inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
   napi_status status;
   napi_ref ref;
   T* instance = static_cast<T*>(this);
-  status = napi_wrap(env, wrapper, instance, FinalizeCallback, nullptr, &ref);
+  FinalizerHint* finalizerHint = FinalizerHint::Get(callbackInfo);
+  status = napi_wrap(env, wrapper, instance, FinalizeCallback, (void*)finalizerHint, &ref);
```
gabrielschulhof: Please avoid C-style casts!

napi-inl.h
```diff
@@ -3697,10 +3723,31 @@ inline napi_value ObjectWrap<T>::ConstructorCallbackWrapper(
...
+      if (finalizerHint != nullptr) {
+         finalizerHint->shouldFinalize = false;
+      }
+      throw;
+    }
+#else
     instance = new T(callbackInfo);
+    if (callbackInfo.Env().IsExceptionPending()) {
+      FinalizerHint* finalizerHint = FinalizerHint::Get(callbackInfo);
+      if (finalizerHint != nullptr) {
```
gabrielschulhof: This check should be a hard check. If `finalizerHint` is NULL, we must go to `Napi::Error::Fatal`. After all, we set it above.

test/objectwrap.cc
```diff
@@ -172,11 +172,35 @@ class Test : public Napi::ObjectWrap<Test> {
...
 Napi::Object InitObjectWrap(Napi::Env env) {
   testStaticContextRef = Napi::Persistent(Napi::Object::New(env));
   testStaticContextRef.SuppressDestruct();
 
   Napi::Object exports = Napi::Object::New(env);
   Test::Initialize(env, exports);
+
+#ifdef NAPI_CPP_EXCEPTIONS
+  TestConstructorExceptions::Initialize(env, exports);
+#endif
```
gabrielschulhof: This initialization should be unconditional.

test/objectwrap.js
```diff
@@ -256,9 +256,19 @@ const test = (binding) => {
     testFinalize(clazz);
   };
 
+  const testConstructorExceptions = () => {
+    const TestConstructorExceptions = binding.objectwrap.TestConstructorExceptions;
+    console.log("Runnig test testConstructorExceptions");
```
gabrielschulhof:
```suggestion
```

The console log is unnecessary.

test/objectwrap.js
```diff
@@ -256,9 +256,19 @@ const test = (binding) => {
...
+  }
+
   // `Test` is needed for accessing exposed symbols
   testObj(new Test(), Test);
   testClass(Test);
+
+  testConstructorExceptions();
 }
 
 test(require(`./build/${buildType}/binding.node`));
```
gabrielschulhof:
```suggestion
test(require(`./build/${buildType}/binding.node`));
test(require(`./build/${buildType}/binding_noexcept.node`));
```

Please test both scenarios!

blagoev:
closing this in favor of #600

PR #649, opened 2020-01-15
test: update BigInt test for recent change in core

mhdawson:
https://github.com/nodejs/node/commit/689ab46c646bc8add5560a606ba73d3760f8f924
changed the expected error for one of the BigInt test cases. This is
ok because BigInt is still in experimental. However, the
result was a failed/hanging test. See
https://github.com/nodejs/build/issues/2131

This changes the test to accept either the new or old behaviour.
We need that so that older versions of Node.js will also pass the
test until the the node core commit above is backported.

### DIFF
--- a/test/bigint.js
+++ b/test/bigint.js
@@ -46,7 +46,7 @@ function test(binding) {
   });
 
   assert.throws(TestTooBigBigInt, {
-    name: 'RangeError',
-    message: 'Maximum BigInt size exceeded',
+    name: /^(RangeError|Error)$/,
+    message: /^(Maximum BigInt size exceeded|Invalid argument)$/,
   });
 }


mhdawson:
I'd like to fast track landing this as it is breaking the nightly testing and worse hanging and keeping machines busy.

richardlau:
> 
> 
> I'd like to fast track landing this as it is breaking the nightly testing and worse hanging and keeping machines busy.

If nobody has any objections I'd like to enable timeouts on the node-test-node-addon-api-new job so that in the future any hangs won't tie up infrastructure until someone manually aborts the job: https://github.com/nodejs/build/issues/2131#issuecomment-574380713

gabrielschulhof:
@mhdawson this is ready to go.

mhdawson:
Landed as 2fde5c3ca3bd2fefb87be1fc897e0ae606531298

PR #650, opened 2020-01-17
Fix syntax error in example.

timarandras:
Add override specifier to vitual functions.


gabrielschulhof:
Landed in 0f8d73048330fa11b2218150e7b5ac4f2ee36e27.

Issue #653, opened 2020-01-18
Context parameter on thread safe function callback


legendecas:
Currently `Napi::ThreadSafeFunction` accepts a type implementing `void operator()(Napi::Env env, Function jsCallback, DataType* data)`. It would be hard to retrieve the initial context parameter on `Napi::ThreadSafeFunction::New` since there is only one parameter `DataType* data` or we have to wrap the data with a reference to context _again_.

legendecas:
As a reference `napi_threadsafe_function_call_js` has a signature of

```
typedef void (*napi_threadsafe_function_call_js)(napi_env env,
                                                 napi_value js_callback,
                                                 void* context,
                                                 void* data);
```

KevinEady:
Looking at the signature for our `CallJS` wrapper passed to `napi_create_threadsafe_function`:

https://github.com/nodejs/node-addon-api/blob/4e885069f1721dec1705c898383625dd2907f9e6/napi-inl.h#L4572-L4575

... we completely ignore the `context` data type.

The context is not wrapped, so one could use `ThreadSafeFunction::GetContext()` but it is an additional call that shouldn't be necessary.

legendecas:
> The context is not wrapped, so one could use ThreadSafeFunction::GetContext() but it is an additional call that shouldn't be necessary.

I've considered this option. It seems on the user callback we don't have the reference to call `ThreadSafeFunction::GetContext()` if we don't wrap it in the data. So the option still relies on wrapping around data.

gabrielschulhof:
From https://github.com/nodejs/node-addon-api/issues/594#issuecomment-581502550 it looks like we may need to substantially revamp `Napi::ThreadSafeFunction`. When we do so, we should also address this issue.

KevinEady:
Sooo I have been working on this for some time now... After struggling with templated std::functions for hours, I really see two ways of implementing:
1. Allow the existing `[Non]BlockingCall()`s to take a three-argument callback, with the context being `void*`. 
[...]

PR #676, opened 2020-02-24
[WIP] tsfn: Add context to [Non]BlockingCall callback

KevinEady:
This is WIP, currently missing the overload functionality for `[Non]BlockingCall(DataType*, Callback)`, to allow the Callback to have either signature `void(Napi::Env, Function, DataType*)` [existing functionality] or `void(Napi::Env, Function, DataType*, void*)` [new context param]

Todo:
- [ ] Add `[Non]BlockingCall` with DataType* overloads
- [ ] Update docs

Fixes: #653

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -4574,7 +4574,7 @@ inline napi_status ThreadSafeFunction::CallInternal(
 // static
 inline void ThreadSafeFunction::CallJS(napi_env env,
                                        napi_value jsCallback,
-                                       void* /* context */,
+                                       void* context,
                                        void* data) {
   if (env == nullptr && jsCallback == nullptr) {
     return;
@@ -4582,7 +4582,7 @@ inline void ThreadSafeFunction::CallJS(napi_env env,
 
   if (data != nullptr) {
     auto* callbackWrapper = static_cast<CallbackWrapper*>(data);
-    (*callbackWrapper)(env, Function(env, jsCallback));
+    (*callbackWrapper)(env, Function(env, jsCallback), context);
     delete callbackWrapper;
   } else if (jsCallback != nullptr) {
     Function(env, jsCallback).Call({});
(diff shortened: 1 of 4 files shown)


KevinEady:
I ran into this problem when finishing up the PR and now stuck...

Basically, the `[Non]BlockingCall` with data overload currently has this signature:

```cpp
template <typename DataType, typename Callback>
inline napi_status ThreadSafeFunction::BlockingCall(
    DataType* data, Callback callback)
```

and is used something like this:

```cpp
tsfn.BlockingCall(data, [](Napi::Env env, Function cb, DataType* data) { ... })
```

and we want to be able to do this...


```cpp
tsfn.BlockingCall(data, [](Napi::Env env, Function cb, DataType* data, void* context) { ... })
```

Similar to how I modified this [`CallbackWrapper` to be a class with two constructor signatures](https://github.com/nodejs/node-addon-api/pull/676/files#diff-8cd24746fa3a8de7425641e24a06b184R2218) (with or without void* context), I think I would need to create a _new_ template class for this specific Data callback...

```cpp
template <typename DataType>
class DataCallbackWrapper {
    private:
      using ContextlessCallback = std::function<void(Env, Function, DataType*)>;
      using ContextedCallback = std::function<void(Env, Function, DataType*, void *)>;

      ContextlessCallback ctxless;
      ContextedCallback ctxfull;

    public:
      CallbackWrapper(ContextlessCallback arg) : ctxless(arg){};
      CallbackWrapper(ContextedCallback arg) : ctxfull(arg){};

      void operator()(Napi::Env env, Napi::Function callback, DataType* data, void *context) {
        ctxfull ? ctxfull(env, callback, data, context) : ctxless(env, callback, data);
      };
    };
}
```

It just seems like a lot of wrapping and objects, I don't know if this is the right approach. Hopefully we can discuss in today's NAPI meeting.

KevinEady:
Closing this in favor of #687

Issue #556, opened 2019-10-05
ThreadSafeFunction(napi_threadsafe_function) works badly with BlockingCall()


o2genum:
When a `ThreadSafeFunction` instance is created from `napi_threadsafe_function`, its `BlockingCall` method doesn't work properly:
- The passed callback is ignored, but still required by the method signature (the one supplied with `napi_create_threadsafe_function` is used)
- The passed `data` pointer is incorrect for some reason. `napi_call_threadsafe_function` works fine.

I'm doing this because I want to `napi_unref_threadsafe_function` without using all the verbose C APIs.

KevinEady:
Hello @o2genum ,

Can you post some code that you are using? Here's some code using BlockingCall() in all three overloads:

https://github.com/nodejs/node-addon-api/blob/dd9fa8a4a80103b8fb5cea4a891c32d04d6ae1d2/test/threadsafe_function/threadsafe_function.cc#L52-L66

We have two tests: one using `BlockingCall` with no callback or data, and one using `BlockingCall` with data and callback, passing an `int*` to `callback`, which receives that value.

https://github.com/nodejs/node-addon-api/blob/e10e683a078a53be6984d2602e6d3be5da0d5d69/test/threadsafe_function/threadsafe_function_sum.cc#L35-L41

And here we use `BlockingCall()` with only callback and no data.

KevinEady:
Ahh, I am sorry, I misunderstood your issue. Using the node-addon-api with a pre-constructed `napi_threadsafe_function` fails. I will take a look. Thanks!

o2genum:
Nice! By the way, I want to note I could have done without wrapping TSFN from C API into C++ API one if there was a way to call `napi_unref_threadsafe_function` from C++ API, to prevent the TSFN from keeping event loop alive.

KevinEady:
Hi @o2genum ,

Yeah, so I'm not sure how we can wrap an existing `napi_tsfn` in a new `ThreadSafeFunction`... And yep, the main problem is with how the `TSFN` is not a "1:1 wrapper" with napi_tsfn functions I suppose.

[...]

PR #687, opened 2020-03-22
tsfn: implement ThreadSafeFunctionEx for GCC 5+

KevinEady:
This PR backs the existing `ThreadSafeFunction` class with the new `ThreadSafeFunctionEx` class. However, there are compilation issues on gcc-4.8 used by the CI.  node v10 has a minimum build requirement of [gcc-4.9](https://github.com/nodejs/node/blob/v10.x/BUILDING.md#unixmacos). 

We discussed somewhere (possibly an N-API meeting) that we can have this PR blocked until node 10 is decommissioned, so we can update our CI's gcc and possibly incorporate these changes.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -218,6 +218,44 @@ struct ThreadSafeFinalize {
   FinalizerDataType* data;
   Finalizer callback;
 };
+
+template <typename ContextType, typename DataType, typename CallJs, CallJs call>
+typename std::enable_if<call != nullptr>::type
+static inline CallJsWrapper(napi_env env, napi_value jsCallback, void *context, void *data) {
+  call(env, Function(env, jsCallback), static_cast<ContextType *>(context),
+       static_cast<DataType *>(data));
+}
+
+template <typename ContextType, typename DataType, typename CallJs, CallJs call>
+typename std::enable_if<call == nullptr>::type
+static inline CallJsWrapper(napi_env env, napi_value jsCallback, void * /*context*/,
+              void * /*data*/) {
+  if (jsCallback != nullptr) {
+    Function(env, jsCallback).Call(0, nullptr);
+  }
+}
+
+#if NAPI_VERSION > 4
+
+template <typename CallbackType, typename TSFN>
+napi_value DefaultCallbackWrapper(napi_env /*env*/, std::nullptr_t /*cb*/) {
+  return nullptr;
+}
+
+template <typename CallbackType, typename TSFN>
+napi_value DefaultCallbackWrapper(napi_env /*env*/, Napi::Function cb) {
+  return cb;
+}
+
+#else
+template <typename CallbackType, typename TSFN>
+napi_value DefaultCallbackWrapper(napi_env env, Napi::Function cb) {
+  if (cb.IsEmpty()) {
+    return TSFN::EmptyFunctionFactory(env);
+  }
+  return cb;
+}
+#endif  // NAPI_VERSION > 4
 #endif  // NAPI_VERSION > 3 && !defined(__wasm32__)
 
 template <typename Getter, typename Setter>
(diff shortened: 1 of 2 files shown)


napi-inl.h
```diff
@@ -4256,6 +4256,131 @@ inline void AsyncWorker::OnWorkComplete(Napi::Env /*env*/, napi_status status) {
...
+inline ThreadSafeFunctionEx<ContextType> ThreadSafeFunctionEx<ContextType>::New(napi_env env,
+                                                  const Function& callback,
+                                                  const Object& resource,
+                                                  ResourceString resourceName,
+                                                  size_t maxQueueSize,
+                                                  size_t initialThreadCount,
+                                                  ContextType* context,
+                                                  Finalizer finalizeCallback,
+                                                  FinalizerDataType* data,
+                                                  napi_threadsafe_function_call_js call_js_cb) {
```
legendecas: I believe it is worth to type the `napi_threadsafe_function_call_js` with `ContextType` and `DataType`.

KevinEady: @legendecas I was thinking that too, however looking at https://github.com/nodejs/node-addon-api/issues/594#issuecomment-581502550 , there were complaints that the "wrapping" for CallJS was a performance hit... and I can't think of a way to type `napi_threadsafe_function_call_js` without making a wrapper.

Furthermore, I don't think `DataType` can be typed without a wrapper, because it's "declared" in `::New` but "used" in `::[Non]BlockingCall`, and not sure how to link the two together. The way it works in the existing TSFN is that the `napi_threadsafe_function_call_js` takes this [`CallbackWrapper`](https://github.com/nodejs/node-addon-api/blob/89e62a9154b6057ed086a227db25caf26ac8c92a/napi-inl.h#L4585) data which is appropriately typed. Without a wrapper, not sure how to accomplish...

KevinEady:
Hi @mhdawson / @gabrielschulhof / @legendecas,

In yesterday's meeting we discussed how I removed the multiple overloads for the `New` method and only kept one that has _all_ parameters. Looking at https://github.com/nodejs/node-addon-api/issues/544, the last two comments highlight two additional concerns regarding the optional parameter overloads. I believe those are valid, and that's why we I feel we should not have all the possible overloads. This leaves it up to the developer to pass appropriate values. It would also inherently keep up with changing N-API functionality (eg. the new optional `Function` parameter). 

On the contrary, if we _did_ have the multiple overloads, we could N-API version target them via various preprocessor directives, eg. having `New` with no `Function` if NAPI > x. 

I do think we should just stick with one `New` method.

Thoughts?

legendecas:
@KevinEady I submitted another issue #690 to propose an alternative for the optional `Function` of `New` parameter. 

So with `Maybe` boxes, I'd believe we can lean on the one single `New`.

KevinEady:
So @gabrielschulhof / @legendecas  / @mhdawson ,

I was working on this over the weekend, and regarding the CallJS wrapper... 

To recall, we wanted to wrap it so the `New` function's `call_js_cb`

https://github.com/nodejs/node-addon-api/blob/90ebd80ae1f7d703b11529433c5cf3d72ee97491/napi.h#L2045-L2054

eg. callback

https://github.com/nodejs/node-addon-api/blob/90ebd80ae1f7d703b11529433c5cf3d72ee97491/test/threadsafe_function/threadsafe_function_ex.cc#L82-L85

would have signature
```cpp
void(Napi::Env, Napi::Function, ContextType *context, void *data)
```

However, if we do wrap the callback given to `napi_create_threadsafe_function`, you would not be able to get the wrapper `ThreadSafeFunctionEx<Context>`'s underlying `napi_threadsafe_function`  and use it directly with `napi_call_threadsafe_function`, eg. this behavior would not work:

```cpp
auto tsfn = Napi::ThreadSafeFunctionEx<Context>::New(env, ....);
napi_call_threadsafe_function(tsfn, ...)
```

I imagine this is currently an issue in the existing `ThreadSafeFunction` implementation as well. 

FWIW, there was a related problem https://github.com/nodejs/node-addon-api/issues/556 due to wrapping

KevinEady:
Hey @mhdawson / @legendecas / @gabrielschulhof ,

I was able to template the callback parameter as Gabe suggested (woohoo!), and I _think_ the API feels alright... I've included two test cases that show how to use the API:
1. [threadsafe_function_ex/context.cc](https://github.com/nodejs/node-addon-api/blob/561b9687ffe615fb1f41fecab7b18ff9710091b6/test/threadsafe_function_ex/context.cc) - showcases a use of ThreadSafeFunctionEx with all templated parameters (context, data type, call js)
2. [threadsafe_function_ex/simple.cc](https://github.com/nodejs/node-addon-api/blob/561b9687ffe615fb1f41fecab7b18ff9710091b6/test/threadsafe_function_ex/simple.cc) - a very basic tsfn that gives the bare-minimum options (except finalizer, working on that..)

What do you guys think?

gabrielschulhof:
@KevinEady could you please rebase this so I can try a CI run? As it stands, the CI is affected by https://github.com/nodejs/node-addon-api/issues/695 and our CI doesn't automatically rebase for testing like core's CI does.

KevinEady:
Hi @gabrielschulhof , I've updated with master. 

Also, I've added another test to showcase intended api some more -- [threadsafe_function_ex/call.cc](https://github.com/KevinEady/node-addon-api/blob/contexted-tsfn-api/test/threadsafe_function_ex/call.cc) with no context nor finalizer. /cc @mhdawson  @legendecas .

napi-inl.h
```diff
@@ -196,6 +196,22 @@ struct ThreadSafeFinalize {
   FinalizerDataType* data;
   Finalizer callback;
 };
+
+template <typename ContextType, typename DataType, typename CallJs, CallJs call>
+typename std::enable_if<call != nullptr>::type
+static inline CallJsWrapper(napi_env env, napi_value jsCallback, void *context, void *data) {
+  call(env, Function(env, jsCallback), static_cast<ContextType *>(context),
```
mhdawson, line 203: Does this need the  if (jsCallback != nullptr) check like the version below or do we expect the CallJs

mhdawson:
I assume we need some additions to the docs as well?

napi-inl.h
KevinEady: Hi @mhdawson ,

Your comment might have got cut off / incomplete but I think I got you...

I do **not** think we need the `jsCallback != nullptr` check like below. The call will go into the user-defined `CallJs call` with an empty function, which I believe is the normal behavior of using the underlying napi tsfn calls with no callback Function but with a CallJs.

KevinEady:
> The implementation looks good, though AFAICT none of the tests are actually multithreaded. It would be good to have the `Napi::ThreadSafeFunction`s tests reproduced here to ensure that the class deals correctly with threads.

Hi @gabrielschulhof / @mhdawson / @legendecas 

Since the `ThreadSafeFunctionEx` class can be considered the "base feature" class, I was able to make the existing `ThreadSafeFunction` class inherit from `Ex`, thus the existing tests inherently go through `Ex`. Does this satisfy? Please take a look... None of the public members were changed so I do not think this would affect any backwards compatibility

Well... Going through the different CI errors, messing with templates is not as easy as I imagined.

KevinEady:
Hi team,

Going through the errors seen here https://travis-ci.com/github/nodejs/node-addon-api/jobs/323508516#L743 ...

I noticed the node-addon-api CI uses [gcc 4.8.4](https://travis-ci.com/github/nodejs/node-addon-api/jobs/323508516#L601) but the minimum supported gcc version for building node is [gcc 6](https://github.com/nodejs/node/blob/master/BUILDING.md#supported-toolchains). Aaaand of course, when I test this PR on a gcc 6.1 docker image, it runs fine. 

Do we need to support gcc 4? Let's discuss in the next call.

richardlau:
> Do we need to support gcc 4? Let's discuss in the next call.

If you ever want this to land on 10.x then, yes:
https://github.com/nodejs/node/blob/v10.x/BUILDING.md#unix

mhdawson:
We did not discuss in the Node-api team meeting as we ran out of time, but I think since 10.x is officially not supported it can be revisited. See just above https://github.com/nodejs/node-addon-api#setup for where we have documented that.

PR #691, opened 2020-03-28
add doc auto deploy action

### LINKED ISSUE #674: Auto publishing doc using github actions.
With this we can simplify doc publish process.

Ref: https://github.com/JamesIves/github-pages-deploy-action.

gengjiawen:
Fix #674

See: http://www.gengjiawen.com/node-addon-api/.

After merging this, I will do add to `README.md` and other doc fix.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -1,0 +1,26 @@
+name: Node.js Addon API Doc Deploy
+
+on:
+  push:
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    container: gengjiawen/node-build
+    steps:
+      - name: Checkout 🛎️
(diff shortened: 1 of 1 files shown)


mhdawson:
Can you add a bit more details as to what this is doing?  At some point I think we were going to delete the generated doc, so not sure if this is related to that or not?

gengjiawen:
> At some point I think we were going to delete the generated doc

Why ? It the doc outdated or something ?

legendecas:
I believe the docs under `/doc` directory are recommended to use.

gengjiawen:
> I believe the docs under `/doc` directory are recommended to use.

The generated doc is `yarn run doc` in package.json. So the generated doc is using source from `/doc` folder.

legendecas:
IIUC, doxygen doesn't read these markdown files in `/doc`.

gengjiawen:
I see, we are going to remove doxygen part in current doc system ?

mhdawson:
@gengjiawen that was the plan, we have been remiss in not removing it.

PR #703, opened 2020-04-17
src: reset the NAPI_VERSION by fixing compiler error with NAPI_EXPERIMENTAL

yorkie:
This fixes a compiler error like this:

```shell
/Users/yorkie/workspace/pipcook/packages/boa/node_modules/node-addon-api/napi-inl.h:405:24: error: use of undeclared identifier 'napi_is_date'; did you mean 'napi_is_dataview'?
  napi_status status = napi_is_date(_env, _value, &result);
                       ^~~~~~~~~~~~
                       napi_is_dataview
```

This is because on 10.15.x, the Node.js will automatically define NAPI_VERSION to be a big value when NAPI_EXPERIMENTAL is defined, so it makes a mistake on resolving the `NAPI_VERSION > 4`.

The following commit introduces a pre-define which sets NAPI_VERSION to a magic number when NAPI_EXPERIMENTAL, this may cause an error when enabling NAPI_EXPERIMENTAL on an
actual napi3,4.

Node.js Commit: https://github.com/nodejs/node/commit/8476053c132fd9613aab547aba165190f8064254.

Now Node.js has removed that at the later commit, too. So let's #undef it.

### DIFF
--- a/napi.h
+++ b/napi.h
@@ -9,6 +9,12 @@
 #include <string>
 #include <vector>
 
+// Reset the NAPI_VERSION which is set by some versions of node_api.h
+// See: https://github.com/nodejs/node/commit/8476053c132fd9613aab547aba165190f8064254
+#if defined(NAPI_EXPERIMENTAL) && NAPI_VERSION == 2147483647
+#undef NAPI_VERSION
+#endif
+
 // VS2015 RTM has bugs with constexpr, so require min of VS2015 Update 3 (known good version)
 #if !defined(_MSC_VER) || _MSC_FULL_VER >= 190024210
 #define NAPI_HAS_CONSTEXPR 1


legendecas:
After js_native_api.h was introduced to split JS related API with Node.js specific API into two header files, NAPI_EXPERIMENTAL macro was moved to js_native_api.h. See https://github.com/nodejs/node/blob/master/src/js_native_api.h#L9 .

yorkie:
So is that required to add "NAPI_VERSION < 2147483647" for all "NAPI_VERSION > N" cases?

mhdawson:
@yorkie the desired behaviour is that if you set NAPI_VERSION you will be able to use any N-API functions which are included in that version or later.  If you define NAPI_EXPERIMENTAL you should be able to use all functions. The follow on for node-addon-api is that it should only expose methods that use N-API functions which are available in the NAPI_VERSION defined, or in the case of NAPI_EXPERIMENTAL all methods.

yorkie:
See https://github.com/alibaba/pipcook/pull/103/files, it fixes a compiler error by adding the NAPI_VERSION=6, if the current behavior is correct, we are able to use all functions if NAPI_EXPERIMENTAL without NAPI_VERSION.

mhdawson:
@yorkie is this still required?

yorkie:
Actually, it's not always required, and we could close this.

PR #714, opened 2020-05-01
src: Fix crashing issue in DefineClass

bghgary:
When using DefineClass, if an instance of the class is created
but the resulting class constructor itself is not held by a reference,
then this can result in a crash. The attached instance data is
currently attached to the constructor. The constructor, if not held,
can be garbage collected which will trigger the finalizers. The
attached instance data will be destroyed while the instance of
the class still needs it, resulting in a crash.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3228,9 +3228,15 @@ ObjectWrap<T>::DefineClass(Napi::Env env,
                              &value);
   NAPI_THROW_IF_FAILED(env, status, Function());
 
+  napi_value proto;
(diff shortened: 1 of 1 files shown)


legendecas:
Nice catch! Just a quick question (not investigated): how can constructor get gc-ed while there are active alive instances? We can get constructor function by `instance.constructor` 🤔.

bghgary:
Hmm, fair point. Maybe I have an issue in my implementation. I'm implementing a JSC version.

bghgary:
Yup. That was it. I didn't add the constructor property to the prototype.

PR #718, opened 2020-05-05
WIP: add safer unwrap

DaAitch:
Hello,

this PR prevents that `ObjectWrap::Unwrap` can be used with any type. It has to be a type having `ObjectWrap<T>` as sub class.

I spend some time finding a bug in my appliction, which did not segfault, because I did something like this [test/objectwrap-saferunwrap.cc](https://github.com/nodejs/node-addon-api/compare/master...DaAitch:feature/safer-objectwrap-unwrap?expand=1#diff-a71d3d02039e2af6cf01c920dbf97ae4) and it compiles and the reason I found out something is really going wrong is, that the stacktrace shows calling a method `->A()` actually called `->B()`. Then I realized, that it might be is a vtable pointer fixup problem and understand that of course it's possible, but not allowed to reinterpret cast `void *` to any type but the the original one.

So to prevent that s.o. tries to seductively unwrap an `ObjectWrap<Car>` or `ObjectWrap<Truck>` to any sub class like `IVehicle` (which obviously is not a ObjectWrap) it should not compile.

What do you think?

If this has a chance to land, we my need "should not compile" tests.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3167,12 +3167,29 @@ inline ObjectWrap<T>::~ObjectWrap() {
   }
 }
 
+namespace details {
+  // This type resolution fails, if the T is not convertible to ObjectWrap<T>
+  template<class T>
+  using ObjectWrapSaferType = typename std::enable_if<
+    std::is_convertible<T*, ObjectWrap<T>*>::value, T>::type;
+
+  template<class T>
+#ifndef NAPI_UNSAFER_UNWRAP
+  ObjectWrapSaferType<T>*
+#else
+  T*
+#endif
+  SaferUnwrap(Object wrapper) {
+    T* unwrapped;
+    napi_status status = napi_unwrap(wrapper.Env(), wrapper, reinterpret_cast<void**>(&unwrapped));
+    NAPI_THROW_IF_FAILED(wrapper.Env(), status, nullptr);
+    return unwrapped;
+  }
+}
+
 template<typename T>
 inline T* ObjectWrap<T>::Unwrap(Object wrapper) {
-  T* unwrapped;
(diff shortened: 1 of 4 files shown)


mhdawson:
I like the concept, do you have ideas of how we'd add the "should not compile" tests?

gabrielschulhof:
We should call the function `SafeUnwrap` not `SaferUnwrap` and the C preprocessor variable should be `NAPI_SAFE_UNWRAP`. If present, it should do the safe `napi_unwrap`, otherwise it should do the regular `napi_unwrap`. The way it is now is kind of confusing to me.

@DaAitch is this related to https://github.com/nodejs/node-addon-api/pull/732 (multiple inheritance)?

@mhdawson maybe we should have a second target in the test/binding.gyp file which only gets run in the CI and is considered failed if it successfully compile. Might be kind of a crude check though, because it might be quite an effort to start analyzing the compiler's stderr output to see if it generated the right error message. Our many supported platforms each have a compiler that produces its own individual message, and the message may change with the version of the compiler, so if we analyzed compiler stderr output we might be signing up for quite a maintenance burden.

The PR needs a rebase.

mhdawson:
@gabrielschulhof having that second target which fails makes sense to me :)

mhdawson:
@DaAitch any chance you can rebase so we can move forward.

DaAitch:
@mhdawson sorry for the delay, I didn't read notifications for a long time :(. If you still think this may help, please feel free to copy and adapt my code and close this PR. At present I think I'm too far away from this topic. Sorry for that

mhdawson:
I think at this point based on the last comment we'll close, and somebody can reopen in the future if they'd like to get it landed.

Issue #722, opened 2020-05-10
SIGABRT under debug build (node_g) only


davedoesdev:
Cause is "Cannot create a handle without a HandleScope" because `~ObjectWrap()` calls `Value()` without a taking a `HandleScope`.

Stack trace:

```
Thread 9 "node_g" received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffff4ddd700 (LWP 24758)]
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) where
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7a5b899 in __GI_abort () at abort.c:79
#2  0x00005555561ebf12 in node::Abort () at ../src/node_errors.cc:239
#3  0x00005555561ece0e in node::OnFatalError (
    location=0x5555584d00d8 "v8::HandleScope::CreateHandle()", 
    message=0x5555584d00a8 "Cannot create a handle without a HandleScope")
    at ../src/node_errors.cc:426
#4  0x000055555648529f in v8::Utils::ReportApiFailure (
    location=location@entry=0x5555584d00d8 "v8::HandleScope::CreateHandle()", 
    message=message@entry=0x5555584d00a8 "Cannot create a handle without a HandleScope") at ../deps/v8/src/api/api.cc:478
#5  0x000055555676dc7e in v8::Utils::ApiCheck (
    message=0x5555584d00a8 "Cannot create a handle without a HandleScope", 
    location=0x5555584d00d8 "v8::HandleScope::CreateHandle()", 
    condition=<optimised out>) at ../deps/v8/src/api/api.h:141
#6  v8::internal::HandleScope::Extend (isolate=isolate@entry=0x7fffdc000cf0)
    at ../deps/v8/src/handles/handles.cc:64
#7  0x000055555645e038 in v8::internal::HandleScope::CreateHandle (
    isolate=0x7fffdc000cf0, value=292238566193)
    at ../deps/v8/src/handles/handles-inl.h:156
#8  0x000055555617eb3f in v8::Local<v8::Value>::New (isolate=0x7fffdc000cf0, 
    that=0x7fffdc0c44d0) at ../deps/v8/include/v8.h:10627
[...]

PR #729, opened 2020-05-14
Fix ObjectWrap destructor from env exit under node_g

davedoesdev:
Only fails under debug buildtype (node_g)

Update: Also fails using release 12.16.3

#722

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -55,6 +55,7 @@ let testModules = [
   'objectwrap_constructor_exception',
   'objectwrap-removewrap',
   'objectwrap_multiple_inheritance',
+  'objectwrap_worker_thread',
   'objectreference',
   'reference',
   'version_management'
--- a/test/index.js
+++ b/test/index.js
@@ -91,6 +91,10 @@ if (napiVersion < 6) {
   testModules.splice(testModules.indexOf('typedarray-bigint'), 1);
 }
 
+if (majorNodeVersion < 12) {
+  testModules.splice(testModules.indexOf('objectwrap_worker_thread'), 1);
+}
+
 if (typeof global.gc === 'function') {
   (async function() {
   console.log(`Testing with N-API Version '${napiVersion}'.`);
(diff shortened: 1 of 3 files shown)


mhdawson:
@davedoesdev thanks for submitting a PR for this.

Does the stack trace suggest is occurring from a path triggered by the gc. I've not had time yet but wanted to think through if it was a concern or not if that was occurring (also away for next few days so will probably be mid next week before I find time).

davedoesdev:
@mhdawson it looks like it's when the environment is closing due to the worker thread exiting.
The handles are being cleaned up at that point.

davedoesdev:
@mhdawson the test also fails under Node 12.16.3 release build!

mhdawson:
Spent some time looking at this see https://github.com/nodejs/node/pull/33508

mhdawson:
@davedoesdev https://github.com/nodejs/node/pull/33508 just landed which should fix the reported problem (it will need to land in 14.x and 12.x as well).

I still think adding the test on the node-addon-api side makes sense in addition to the one in core.  Can you remove the non-test change and then we'll get it landed once the core backports are complete.

davedoesdev:
@mhdawson done

mhdawson:
@davedoesdev thanks, may be a while before we can land as we have to wait for Node.js backports but thanks for all of your working finding and helping to resolve. Much appreciated.

davedoesdev:
@mhdawson no worries. Any rough ETA?

mhdawson:
@davedoesdev sorry for the late update. Looks like its been backported as far back as 12.x.  I can't remember if we thought it would apply to 10.x but if you update the PR to just add the test we can run the CI and confirm one way or the other, and if not land it.

davedoesdev:
@mhdawson PR should already have been updated to contain just the test

mhdawson:
Added commit to only run on Node.js 12 and higher as it requires worker threads

mhdawson:
Landed as: 518cfdcdc10301723b9d48848e8297f43dc54209

mhdawson:
Thanks for all your work on this @davedoesdev

davedoesdev:
@mhdawson thanks for merging

PR #732, opened 2020-05-16
Fix wrap/unwrap of objects also inheriting from non-ObjectWrap

davedoesdev:
`ObjectWrap` methods need to cast to `T*` otherwise `this` gets set wrongly when unwrapping.

Included test fails without the fix because the object is unwrapped to the wrong address.

Tested using gcc 9.3.0


mhdawson:
@nodejs/n-api would be great to get one more review before I land this.

mhdawson:
Landed as ba7ad37d4462c5f3b61a8a7750df2906b7914609

PR #736, opened 2020-05-24
fix testEnumerables on ObjectWrap

fholzer:
The assertion on key length is broken. It's an assignment instead of a bool expression.
Also changes the tests for properties to the same style as the tests for "own properties."

### DIFF
--- a/test/objectwrap.js
+++ b/test/objectwrap.js
@@ -108,11 +108,11 @@ const test = (binding) => {
         keys.push(key);
       }
 
-      assert(keys.length = 4);
-      assert(obj.testGetSet);
-      assert(obj.testGetter);
-      assert(obj.testValue);
-      assert(obj.testMethod);
+      assert(keys.length == 6);
+      assert(keys.includes("testGetSet"));
+      assert(keys.includes("testGetter"));
+      assert(keys.includes("testValue"));
+      assert(keys.includes("testMethod"));
     }
   };
 
--- a/test/objectwrap.js
+++ b/test/objectwrap.js
@@ -109,10 +109,14 @@ const test = (binding) => {
       }
 
       assert(keys.length == 6);
+      // on prototype
       assert(keys.includes("testGetSet"));
       assert(keys.includes("testGetter"));
       assert(keys.includes("testValue"));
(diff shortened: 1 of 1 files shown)


fholzer:
Also since the key length assertion was broken, no one noticed that the key length actually changed because addition instance properties were added.  (in e8935bd8d97ba61cd4c005fab571d9fb39ab9249)

test/objectwrap.js
```diff
@@ -108,11 +108,11 @@ const test = (binding) => {
...
-      assert(keys.length = 4);
-      assert(obj.testGetSet);
-      assert(obj.testGetter);
-      assert(obj.testValue);
-      assert(obj.testMethod);
+      assert(keys.length == 6);
+      assert(keys.includes("testGetSet"));
+      assert(keys.includes("testGetter"));
+      assert(keys.includes("testValue"));
+      assert(keys.includes("testMethod"));
```
mhdawson, line 115: Since there are now 6, there should likely be an assert for the 2 additional methods right?

fholzer: From what I see, tests for those were added in the same commit the properties were added. See https://github.com/nodejs/node-addon-api/commit/e8935bd8d97ba61cd4c005fab571d9fb39ab9249#diff-ec20a47eb5374e448ec0794a5fb6ac6eR100
Though maybe I'm missing something...

mhdawson: I just though that if it validates that the number is 6, then it would be followed by 6 asserts as to what those are.  It's not clear to without taking more time to check that that the earlier 

```
 assert(Object.keys(obj).length === 2);
    assert(Object.keys(obj).includes('ownProperty'));
    assert(Object.keys(obj).indexOf('ownPropertyT') >= 0);
```

covers that and if it it does it may may sense to still have the additional checks so that if the length is 6 we follow it with 6 checks.

fholzer: I added the check in 2405844c2bae0482d13494d42c1aecb1f3290742

mhdawson:
@fholzer thanks!. Landed as d463f02bc78bd90d4149a1888e19e9a8159b44d7

PR #768, opened 2020-07-15
Windows CI: synchronise Node.js versions with those used for Linux/macOS CI

lovell:
Also updates to VS2017 so tests will compile - see #766 for further context.

Example successful build - https://ci.appveyor.com/project/lovell/node-addon-api/builds/34119509

Appveyor doesn't appear to currently run for PRs etc, so someone with the relevant access level to this repo may need to switch it back on again.


mhdawson:
I'm happy to land this update, but I don't think we've use it for the team's testing. It may have been used externally by the Microsoft people who were contributing to N-API a while back.

Issue #803, opened 2020-09-01
Add facilities to ease usage of async native patterns like boost::asio


Superlokkus:
I have been using 2 patterns of code to use native code in a promised way with completion handlers from native code, which iof course usually have their own thread regime and don't want to use worker threads from node. Specifically I am using `boost::asio` oriented code with own managed threads.

See https://github.com/Superlokkus/node-addon-api/blob/master/doc/generic_callback_wrapper.md for documentation and example

I have put my first pattern for simple "Get value from native function which calls a completion handler to a promise" into a class `GenericCallbackWrapper` and added unit tests. See my PR in a couple of seconds, I guess you want some changes or discussion since for now I hoped to be able to use Node >=10, but I can change that.

Superlokkus:
BTW here are the wrappers you could also use stand alone without PR:

```
#include <napi.h>
#include <functional>
#include <future>
#include <memory>
#include <utility>

#include <stdexcept>

template<typename T>
class generic_callback_wrapper {
public:
    using result_t [[maybe_unused]] = T;
    using callback_t = std::function<void(std::future<T>)>;
    using conversion_function_t = std::function<Napi::Value(const Napi::Env &, std::future<T> &&)>;

    generic_callback_wrapper(const Napi::Promise::Deferred &deferred, conversion_function_t conversion_function)
            : internal_(std::make_shared<internal>(deferred, conversion_function)) {
        internal_->set_up();
    }

    /*!
     * @brief The native callback shall only be called once per wrapper
     * @return
     */
    callback_t get_native_callback() {
        return internal_->get_native_callback();
    }

private:
    struct internal : std::enable_shared_from_this<internal> {
        Napi::Promise::Deferred deferred_;
        Napi::ThreadSafeFunction function_;
        std::future<T> result_;
        conversion_function_t conversion_function_;

        internal(const Napi::Promise::Deferred &deferred,
                 conversion_function_t conversion_function) : deferred_(deferred),
                                                              conversion_function_(conversion_function) {}

        void set_up() {
            auto resolver = Napi::Function::New(deferred_.Env(), [me = this->shared_from_this()]
                    (const auto &info) -> Napi::Value {
                if (info.Env() != me->deferred_.Env())
                    throw std::logic_error("generic_callback_wrapper: Napi Environments not equal");
                me->set_deferred();
                return info.Env().Undefined();
            });
[...]

PR #804, opened 2020-09-01
Add Wrappers for generic RAII async usage

Superlokkus:
This is related to #803 : It adds 2 classes, `GenericCallbackWrapper` and `GenericSubscriptionWrapper`
which makes usage of native async code (e.g. boost::asio, drivers) from javascript as easy as this:

```
void start_task(value_t input, std::function<void(std::future<value_t>)> completion_handler){
/* will initiate a call of completion_handler from an arbitrary thread with no need for 
synchronization or further lifetime management but with exception forwarding */
}

Napi::Value JSFrontingNativeFunction(const Napi::CallbackInfo& info){
    Napi::Promise::Deferred deferred = Napi::Promise::Deferred::New(info.Env());

    auto wrapper = Napi::GenericCallbackWrapper<value_t>
            (deferred,[](const auto &env, auto &&future) -> Napi::Value {
                return Napi::Number::New(env, future.get());
            });

    start_task(info[0].As<Napi::Number>().Uint32Value(), wrapper.get_native_callback());
    return deferred.Promise();
}
}
```
All done:
- [X] Implementation
- [X] Documentation
- [X] Tests

Assumed Node >= 10 (because of assert) is up to discussion of course

For email if needed please use "markus@markusklemm.net" not the commit one.


Superlokkus:
Sorry for not using the draft mode btw, this PR is considered to be finished

mhdawson:
@Superlokkus since the team has not been able to get to this yet would you be interested in coming to the next N-API team meeting and giving the intro to the motivation/implementation.  That might helps us move it forward.

Superlokkus:
> @Superlokkus since the team has not been able to get to this yet would you be interested in coming to the next N-API team meeting and giving the intro to the motivation/implementation. That might helps us move it forward.

Yes, I assume it's a remote meeting? Shall I prepare something?

mhdawson:
@Superlokkus, remote yes, the meeting is open to join. There are no invites but the meetings are in the project calendar here: nodejs.org/calendar.

They are currently Fridays at 11 Eastern.  The zoom is in the calendar entry, but for convenience it is: 
https://zoom.us/j/363665824

It does not have to be anything too formal, you can just come and talk to us about it, but if you want to prepare slides or something you can do that as well.  Looking forward to talking to you.

gabrielschulhof (commented):
This is a very convenient way of using `ThreadSafeFunction` 👍 Please move the implementation code to `napi-inl.` and leave only the prototypes in `napi.h`!

mhdawson:
From @Superlokkus video which talks more about this: https://www.youtube.com/watch?v=jpH5-DXDovk

test/generic_subscription_wrapper.cc
```diff
@@ -0,0 +1,65 @@
...
+
+#include <thread>
+#include <stdexcept>
+#include <cstdint>
+#include <future>
+#include <utility>
+#include <functional>
+#include <atomic>
+
+#if (NAPI_VERSION > 3)
```
mhdawson, line 11: I think this needs to be >4 as the threadsafe functions were introduced in N-API version 4 -> https://nodejs.org/api/n-api.html#n_api_napi_threadsafe_function

test/generic_callback_wrapper.cc
```diff
@@ -0,0 +1,89 @@
+#include "napi.h"
+
+#include <thread>
+#include <stdexcept>
+#include <cstdint>
+#include <future>
+#include <utility>
+#include <functional>
+
+#if (NAPI_VERSION > 3)
```
mhdawson, line 10: I think this needs to be >4 as the threadsafe functions were introduced in N-API version 4 -> https://nodejs.org/api/n-api.html#n_api_napi_threadsafe_function

test/generic_subscription_wrapper.cc
gabrielschulhof: `>= 4` would probably be clearer, because it mentions the N-API version needed for the tsfn while at the same time indicating with which minimum version of N-API it works.

test/generic_callback_wrapper.cc
gabrielschulhof: `>= 4` would probably be clearer, because it mentions the N-API version needed for the tsfn while at the same time indicating with which minimum version of N-API it works.

mhdawson:
Run on v10.x to make sure we are not using any C++ features not supported in the compilers used to build earlier releases: https://ci.nodejs.org/job/node-test-node-addon-api-new/3131/

test/generic_subscription_wrapper.cc
mhdawson: I agreed >=4 would be better. I just realised that it was ok because it was a > and was coming back to leave that comment, but @gabrielschulhof beat me to it:)

test/generic_callback_wrapper.cc
mhdawson: I agreed >=4 would be better. I just realised that it was ok because it was a > and was coming back to leave that comment, but @gabrielschulhof beat me to it:)

mhdawson:
Seems to fail on one of the windows variants: https://ci.nodejs.org/job/node-test-node-addon-api-new/3132/nodes=win-vs2017/console

doc/generic_callback_wrapper.md
```diff
@@ -0,0 +1,77 @@
+# GenericCallbackWrapper
+
+`Napi::GenericCallbackWrapper<T>` is a class with one template parameter `T`, 
+which is intended to allow you to return `T` from a C++ function running arbitrary async i.e. concurrency 
+regimes like `boost::asio` via a `Napi::Promise` i.e. JS promise, including exception and conversion function handling. 
+
+`Napi::GenericCallbackWrapper<T>` requires C++ exceptions enabled, so works by default.
```
mhdawson, line 7: I think this should be expanded, with a reference to https://github.com/nodejs/node-addon-api/blob/fcf173d2a177cca2cc5a722117fe374513d7d1fb/doc/error_handling.md so that its clear this feature is only supported when you use the "Handling Errors With C++ Exceptions" approach.

doc/generic_subscription_wrapper.md
```diff
@@ -0,0 +1,48 @@
...
+which is intended to allow you to call a JavaScript callback with `T` as parameter repeatedly  
+from a C++ function (subscribe) running arbitrary async i.e. concurrency/timer regimes like `boost::asio` via a `Napi::Promise`
+until a callback in JavaScript is called (unsubscribe).
+
+Overall this Object is quite similar to [GenericCallbackWrapper](generic_callback_wrapper.md) but intended to be used
+for efficient repeatedly data push pattern, for example pushing device states triggered by arbitrary timers to maybe
+multiple javascript callbacks. To give satisfying freedom for such house keeping, the unsubscribe function 
+is also promised, for cases where the according native facilities has also to be looked up or messaged first.
+
+`Napi::GenericSubscriptionWrapper<T>` requires C++ exceptions enabled, so works by default.
```
mhdawson, line 13: Same comment for other class

napi-inl.h
```diff
@@ -5078,6 +5078,170 @@ Addon<T>::DefineProperties(Object object,
 }
 #endif  // NAPI_VERSION > 5
 
+#if NAPI_VERSION > 4
+#ifdef NAPI_CPP_EXCEPTIONS
+template<typename T>
+class GenericCallbackWrapper {
```
mhdawson, line 5084: All of napi-inl is inline and this does not look like it is.

doc/generic_callback_wrapper.md
```diff
@@ -0,0 +1,77 @@
...
+
+All you need is the `Napi::Promise::Deferred` you want to Resolve/Reject, a function to convert your type `T` 
+to a `Napi::Value` to construct your wrapper, which then can give you a `std::function` callback, which
+can be called later from any custom thread, in order to resolve/reject the promise. 
+Life time handling included. That means that the wrapper is internally reference counted and don't need to be stored, 
+the function object you get from `get_native_callback` will ensure the lifetime.
```
mhdawson, line 13: It would be useful to document the methods/parameters on GenericCallbackWrapper that can be called by code that uses the class. That would include the constructor, set_unsubscription_function, and get_native_callback.  That would be consistent with the docs for other parts of node-addon-api I think help explain how it is used.

Same comment for the GenericSubscriptionWrapper as well.

mhdawson:
We discussed again in todays meeting and the what we agreed:

* We don't want to pull to much into node-addon-api beyond being a wrapper around the C API, we already have a few different higher level wrappers (PRs like this one) for the Threadsafe function and don't think it would make sense to pull them all in.
* We talked about creating a new organization which groups extras for Node-api and in cases like this offer a repo in that org, with the expectation that the author would maintain that repo.
* The alternative is of course that authors just create their own repos and we just have a list in the docs which points to that.

Is one of those something that you'd be interested in?  @Superlokkus

Superlokkus:
> Is one of those something that you'd be interested in? @Superlokkus
I am sorry for my abstinence the last weeks, I am not completely sure, I will try to join one of upcoming meetings again.

aminya:
This PR looks very interesting. @Superlokkus Are you interested in making a separate repository for it? Maybe something like `napi-async`?

mhdawson:
I'm going to close this as there does has not been recent discussions and the paths forward currently don't include landing the PR.  Please let me know if you feel that was no the right thing to do.

Issue #806, opened 2020-09-03
test: threadsafe function test is not executed after last changes


NickNaso:
The test for the thread safe function is not executed anymore after the last changes.

PR #787, opened 2020-08-07
test: refactor threadsafe_function test with async/await

Velmisov:
#776

PR #807, opened 2020-09-03
test: fixed the execution for the threadsfafe function test.

NickNaso:
Fixes: https://github.com/nodejs/node-addon-api/issues/806

### DIFF
--- a/test/threadsafe_function/threadsafe_function.js
+++ b/test/threadsafe_function/threadsafe_function.js
@@ -4,10 +4,10 @@ const buildType = process.config.target_defaults.default_configuration;
 const assert = require('assert');
 const common = require('../common');
 
-module.exports = async function() {
+module.exports = (async function() {
   await test(require(`../build/${buildType}/binding.node`));
   await test(require(`../build/${buildType}/binding_noexcept.node`));
-};
+})();
 
 async function test(binding) {
   const expectedArray = (function(arrayLength) {


mhdawson:
Can you add a bit more detail on what you saw in terms of it not running?

When I run the tests before apply the patch I see:

```
Running test 'threadsafe_function/threadsafe_function_ctx'
Running test 'threadsafe_function/threadsafe_function_existing_tsfn'
Running test 'threadsafe_function/threadsafe_function_ptr'
Running test 'threadsafe_function/threadsafe_function_sum'
Running test 'threadsafe_function/threadsafe_function_unref'
Running test 'threadsafe_function/threadsafe_function'
```

NickNaso:
Hi @mhdawson,
what you see printed on the console is in https://github.com/nodejs/node-addon-api/blob/master/test/index.js#L106 but this does not mean that the corresponding test is executed. In the test for threadsafe function https://github.com/nodejs/node-addon-api/blob/master/test/threadsafe_function/threadsafe_function.js#L7 we are exporting only the function instead we should export a a function invocation. Trying to insert a `console.log('...')` like reported below:
```js
module.exports = async function() {
  console.log('Execution of test for threadsafe function');
  await test(require(`../build/${buildType}/binding.node`));
  await test(require(`../build/${buildType}/binding_noexcept.node`));
};
```
you could see that the test will not be executed. This happened in the PR https://github.com/nodejs/node-addon-api/pull/787.

mhdawson (approved):
LGTM

mhdawson:
Landed as a13b36c96ef2f60a5646676e2d6680301d9ff3e6

PR #820, opened 2020-10-08
Fixed example in addon.md.

nempoBu4:


doc/addon.md
```diff
@@ -41,9 +41,9 @@ class ExampleAddon : public Napi::Addon<ExampleAddon> {
 
       // We can also attach plain objects to `exports`, and instance methods as
       // properties of those sub-objects.
-      InstanceValue("subObject", DefineProperties(Napi::Object::New(), {
-        InstanceMethod("decrement", &ExampleAddon::Decrement
-      })), napi_enumerable)
+      InstanceValue("subObject", DefineProperties(Napi::Object::New(env), {
```
mhdawson, line 44: I understand adding the env to New() but not removing  `napi_enumerable` as the signature seems like it requires that: https://github.com/nodejs/node-addon-api/blob/ceb27d49498c46e09d057c87ff3bd22d728946bf/napi-inl.h#L3425

nempoBu4: According to [InstanceWrap doc](https://github.com/nodejs/node-addon-api/blob/master/doc/instance_wrap.md):
```c++
template <typename T>
static Napi::ClassPropertyDescriptor<T>
Napi::InstanceWrap<T>::InstanceValue(const char* utf8name,
                            Napi::Value value,
```
`napi_enumerable` can be set but it is not required. I have verified that this code works.

But if you think that `napi_enumerable` needs to be included into this example then it still needs to be fixed because of wrong position of brackets:
```c++
InstanceValue("subObject", DefineProperties(Napi::Object::New(env), {
    InstanceMethod("decrement", &ExampleAddon::Decrement) // <- missing bracket
})/*) <- wrong bracket */, napi_enumerable)
```

gabrielschulhof: Without `napi_enumerable` the items cannot be seen from JS without some effort. Please leave it in place! Everything else LGTM.

nempoBu4: @gabrielschulhof I thought `napi_enumerable` means this items can be seen through `for..in` iteration or `Object.keys` as stated in the MDN's [Enumerability and ownership of properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties#Detection_Table) article. I thought without `napi_enumerable` the property still can be accessed through dot (`obj.property`) or brackets (`obj['property']`) or destructuring assignment (`{ property } = obj`) and there is no harm to remove `napi_enumerable` because there is no `for..in` iteration in `addon.md`.

So, I was wrong. I put `napi_enumerable` back in place. Does it mean that `napi_enumerable` needs to be added to other items too?
```c++
    DefineAddon(exports, {
      InstanceMethod("increment", &ExampleAddon::Increment, napi_enumerable), // <- here

      // We can also attach plain objects to `exports`, and instance methods as
      // properties of those sub-objects.
      InstanceValue("subObject", DefineProperties(Napi::Object::New(env), {
        InstanceMethod("decrement", &ExampleAddon::Decrement, napi_enumerable) // <- and here
      }), napi_enumerable)
    });
```

mhdawson (approved):
LGTM

PR #828, opened 2020-10-26
doc: avoid directing users to HTTP

tniessen:
There is no reason to use plain HTTP, it only makes things easier to track/hijack for ISPs and other untrustworthy entities.


NickNaso (approved):
LGTM

mhdawson (approved):
LGTM

PR #819, opened 2020-10-07
build: add incremental clang-format checks

legendecas:
git-clang-format only check lines modified between two refs.

Fixes: https://github.com/nodejs/node-addon-api/issues/543

PR #742, opened 2020-06-08
tsfn: implement TypedThreadSafeFunction

KevinEady:
Implements a templated `TypedThreadSafeFunction<ContextType, DataType, Callback>` class. The class also adds static, compile-time support for the optional `Function` argument on N-API 5+ using `std::nullptr_t` (as well as completely missing `Function` argument) on N-API 5+.

TODO:
- [x] Documentation
- [x] (optional) see if the existing `ThreadSafeFunction` class can utilize this one, as done in previous POC https://github.com/nodejs/node-addon-api/pull/687#issuecomment-619555274
  - Completed in https://github.com/nodejs/node-addon-api/pull/742#issuecomment-675797922

Fixes:
- #665 
- #653 
- #594

PR #832, opened 2020-11-16
create a HandleScope in FinalizeCallback

blagoev:
Seems like FinalizeCallback needs to create a HandleScope since it calls ObjectWrap<T>::~ObjectWrap() which might need to create a temporary Object handle here https://github.com/nodejs/node-addon-api/blob/master/napi-inl.h#L3558

at Realm (https://github.com/realm/realm-js) we have crashes with stacktrace at this location.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3977,6 +3977,8 @@ inline napi_value ObjectWrap<T>::StaticSetterCallbackWrapper(
 
 template <typename T>
 inline void ObjectWrap<T>::FinalizeCallback(napi_env env, void* data, void* /*hint*/) {
+  HandleScope scope(env);
+
   T* instance = static_cast<T*>(data);
   instance->Finalize(Napi::Env(env));
   delete instance;
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3978,7 +3978,6 @@ inline napi_value ObjectWrap<T>::StaticSetterCallbackWrapper(
 template <typename T>
 inline void ObjectWrap<T>::FinalizeCallback(napi_env env, void* data, void* /*hint*/) {
   HandleScope scope(env);
-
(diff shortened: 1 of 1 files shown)


blagoev:
seems like clang-format fails on the whole `napi-inl.h` file and is not related to the changes in this PR

mhdawson:
@blagoev could you add  the stack trace to the issue? It would be good to understand the path that leads to the case where there is no HandleScope already on the stack.

gabrielschulhof:
@mhdawson I think this needs to be fixed in core. This fix is certainly not incorrect, but @KevinEady's backtraces in #742 indicate that there is no handle scope when the `napi_env` is iterating over outstanding references during `~napi_env`, because each reference's `Finalize()` is being called without `env->CallIntoModule()`. We can have both fixes for sure.

gabrielschulhof:
@mhdawson nevermind. This was fixed in core (https://github.com/nodejs/node/pull/34366) in RunAndClearNativeImmediates, which is part of @KevinEady's stack.

gabrielschulhof (approved):
We need this for older versions of Node.js 👍

blagoev:
yes I can confirm the crash was happening when closing the application which might be where `~napi_env` is called.

mhdawson (commented):
LGTM

NickNaso (approved):
LGTM

KevinEady (approved):
LGTM

mhdawson:
I think this is just missing https://github.com/nodejs/node-addon-api/pull/819 in terms of the lint checks. As far as I can see it passes in my environment. Going to land.

mhdawson (approved):

mhdawson:
Landed as 1427b3ef7853271bfdf9057c33ab9a03e7fc67b1

PR #840, opened 2020-11-23
test: fixup testing for specific N-API version

mhdawson:
Fixup testing for a specific N-API version
and document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,15 @@ npm install
 npm test --disable-deprecated
 ```
 
+To run the tests targetting a specific version of N-API run
+```
+npm install
+export NAPI_VERSION=X
+npm test --NAPI_VERSION=X
+```
+
+where X is the version of N-API you want to target.
+
 ### **Debug**
 
 To run the **node-addon-api** tests with `--debug` option:
--- a/test/index.js
+++ b/test/index.js
@@ -62,7 +62,15 @@ let testModules = [
   'version_management'
 ];
 
-const napiVersion = Number(process.versions.napi)
+let napiVersion = Number(process.versions.napi)
+if (process.env.NAPI_VERSION) {
+  // we need this so that we don't try run tests that rely
+  // on methods that are not available in the NAPI_VERSION
+  // specified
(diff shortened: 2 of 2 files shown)


mhdawson:
@KevinEady, @legendecas I don't think the linting is working correctly. It's reporting problems in files I've not changed.

NickNaso (approved):
LGTM

legendecas:
@mhdawson the check is comparing the local master with the current PR branch. So on the CI, it will compare with the latest master. We have to update the PR branch with the latest master branch to make sure the diff do not include changes landed on master after the fork of the development branch.

mhdawson:
@legendecas I'm not following what we need to fix this?

legendecas:
＠mhdawson Sorry but maybe I'm misunderstanding the problem. I didn't find the CI is failing. Can you help me explain what failure you're referring to?

mhdawson:
@legendecas it was reporting failures when I ran locally and I had to add the comment to ingore the pre-commit checks.

legendecas:
@mhdawson ok. The linter compares the current working branch with the local master branch. If the local master branch is not up to date or contains the latest commits that not linted, which commits the working branch doesn't include, then the diff will include those _unrelated_ changes and the linter will complain about them.

mhdawson:
@legendecas that may cause people confusion/trouble since I'm not sure how many people keep their local master branch up to date.

KevinEady:
@mhdawson FWIW, the script allows you to specify a process env [`CLANG_FORMAT_START`](https://github.com/nodejs/node-addon-api/blob/master/tools/clang-format.js#L7), eg `upstream/master` .. but yes, it definitely caused me confusion/trouble, which is how I stumbled across this setting 🤷

legendecas:
The local master branch might not always tracking nodejs/node-addon-api but rather someones-fork/node-addon-/api so I didn't comparing with local upstream/master or some-local-remote/master.

Maybe we can print instructions when the check failed how to fix the local check to improve the experience.

PR #843, opened 2020-11-24
doc: add support info

mhdawson:
Add support info as per:
https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md

### DIFF
--- a/package-support.json
+++ b/package-support.json
@@ -1,0 +1,21 @@
+{
+  "versions": [
+    {
+      "version": "*",
+      "target": {
+        "node": "lts"
+      },
+      "response": {
+        "type": "time-permitting",
+        "paid": false,
+        "contact": {
+          "name": "node-addon-api team",
+          "url": "https://github.com/nodejs/node-addon-api/issues"
+        }
+      },
+      "backing": [ { "project": "https://github.com/nodejs" },
+                   { "foundation": "https://openjsf.org/" }
+      ]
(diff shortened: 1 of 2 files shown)


NickNaso (approved):
LGTM

KevinEady (commented):
According to [Integration into package.json](https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md#integration-into-packagejson), it says:

> In all of these cases `package-support.json` (or your alternative filename) should be included in your `files` array, or `.npmignore` configured to avoid excluding it.

Should this be done on this PR as well...?

package-support.json
```diff
@@ -0,0 +1,21 @@
+{
+  "versions": [
+    {
+      "version": "*",
+      "target": {
+        "node": "lts"
```
KevinEady, line 6: According to [`node` namespace](https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md#node-namespace), should this be `active`? Since we support node 10,12,14,15 (whereas `lts` is only 10,12)

mhdawson: LTS is now 10,12,14, but regardless good point.

mhdawson: @KevinEady fixed.

legendecas (approved):

PR #846, opened 2020-11-30
tools: print more instructions on clang-format check failed

legendecas:
Refs: #https://github.com/nodejs/node-addon-api/pull/840#issuecomment-735895055


mhdawson (approved):
LGTM, this will help

Issue #821, opened 2020-10-12
AsyncProgressWorker::OnProgress is invoked without arg sometimes


zbjornson:
I've got a basic AsyncProgressWorker that's very similar to the [example](https://github.com/nodejs/node-addon-api/blob/master/doc/async_worker_variants.md#example). I'm invoking `progress.Send(thing, 1)` a few hundred times per second, and seeing `OnProgress` invoked *sometimes* with `(info=nullptr, count=0)`:

```gdb
Thread 1 "node" received signal SIGSEGV, Segmentation fault.
0x00007ffff523ae45 in UmapWorker::OnProgress (this=0x486e420, info=0x0, count=<optimized out>) at /home/zbjornson/node-cuml/node_modules/node-addon-api/napi-inl.h:4248
4248      return Napi::Env(_env);
(gdb) bt
#0  0x00007ffff523ae45 in UmapWorker::OnProgress (this=0x486e420, info=0x0, count=<optimized out>) at /home/zbjornson/node-cuml/node_modules/node-addon-api/napi-inl.h:4248
#1  0x00007ffff52354ef in Napi::AsyncProgressWorker<ProgressInfo>::OnWorkProgress (this=0x486e420) at /usr/include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:778
#2  Napi::AsyncProgressWorkerBase<void>::OnAsyncWorkProgress (data=<optimized out>) at /home/zbjornson/node-cuml/node_modules/node-addon-api/napi-inl.h:4732
#3  0x00007ffff5237083 in std::function<void (Napi::Env, Napi::Function)>::operator()(Napi::Env, Napi::Function) const (__args#1=..., __args#0=..., this=0x7fff466e62f0) at /usr/include/c++/8/bits/std_function.h:682
#4  Napi::ThreadSafeFunction::CallJS (env=0x47e10c0, jsCallback=0x45443b8, data=0x7fff466e62f0) at /home/zbjornson/node-cuml/node_modules/node-addon-api/napi-inl.h:4667
#5  0x0000000000a0d738 in v8impl::(anonymous namespace)::ThreadSafeFunction::IdleCb(uv_idle_s*) ()
#6  0x00000000013772e9 in uv__run_idle (loop=loop@entry=0x4444220 <default_loop_struct>) at ../deps/uv/src/unix/loop-watcher.c:68
[...]

PR #853, opened 2020-12-11
fix: empty data OnProgress by race conditions of AsyncProgressWorker

legendecas:
Fixes https://github.com/nodejs/node-addon-api/issues/821

The callback of ThreadSafeFunction is not been invoked immediately on the callback of uv_async_t (uv io poll), rather the callback of TSFN is invoked on the right next uv idle callback. There are chances that during the deferring the signal of uv_async_t is been sent again, i.e. potential not coalesced two calls of the TSFN callback.

However, when the race condition happens, the AsyncProgressWorker::OnProgress is likely not been called yet. There is hardly a fix on ensuring the race condition not happening. 

In summary, the fix is suppressing the unexpected duplicated calls of AsyncProgressWorker::OnProgress.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 
 jobs:
   test:
+    timeout-minutes: 30
     strategy:
       matrix:
         node-version:
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -5390,6 +5390,17 @@ inline void AsyncProgressWorker<T>::OnWorkProgress(void*) {
     this->_asyncsize = 0;
   }
 
+  /**
+   * The callback of ThreadSafeFunction is not been invoked immediately on the
+   * callback of uv_async_t (uv io poll), rather the callback of TSFN is
+   * invoked on the right next uv idle callback. There are chances that during
+   * the deferring the signal of uv_async_t is been sent again, i.e. potential
+   * not coalesced two calls of the TSFN callback.
+   */
+  if (data == nullptr) {
+    return;
+  }
+
   this->OnProgress(data, size);
   delete[] data;
 }
--- a/test/asyncprogressworker.cc
+++ b/test/asyncprogressworker.cc
@@ -84,7 +84,6 @@ class MalignWorker : public AsyncProgressWorker<ProgressData> {
       ProgressData data{0};
       progress.Send(&data, 1);
     }
-    _cv.wait(lock);
   }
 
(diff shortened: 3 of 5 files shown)


mhdawson (approved):
LGTM

gabrielschulhof (approved):

mhdawson:
@legendecas I tried to land this and had 2 issues

1) it reported an unused value warning on line 90 of test/asyncprogressworker.cc.  I update the line so not an issue:
-  void OnProgress(const ProgressData* data, size_t count) override {
+  void OnProgress(const ProgressData*, size_t count) override {

2) The tests seem to stop after the asyncprogressworker test:

```shell
napiVersion:6
napiVersion:6
Testing with N-API Version '6'.
Starting test suite

Running test 'addon_build'
   >Preparing native addons to build
   >Building addon: 'echo addon'
   >Running test for: 'echo addon'
   >Building addon: 'echo-addon'
   >Running test for: 'echo-addon'
Running test 'addon'
Running test 'addon_data'
Running test 'arraybuffer'
Running test 'asynccontext'
Running test 'asyncprogressqueueworker'
Running test 'asyncprogressworker'
```

We'll need to figure that out before this can land.

NickNaso (approved):
LGTM

legendecas:
> @mhdawson  The tests seem to stop after the asyncprogressworker test

Yeah, I've identified the problem. There might be an exception leaking in threadsafe function callbacks in node core so that the promise chains of the node-addon-api are broke up yet the thrown JavaScript exception is not handled properly and the process exits silently. I'll try to file a minimum reproduce to get that fixed in node core. Besides that, I'll update the PR to get tests running.

legendecas:
@mhdawson GitHub Actions are up and all tests get passed :) PTAL

legendecas:
> > @mhdawson  The tests seem to stop after the asyncprogressworker test
> 
> Yeah, I've identified the problem. There might be an exception leaking in threadsafe function callbacks in node core so that the promise chains of the node-addon-api are broke up yet the thrown JavaScript exception is not handled properly and the process exits silently. I'll try to file a minimum reproduce to get that fixed in node core. Besides that, I'll update the PR to get tests running.

Just filed WIP https://github.com/nodejs/node/pull/36510. Need to figure out if there are any other cases that also swallowing exceptions.

mhdawson:
@legendecas thanks for figuring out the issue and fixing up the tests. They all pass  for me, landing.

mhdawson:
Landed as 86feeebf5489126ab941afc92866eba0257ff8d2

PR #857, opened 2020-12-15
test: run tests in separated process

legendecas:
To identify corruptions between tests


legendecas:
Seems like duplicated with #738. Closing.

PR #868, opened 2020-12-28
Cannot Ref and throw exception in ObjectWrap constructor

markandrus:
I noticed that you cannot `Ref` in an ObjectWrap instance's constructor and then throw an exception. This produces a segfault, as demonstrated by running `npm test` on this branch:

```
Running test 'objectwrap_constructor_exception'
node(99930,0x111f5ce00) malloc: *** error for object 0x10abb1060: pointer being freed was not allocated
node(99930,0x111f5ce00) malloc: *** set a breakpoint in malloc_error_break to debug
```

If I add a call to `Unref` before throwing the exception, things work just fine; however, I think we should be able to handle this automatically in ObjectWrap's destructor.


markandrus:
I'm closing this in favor of #869, because I probably don't know how `Ref` and `Unref` are supposed to work&hellip;

PR #737, opened 2020-05-25
test: fix up delays for array buffer test

mhdawson:
Refs: https://github.com/nodejs/node-addon-api/issues/735

Signed-off-by: Michael Dawson <michael_dawson@ca.ibm.com>

Issue #735, opened 2020-05-22
Test failed: External ArrayBuffer with finalizer


mhdawson:
Test failed: External ArrayBuffer with finalizer seems to have been failing since May 16, on Windows for master and 14.x
```text
Running test 'objectreference'
Running test 'version_management'

All tests passed!
Test failed: External ArrayBuffer with finalizer
C:\workspace\node-test-node-addon-api-new\nodes\win-vs2019\node-addon-api\test\testUtil.js:17
        throw e;
        ^

AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

1 !== 0

    at Array.<anonymous> (C:\workspace\node-test-node-addon-api-new\nodes\win-vs2019\node-addon-api\test\arraybuffer.js:42:14)
    at runGCTests (C:\workspace\node-test-node-addon-api-new\nodes\win-vs2019\node-addon-api\test\testUtil.js:14:17)
    at Immediate._onImmediate (C:\workspace\node-test-node-addon-api-new\nodes\win-vs2019\node-addon-api\test\testUtil.js:21:9)
    at processImmediate (internal/timers.js:456:21) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: 1,
  expected: 0,
  operator: 'strictEqual'
}
npm ERR! Test failed.  See above for more details.
Build step 'Conditional steps (multiple)' marked build as failure
```

mhdawson:
Don't see any changes to the test files for arraybuffer.js or arraybuffer.cc recently.

@gabrielschulhof are you aware of any finalizer related changes on the node core side recently?

mhdawson:
This looks like it could be related: https://github.com/nodejs/node/pull/33321.  It also went in to 14.x as well so that would explain it failing on both master and 14 but not earlier releases.

mhdawson:
From this comment likely an issue with timing/windows https://github.com/nodejs/node/pull/33321#issuecomment-626253489

as it sounds like the test needed to be updated for core for Windows.

gabrielschulhof:
A fix for this has landed in #737.

PR #895, opened 2021-01-31
test: remove outdated V8 flag

RaisinTen:
The `--concurrent-array-buffer-freeing` flag is going to be removed
upstream in V8 9.0.

Refs: https://github.com/v8/v8/commit/0a480c30d553c1b8dddb0dddcbdec5fb6f3101ff

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -4,7 +4,7 @@ const majorNodeVersion = process.versions.node.split('.')[0];
 
 if (typeof global.gc !== 'function') {
   // Construct the correct (version-dependent) command-line args.
-  let args = ['--expose-gc', '--no-concurrent-array-buffer-freeing'];
+  let args = ['--expose-gc'];
   if (majorNodeVersion >= 14) {
     args.push('--no-concurrent-array-buffer-sweeping');
   }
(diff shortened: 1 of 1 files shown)


RaisinTen:
cc @mhdawson could you please take a look once you are free? :)

test/index.js
```diff
@@ -4,7 +4,7 @@ const majorNodeVersion = process.versions.node.split('.')[0];
 
 if (typeof global.gc !== 'function') {
   // Construct the correct (version-dependent) command-line args.
-  let args = ['--expose-gc', '--no-concurrent-array-buffer-freeing'];
+  let args = ['--expose-gc'];
```
mhdawson, line 7: Did you check the history of why this was added in the first place. I'm wondering if adding it needs to be Node.js version dependent versus just removing completely (ie might results in failures on older Node.js versions...)

RaisinTen: Sorry, I didn't check it before making this PR. I checked now and the PR that introduced this is: https://github.com/nodejs/node-addon-api/pull/737. It fixed this issue: https://github.com/nodejs/node-addon-api/issues/735. I'm not sure whether removing this flag would end up causing the failure again because we don't have any Windows tests in the CI.

RaisinTen: I added a check to conditionally add the option if `v8` has a major version that is less than `9`. PTAL.

mhdawson (approved):
LGTM, thanks for updating.

mhdawson:
Landed as f7ed2490d4c5572281528f6cc620786f261670a7 thanks @RaisinTen

Issue #891, opened 2021-01-29
Valgrind detected memory leak


lygstate:
```
==6515== 
==6515== 20 bytes in 1 blocks are definitely lost in loss record 1 of 7
==6515==    at 0x4834CAB: operator new(unsigned int) (vg_replace_malloc.c:328)
==6515==    by 0x257D1D: Napi::InstanceWrap<node_sqlite3::Database>::InstanceAccessor(char const*, Napi::Value (node_sqlite3::Database::*)(Napi::CallbackInfo const&), void (node_sqlite3::Database::*)(Napi::CallbackInfo const&, Napi::Value const&), napi_property_attributes, void*) (napi-inl.h:3417)
==6515==    by 0x25034D: node_sqlite3::Database::Init(Napi::Env, Napi::Object) (database.cc:22)
==6515==    by 0x1798B1: (anonymous namespace)::RegisterModule(Napi::Env, Napi::Object) (node_sqlite3.cc:21)
==6515==    by 0x17A9E8: Napi::RegisterModule(napi_env__*, napi_value__*, Napi::Object (*)(Napi::Env, Napi::Object))::{lambda()#1}::operator()() const (napi-inl.h:375)
==6515==    by 0x17C4D0: napi_value__* Napi::details::WrapCallback<Napi::RegisterModule(napi_env__*, napi_value__*, Napi::Object (*)(Napi::Env, Napi::Object))::{lambda()#1}>(Napi::RegisterModule(napi_env__*, napi_value__*, Napi::Object (*)(Napi::Env, Napi::Object))::{lambda()#1}) (napi-inl.h:73)
==6515==    by 0x17AA37: Napi::RegisterModule(napi_env__*, napi_value__*, Napi::Object (*)(Napi::Env, Napi::Object)) (napi-inl.h:374)
==6515==    by 0x17A90A: Init(napi_env__*, napi_value__*) (node_sqlite3.cc:118)
==6515==    by 0x1738C6: napi_module_init (node_api_module.c:48)
==6515==    by 0x173A05: napi_module_init_static (node_api_module.c:89)
==6515==    by 0x17A955: iotjs_sqlite3_napi_init (node_sqlite3.cc:121)
==6515==    by 0x1697AA: iotjs_module_get (iotjs_module.c:43)
==6515== 
==6515== 20 bytes in 1 blocks are definitely lost in loss record 2 of 7
[...]

PR #899, opened 2021-02-03
src: fix leak in ObjectWrap instance set/getters

KevinEady:
Fixes: #891

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3263,7 +3263,7 @@ inline void InstanceWrap<T>::AttachPropData(napi_env env,
                                        napi_value value,
                                        const napi_property_descriptor* prop) {
   napi_status status;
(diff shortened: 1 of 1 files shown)


mhdawson (approved):
LGTM, thanks for investigating and fixing

KevinEady:
Flakey tests cause the run failure maybe @mhdawson ?

mhdawson:
@KevinEady it is a bit worrying that we seem to get intermittent failures. We saw that for a while in the CI but then it went away without me being able to figure out the cause.  Might be worth running a few times locally (say 10 times) to make sure it does not recreate.

gabrielschulhof (approved):
Good catch!

mhdawson:
Landed as: a575a6ec60652689c21c744e8da2fb1ecae4350a

PR #909, opened 2021-02-15
feat: add IsNullOrUndefined

januwA:
Added [IsNullOrUndefined](https://v8docs.nodesource.com/node-15.0/dc/d0a/classv8_1_1_value.html#a21955eef281671d01db6b5c82f0ce5c7) detection js type function 😃

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -539,6 +539,10 @@ inline bool Value::IsNull() const {
   return Type() == napi_null;
 }
 
+inline bool Value::IsNullOrUndefined() const {
+  return Type() == napi_undefined || Type() == napi_null;
+}
+
 inline bool Value::IsBoolean() const {
   return Type() == napi_boolean;
 }
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -539,7 +539,7 @@ inline bool Value::IsNull() const {
   return Type() == napi_null;
 }
 
-inline bool Value::IsNullOrUndefined() const {
+inline bool Value::IsNullish() const {
   return Type() == napi_undefined || Type() == napi_null;
 }
 
--- a/napi.h
+++ b/napi.h
@@ -269,8 +269,9 @@ namespace Napi {
 
     napi_valuetype Type() const; ///< Gets the type of the value.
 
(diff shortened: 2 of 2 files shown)


napi.h
```diff
@@ -269,8 +269,9 @@ namespace Napi {
 
     napi_valuetype Type() const; ///< Gets the type of the value.
 
-    bool IsUndefined() const;   ///< Tests if a value is an undefined JavaScript value.
     bool IsNull() const;        ///< Tests if a value is a null JavaScript value.
+    bool IsUndefined() const;   ///< Tests if a value is an undefined JavaScript value.
+    bool IsNullOrUndefined() const; ///< Tests if a value is a null or undefined JavaScript value.
```
KevinEady, line 274: Maybe `IsNullish` would be more concise. With the definition of `??` from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) as (emphasis mine)

> The nullish coalescing operator (??) is a logical operator that returns its right-hand side operand when its left-hand side operand **is null or undefined**, and otherwise returns its left-hand side operand.

napi-inl.h
```diff
@@ -539,6 +539,10 @@ inline bool Value::IsNull() const {
   return Type() == napi_null;
 }
 
+inline bool Value::IsNullish() const {
+  return Type() == napi_undefined || Type() == napi_null;
```
KevinEady, line 543: The compiler may optimize this since it is `const`, but lets store this in a variable to ensure it isn't called twice.

KevinEady (changes requested):
Hi @januwA ,

Thanks for your patience. So take a look at the comments. Also, you will need to add documentation to [`doc/value.md`](https://github.com/nodejs/node-addon-api/blob/main/doc/value.md) and a tests to [`test/basic_types/value.js`](https://github.com/nodejs/node-addon-api/blob/main/test/basic_types/value.js). For the test, I think checking for `true` on `null` and `undefined`, and `false` on something else is a good enough test coverage.

mhdawson:
@januwA are you still planning to move this forward?

PR #914, opened 2021-02-23
tools: fix error detection

RaisinTen:
The `ChildProces` object `result` does not have an `error` property. We
should use `stderr` instead.


mhdawson (approved):
LGTM

Issue #301, opened 2018-07-15
Napi::Reference not copyable, but semantically could be copyable


DaAitch:
Hi,

I don't understand why `Napi::Reference` can't be copied, although for `Napi::Error` it is allowed, implemented as protected. From technical point of view, the C-API allows create multiple reference holders of one value. When using lambdas captures we need copyable types. 

Look at the following example. A 3rd party lib allocates some memory we want to store in an `Napi::External` with a finalizer. We need to reference handle to overlive lambda where it is freed and after handle is unreferenced.

```cpp
// variable
Napi::Reference<Napi::External<H> > handle_; // H is handle type

Napi::Value MyClass::Alloc(const Napi::CallbackInfo& info) {
  Napi::EscapableHandleScope scope(info.Env());
  
  unsigned int size = info[0].ToNumber().Uint32Value();
  unsigned char* mem = lib_alloc(handle_.Value().Data(), size); // 3rd party alloc

  // lambda not copyable, because capture are not copyable

  // Either A)
  // create new Reference
  return scope.Escape(DMA<unsigned char>::constructor.New({
    // Napi::Persistent is noncopyable 
    Napi::External<unsigned char>::New(info.Env(), mem, [size, handle = Napi::Persistent(handle_.Value())](Napi::Env env, unsigned char* data) {
      lib_free(handle.Value().Data(), data, size); // 3rd party free
    })
  }));

  // Or B)
  // copy Reference
  return scope.Escape(DMA<unsigned char>::constructor.New({
    // Napi::Persistent is noncopyable 
    Napi::External<unsigned char>::New(info.Env(), mem, [size, handle = handle_](Napi::Env env, unsigned char* data) {
      lib_free(handle.Value().Data(), data, size); // 3rd party free
    })
  }));
}
```

I think we should make `Napi::Reference` copyable.
What do you think?

Best Regards
Philipp

mhdawson:
@nodejs/abi-stable-node please comment.

jasongin:
`Napi::Reference` is equivalent to `v8::Persistent` (which may be "weak" or not). By default `v8::Persistent` is not copyable, though that restriction may be changed using a traits parameter. See `CopyablePersistentTraits` in v8.

[...]

PR #915, opened 2021-02-26
src: allow non-copyable callbacks be finalizer parameters

legendecas:
Fixes: https://github.com/nodejs/node-addon-api/issues/301

In real-world use cases, `Napi::Reference`s are frequently been passed around by lambda captures in finalizers, callbacks. This PR aims to make those conditions easier to handle.

### DIFF
--- a/test/movable_callbacks.cc
+++ b/test/movable_callbacks.cc
@@ -1,0 +1,21 @@
+#include "napi.h"
+
+using namespace Napi;
+
+Value createExternal(const CallbackInfo& info) {
+  FunctionReference ref = Reference<Function>::New(info[0].As<Function>(), 1);
+  auto ret = External<char>::New(
+      info.Env(), nullptr, [ref = std::move(ref)](Napi::Env env, char* data) {
+        ref.Call({});
+      });
+
+  return ret;
+}
+
+Object InitMovableCallbacks(Env env) {
+  Object exports = Object::New(env);
+
+  exports["createExternal"] = Function::New(env, createExternal);
+
+  return exports;
+}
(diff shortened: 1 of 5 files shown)


mhdawson (approved):
LGTM

test/movable_callbacks.cc
```diff
@@ -0,0 +1,21 @@
+#include "napi.h"
+
+using namespace Napi;
+
+Value createExternal(const CallbackInfo& info) {
+  FunctionReference ref = Reference<Function>::New(info[0].As<Function>(), 1);
+  auto ret = External<char>::New(
+      info.Env(), nullptr, [ref = std::move(ref)](Napi::Env env, char* data) {
```
mhdawson, line 8: This fails to compile for me with:

```
../movable_callbacks.cc:8:61: error: unused parameter ‘env’ [-Werror=unused-parameter]
       info.Env(), nullptr, [ref = std::move(ref)](Napi::Env env, char* data) {
                                                   ~~~~~~~~~~^~~
../movable_callbacks.cc:8:72: error: unused parameter ‘data’ [-Werror=unused-parameter]
       info.Env(), nullptr, [ref = std::move(ref)](Napi::Env env, char* data) {
                                                                  ~~~~~~^~~~
```

mhdawson, line 8:
```suggestion
      info.Env(), nullptr, [ref = std::move(ref)](Napi::Env /*env*/, char* /*data*/) {
```

mhdawson: updated to make comment out the unused parameters like we do in other places

mhdawson: Push a commit to fix

mhdawson (approved):
LGTM

mhdawson:
Fixed up linter complaints as part of landing.

Landed as 74ab50c7754a5557a5ce0843b965648c0850a130

PR #926, opened 2021-03-09
doc: warn about SuppressDestruct()

addaleax:
Not sure how this ever made its way into the API, but people should *never* use this.


mhdawson (approved):
These suggestions/warning make sense given there is now support for worker threads.

NickNaso (approved):

PR #927, opened 2021-03-09
src: return Maybe<> on pending exception when cpp exception disabled

legendecas:
Fixes: https://github.com/nodejs/node-addon-api/issues/690

This PR introduced a Maybe type to enforce value existence check when CPP exceptions disabled and there is a pending JavaScript exception so that the operation can not be performed.

The `Maybe` returning behavior requires a predefined macro `NODE_ADDON_API_ENABLE_MAYBE` to enable.

- [x] make exiting tests compatible with two flavors (with or without Maybes)

### DIFF
--- a/doc/maybe.md
+++ b/doc/maybe.md
@@ -1,10 +1,10 @@
 # Maybe (template)
 
-Class `Napi::Maybe<T>` represents an maybe empty value: every `Maybe` is either
-`Just` and contains a value, or `Nothing`, and does not. `Maybe` types are very
-common in node-addon-api code, as they represent the function may throw a
-JavaScript exception and cause the program unable to evaluation any JavaScript
-code until the exception is been handled.
+Class `Napi::Maybe<T>` represents a value that may be empty: every `Maybe` is
+either `Just` and contains a value, or `Nothing`, and does not. `Maybe` types
+are very common in node-addon-api code, as they represent that the function may
+throw a JavaScript exception and cause the program to be unable to evaluate any
+JavaScript code until the exception has been handled.
 
 Typically, the value wrapped in `Napi::Maybe<T>` is [`Napi::Value`] and its
 subclasses.
@@ -58,7 +58,7 @@ template <typename T>
 T Napi::Maybe::UnwrapOr(const T& default_value) const;
 ```
 
-Return the value of type T contained in the Maybe, or using a default
+Return the value of type T contained in the Maybe, or use a default
 value if this Maybe is nothing (empty).
 
 ### UnwrapTo
@@ -69,6 +69,6 @@ bool Napi::Maybe::UnwrapTo() const;
 ```
 
 Converts this Maybe to a value of type `T` in the `out`. If this Maybe is
-nothing (empty), `false` is returned and `out is left untouched.
+nothing (empty), `false` is returned and `out` is left untouched.
 
 [`Napi::Value`]: ./value.md
--- a/doc/maybe.md
+++ b/doc/maybe.md
@@ -1,0 +1,74 @@
+# Maybe (template)
+
+Class `Napi::Maybe<T>` represents an maybe empty value: every `Maybe` is either
+`Just` and contains a value, or `Nothing`, and does not. `Maybe` types are very
+common in node-addon-api code, as they represent the function may throw a
+JavaScript exception and cause the program unable to evaluation any JavaScript
+code until the exception is been handled.
+
+Typically, the value wrapped in `Napi::Maybe<T>` is [`Napi::Value`] and its
+subclasses.
+
+## Methods
+
+### IsNothing
+
+```cpp
+template <typename T>
+bool Napi::Maybe::IsNothing() const;
+```
+
+Returns if the `Maybe` is `Nothing` and does not contain a value.
+
+### IsJust
+
+```cpp
+template <typename T>
+bool Napi::Maybe::IsJust() const;
+```
+
+Returns if the `Maybe` is `Just` and contains a value.
+
+### Check
+
+```cpp
+template <typename T>
+void Napi::Maybe::Check() const;
+```
+
+Short-hand for `Maybe::Unwrap()`, which doesn't return a value. Could be used
+where the actual value of the Maybe is not needed like `Object::Set`.
+If this Maybe is nothing (empty), node-addon-api will crash the
+process.
+
+### Unwrap
+
+```cpp
+template <typename T>
+T Napi::Maybe::Unwrap() const;
+```
+
+Return the value of type `T` contained in the Maybe. If this Maybe is
+nothing (empty), node-addon-api will crash the process.
+
+### UnwrapOr
+
+```cpp
+template <typename T>
+T Napi::Maybe::UnwrapOr(const T& default_value) const;
+```
+
+Return the value of type T contained in the Maybe, or using a default
+value if this Maybe is nothing (empty).
+
+### UnwrapTo
+
+```cpp
+template <typename T>
+bool Napi::Maybe::UnwrapTo() const;
+```
+
+Converts this Maybe to a value of type `T` in the `out`. If this Maybe is
+nothing (empty), `false` is returned and `out is left untouched.
+
+[`Napi::Value`]: ./value.md
--- a/doc/maybe.md
+++ b/doc/maybe.md
@@ -18,7 +18,8 @@ template <typename T>
 bool Napi::Maybe::IsNothing() const;
 ```
 
-Returns if the `Maybe` is `Nothing` and does not contain a value.
+Returns `true` if the `Maybe` is `Nothing` and does not contain a value, and
+`false` otherwise.
 
 ### IsJust
 
@@ -27,7 +28,8 @@ template <typename T>
 bool Napi::Maybe::IsJust() const;
 ```
 
-Returns if the `Maybe` is `Just` and contains a value.
+Returns `true` if the `Maybe` is `Just` and contains a value, and `false`
+otherwise.
 
 ### Check
 
@@ -65,7 +67,7 @@ value if this Maybe is nothing (empty).
 
 ```cpp
 template <typename T>
-bool Napi::Maybe::UnwrapTo() const;
+bool Napi::Maybe::UnwrapTo(T* result) const;
 ```
 
 Converts this Maybe to a value of type `T` in the `out`. If this Maybe is
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1080,25 +1080,41 @@ inline MaybeOrValue<Symbol> Symbol::WellKnown(napi_env env,
 #endif
 }
 
-inline Symbol Symbol::For(napi_env env, const std::string& description) {
+inline MaybeOrValue<Symbol> Symbol::For(napi_env env,
+                                        const std::string& description) {
   napi_value descriptionValue = String::New(env, description);
   return Symbol::For(env, descriptionValue);
 }
 
-inline Symbol Symbol::For(napi_env env, const char* description) {
+inline MaybeOrValue<Symbol> Symbol::For(napi_env env, const char* description) {
   napi_value descriptionValue = String::New(env, description);
   return Symbol::For(env, descriptionValue);
 }
 
-inline Symbol Symbol::For(napi_env env, String description) {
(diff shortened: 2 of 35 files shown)


legendecas:
I tested around the codebase, finally identified a few types of napi_* usage pattern in c++:

1. Plain function call/constructor returning Napi::* types
2. function call with `operator()`
3. `obj.Get`ters and `obj.Set`ters method
4. dataview read/write data
5. `obj[]` property accessing, like `obj["foo"] = val` and `Napi::Value val = obj["foo"]`
    - and array element accessing, like `arr[idx] = val` and `Napi::Value val = arr[idx]`
6. type casting with operators like `operator int32_t()`

The first 4 cases can be safely migrated with Maybe box without suppressing exceptions. However, the last two are either not able to return boxed values or boxing may invalidate the convenience intent. Thus they are not able to be migrated to Maybe box.

napi-inl.h
```diff
@@ -731,55 +799,74 @@ inline Number::Number(napi_env env, napi_value value) : Value(env, value) {
 }
 
 inline Number::operator int32_t() const {
+#ifdef NODE_ADDON_API_ENABLE_MAYBE
+  return Int32Value().FromMaybe(0);
```
legendecas, line 803: Like said, as the function is designed as a convenience method, a default value is returned if the box is empty.

legendecas:
@nodejs/node-api hey folks, I've marked most of the maybe throwing API with `MaybeOrValue<T>` return type and the `MaybeOrValue` is an alias for the type parameter `T` if the `Maybe` is not enabled. In that case, apis are not duplicated to implement `Maybe`s.

There are few exceptions for API return type changes. Following APIs currently return void on either success and failure (on exceptions disabled), thus they may fail silently if the property is not writable or not configurable:
- Object::Set
- Object::DefineProperties

If people are using the c node-api interface, they can use the returning `napi_status` to determine if the invocation is successful. However, it is not possible in node-addon-api with another node api call like `napi_get_and_clear_last_exception`/`Env::GetAndClearPendingException`. So I've updated them with returning `bool` to indicate if the invocation is succeeded. I'll split these changes into a separate PR so that this PR can focus on the `Maybe` introduction.

Anyway, apart from reusing the existing testing suites, the PR is ready for review. I'll continue to work on reusing most of the existing test suites so that we don't have to write two different test suites on whether or not `Maybe` is enabled.

legendecas:
@nodejs/node-api hi, team, I believe this PR is ready and may take some reviews. Thanks a lot!

doc/maybe.md
```diff
@@ -0,0 +1,74 @@
...
+Short-hand for `Maybe::Unwrap()`, which doesn't return a value. Could be used
+where the actual value of the Maybe is not needed like `Object::Set`.
+If this Maybe is nothing (empty), node-addon-api will crash the
+process.
+
+### Unwrap
+
+```cpp
+template <typename T>
+T Napi::Maybe::Unwrap() const;
```
legendecas, line 48: I've chosen the three value unwrapping method to be named in a more intuitive pattern like `Unwrap()`, `UnwrapOr(const T& default_value)`, and `UnwrapTo(T* out)`. This is somewhat not aligned with v8's `Maybe` interface, which has `FromJust`, `FromMaybe`, `To` respectively in names.

mhdawson:
A few questions that we might discuss more in the nest team meeting

* This seems to change whether some functions are const or not ?
* Changes Number::FloatValue - to add a round trip instead of cast ?
* There are still a number of TODOs 
  * //TODO: find a way to define NAPI_HAS_ATTRIBUTE_WARN_UNUSED_RESULT
  * Many of -> ///TODO(legendecas): find a way to boxing with MaybeOrValue
* We might need more explanation of what additional we need to do when writing tests to make sure the 
  Maybe case is properly covered.

legendecas:
@mhdawson thanks for reviewing!

- This seems to change whether some functions are const or not ?
These type of changes are not expected in this PR. I checked about this and didn't find any occurrence of changes. Please point them out if you find any of them in this PR.

- Changes Number::FloatValue - to add a round trip instead of cast ?
This change is not expected, I've removed them.

- //TODO: find a way to define NAPI_HAS_ATTRIBUTE_WARN_UNUSED_RESULT
As this is not used in code base anymore, I've removed them.

- Many of -> ///TODO(legendecas): find a way to boxing with MaybeOrValue
There are updates regarding the todo, removed.

- We might need more explanation of what additional we need to do when writing tests to make sure the
Maybe case is properly covered.
I've added a wrap-up document in `test/README.md` to help people understand what happens here.

mhdawson:
Maybe I mistaken last time, I looked again at changes to const definitions and I can't find nay now.

test/README.md
```diff
@@ -0,0 +1,102 @@
+# Writing Tests
+
+There are multiple flavors of node-addon-api test build that covers different
+build flags defined in `napi.h`:
+
+1. c++ exceptions enabled,
+2. c++ exceptions disabled,
+3. c++ exceptions disabled, and `NODE_ADDON_API_ENABLE_MAYBE` is defined.
+
+`Napi` functions that calling into JavaScript can have different declared
```
mhdawson, line 10:
```suggestion
`Napi` functions that call into JavaScript can have different declared
```

mhdawson, line 10: This sentence seems to end prematurely, and calling should be call

test/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+using namespace Napi;
+
+Value fn(const CallbackInfo& info) {
+  Object obj = info[0].As<Object>();
+  Value value = MaybeUnwrap(obj->Get("foobar")); // <- `obj->Get` may throws
+}
+```
+
+There are there test helper function to conveniently convert `Napi::MaybeOrValue<>`
```
mhdawson, line 33:
```suggestion
There are three test helper function to conveniently convert `Napi::MaybeOrValue<>`
```

test/README.md
```diff
@@ -0,0 +1,102 @@
...
+Converts `MaybeOrValue<T>` to `T` by checking that `MaybeOrValue` is NOT an
+empty `Maybe`.
+
+Returns the original value if `NODE_ADDON_API_ENABLE_MAYBE` is not defined.
+
+Example:
+
+```cpp
+Object obj = info[0].As<Object>();
+Value value = MaybeUnwrap(obj->Get("foobar")); // we are sure the parameters should not throw
```
mhdawson, line 52: This should probably show a return as I think the point is we should be using these for all returns?

test/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+```cpp
+#include "napi.h"
+#include "test_helper.h"
+
+using namespace Napi;
+
+Value fn(const CallbackInfo& info) {
+  Object obj = info[0].As<Object>();
+  Value value = MaybeUnwrap(obj->Get("foobar")); // <- `obj->Get` may throws
```
mhdawson, line 29: Should this be in a return or is it that we need to use the helpers for more than returns as well?

napi-inl.h
```diff
@@ -658,32 +724,32 @@ inline T Value::As() const {
...
+      _env, status, Napi::Boolean(_env, result), Napi::Boolean);
 }
 
-inline Number Value::ToNumber() const {
+inline MaybeOrValue<Number> Value::ToNumber() const {
   napi_value result;
   napi_status status = napi_coerce_to_number(_env, _value, &result);
-  NAPI_THROW_IF_FAILED(_env, status, Number());
-  return Number(_env, result);
+  NAPI_MAYBE_RETURN_OR_THROW_IF_FAILED(
```
mhdawson, line 737: I think maybe this should just be called:

NAPI_RETURN_OR_THROW_IF_FAILED

napi-inl.h
```diff
@@ -1148,208 +1236,195 @@ inline Object::PropertyLValue<uint32_t> Object::operator [](uint32_t index) {
...
     const std::vector<PropertyDescriptor>& properties) {
   napi_status status = napi_define_properties(_env, _value, properties.size(),
     reinterpret_cast<const napi_property_descriptor*>(properties.data()));
-  NAPI_THROW_IF_FAILED(_env, status, false);
-  return true;
+  NAPI_MAYBE_RETURN_OR_THROW_IF_FAILED(_env, status, status == napi_ok, bool);
 }
 
-inline bool Object::InstanceOf(const Function& constructor) const {
+inline MaybeOrValue<bool> Object::InstanceOf(const Function& constructor) {
```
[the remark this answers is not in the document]
mhdawson: @gabrielschulhof thanks for catching that, not sure why I did not find it on my second look.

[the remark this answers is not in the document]
legendecas: Thanks for catching it, fixed.

napi-inl.h
```diff
@@ -2679,102 +2770,96 @@ inline ObjectReference::ObjectReference(const ObjectReference& other)
...
-inline Napi::Value ObjectReference::Get(uint32_t index) const {
-  EscapableHandleScope scope(_env);
-  return scope.Escape(Value().Get(index));
+inline MaybeOrValue<Napi::Value> ObjectReference::Get(uint32_t index) const {
+  return Value().Get(index);
 }
 
-inline bool ObjectReference::Set(uint32_t index, napi_value value) {
-  HandleScope scope(_env);
+inline MaybeOrValue<bool> ObjectReference::Set(uint32_t index,
```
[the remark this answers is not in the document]
legendecas: The change is not necessary. Reverted.

napi-inl.h
legendecas: Sounds much simpler. Renamed.

mhdawson:
@legendecas looks like this needs to be rebased.

test/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+void fn(const CallbackInfo& info) {
+  Object obj = info[0].As<Object>();
+  Value value = MaybeUnwrap(obj->Get("foobar")); // <- `obj->Get` is calling
+  // into JavaScript and may throw JavaScript Exceptions. Here we just assert
+  // getting the parameter must not fail for convenience.
+
+  // ... do works with the value.
+}
+```
```
mhdawson, line 35: I think this might be better after the three examples or maybe it could just be removed if we had examples in each of the sections for the helpers?

test/functionreference.cc
```diff
@@ -8,15 +9,15 @@ Value Call(const CallbackInfo& info) {
   FunctionReference ref;
   ref.Reset(info[0].As<Function>());
 
-  return ref.Call({});
+  return MaybeUnwrapOr(ref.Call({}));
```
mhdawson, line 12: Same question in a lot of places. Is it ok to use the default object in MaybeUnwrapOr because these are tests and we are pretty sure we know there will be no exception?  ie it's ok for tests but real code should actually be providing a default or using MaybeUnwrap?

mhdawson (approved):
LGTM

legendecas:
@gabrielschulhof 👋  may I have your review on this again? thanks a lot!

test/objectwrap.cc
```diff
@@ -115,64 +120,120 @@ class Test : public Napi::ObjectWrap<Test> {
     Napi::Symbol kTestMethodTInternal     = Napi::Symbol::New(env, "kTestMethodTInternal");
     Napi::Symbol kTestVoidMethodTInternal = Napi::Symbol::New(env, "kTestVoidMethodTInternal");
 
-    exports.Set("Test", DefineClass(env, "Test", {
```
gabrielschulhof, line 118: This one and many lines below look like pure whitespace changes. Are they necessary?

legendecas: There are two new elements at the bottom of the block. git-clang-format will check with blocks as minimum units.

legendecas:
@gabrielschulhof hi, PR updated with suggestion applied. PTAL, thanks a lot!

mhdawson:
@gabrielschulhof I'll plan to land end of this week unless you have objections/additional comments before then.

legendecas:
@mhdawson @gabrielschulhof should we land this PR?

mhdawson:
@legendecas could you squash into 1 commit?  I tried landing and I ended up with conflicts.

legendecas:
@mhdawson used the GitHub interface to squash merging. It seems no conflicts at all..

PR #928, opened 2021-03-10
Function tests

JoseExposito:
This PR adds tests for every Function method.

 I splitted the changes in commits so the diff is easier to follow.

EDIT :thinking:
```
All tests passed!
Tests aborted with SIGSEGV
```

### DIFF
--- a/test/function.cc
+++ b/test/function.cc
@@ -130,6 +130,16 @@ Value CallConstructorWithVector(const CallbackInfo& info) {
    return func.New(args);
 }
 
+Value CallConstructorWithCStyleArray(const CallbackInfo& info) {
+  Function func = info[0].As<Function>();
+  std::vector<napi_value> args;
+  args.reserve(3);
+  args.push_back(info[1]);
+  args.push_back(info[2]);
+  args.push_back(info[3]);
+  return func.New(args.size(), args.data());
+}
+
 void IsConstructCall(const CallbackInfo& info) {
    Function callback = info[0].As<Function>();
    bool isConstructCall = info.IsConstructCall();
@@ -164,6 +174,8 @@ Object InitFunction(Env env) {
   exports["callWithInvalidReceiver"] = Function::New(env, CallWithInvalidReceiver);
   exports["callConstructorWithArgs"] = Function::New(env, CallConstructorWithArgs);
   exports["callConstructorWithVector"] = Function::New(env, CallConstructorWithVector);
+  exports["callConstructorWithCStyleArray"] =
+      Function::New(env, CallConstructorWithCStyleArray);
   exports["isConstructCall"] = Function::New(env, IsConstructCall);
   exports["callWithFunctionOperator"] =
       Function::New(env, CallWithFunctionOperator);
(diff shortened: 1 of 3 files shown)


mhdawson:
@JoseExposito thanks for working on this.

I see in your output:

```
All tests passed!
Tests aborted with SIGSEGV
```

Do you see that every time? It indicates something is not quite right so wondering if its related to the PR or possibly an existing intermittent issue.

JoseExposito:
Hi @mhdawson 

I don't have permissions to re-run the failed jobs upstream, however, in my fork, they finished successfully:
https://github.com/JoseExposito/node-addon-api/actions/runs/638531308

Could you re-run them, please? My guess is that it is an intermittent crash on macOS, but let's see...

JoseExposito:
It worked :tada: It looks like an intermittent issue, none of the runs on my fork failed:
https://github.com/JoseExposito/node-addon-api/actions

test/function.cc
```diff
@@ -6,6 +6,18 @@ namespace {
 
 int testData = 1;
 
+Boolean EmptyConstructor(const CallbackInfo& info) {
```
NickNaso, line 9: Mybe it's possible to handle this case with one function instead of the two `EmptyConstructor` and `NonEmptyConstructor` something like this:
```cpp
Boolean EmptyConstructor(const CallbackInfo& info) {
  auto env = info.Env();
  bool is_empty = info[0].As<Boolean>();
  Function function = is_empty ? Function(): Function(env, Object::New(env));
  return Boolean::New(env, function.IsEmpty());
}
```
and from the JavaScript side:

```js
assert.strictEqual(binding.emptyConstructor(true), true);
assert.strictEqual(binding.emptyConstructor(false), false);
```

NickNaso (commented):
@JoseExposito thanks for adding this test cases.  Just some  comments.

test/function.cc
JoseExposito: Sure: https://github.com/nodejs/node-addon-api/pull/928/commits/da75b85dee3deec54a535f1264ca93156e65387c

NickNaso (approved):
LGTM.

mhdawson (approved):
LGTM

mhdawson:
I merged this but I found that the tests were failing after having done that with this output on 14.x

```JavaScript
Running test 'object/finalizer'
(node:1257166) UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: function should not have been called
at mustNotCall (/home/midawson/newpull/land/node-addon-api/test/common/index.js:74:12)
(Use node --trace-warnings ... to show where the warning was created)
(node:1257166) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag --unhandled-rejections=strict (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:1257166) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

and this output on 16.x

```
Running test 'object/finalizer'
node:internal/process/promises:227
          triggerUncaughtException(err, true /* fromPromise */);
          ^

AssertionError [ERR_ASSERTION]: function should not have been called
    at mustNotCall (/home/midawson/newpull/land/node-addon-api/test/common/index.js:74:12) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: undefined,
  expected: undefined,
  operator: 'fail'
}
```

So I've reverted it. Since it's in a different test it must be some interaction between the tests.

mhdawson:
@JoseExposito could you take a look and then submit and updated PR?

JoseExposito:
Good catch @mhdawson , I overlooked that log. I created a new PR (https://github.com/nodejs/node-addon-api/pull/937) fixing the issue.

PR #933, opened 2021-03-16
doc: sync Object::Set value arg with Value::From

tniessen:
The list of types that are valid for the `value` argument of `Napi::Object::Set` is inaccurate. Instead of trying to maintain it, let's just add a reference to `Napi::Value::From` since any argument to `Napi::Object::Set` is passed to `Napi::Value::From` anyway.


mhdawson (approved):
LGTM

PR #939, opened 2021-03-19
Add tests for global objects and missing object tests using uint32 as key

JckXia:
This pull request aims to test the various CRUD actions that we can perform on the Global object. Because C++ is statically typed and the keys could be of type int, string, C string or napi_value, each of these cases should be checked thoroughly.
(i.e Changes made from the c++ side should be reflected on the javascript side as well)
Also adding some missing object tests where they use uint32 as key


mhdawson:
@JckXia could you squash into 1 commit?  I tried to pull in locally but I think due to the intermediate changes it was reporting a conflict.

JckXia:
@mhdawson  Yeah of course! I will get to it.

test/binding.gyp
```diff
@@ -28,6 +28,11 @@
         'movable_callbacks.cc',
         'memory_management.cc',
         'name.cc',
+        'globalObject/delete_property.cc',
+        'globalObject/has_own_property.cc',
+        'globalObject/set_property.cc',
+        'globalObject/get_property.cc',
+        'globalObject/object.cc',
```
NickNaso, line 35: The build for the test fails on windows. In windows I'm getting the error reported below:
build-win
The problem is that on windows the object files will be created on `build/Release/obj/binding` or `build/Release/obj/binding_noexcept` instead in unix system the object files will be created on the subfolders you set. 
One solution for me  was to chenge the name of the following files:
```
 'globalObject/delete_property.cc',
 'globalObject/has_own_property.cc',
 'globalObject/set_property.cc',
 'globalObject/get_property.cc',
  'globalObject/object.cc',
```
to something like this:

```
 'globalObject/global_object_delete_property.cc',
 'globalObject/global_object_has_own_property.cc',
 'globalObject/global_object_set_property.cc',
 'globalObject/global_object_get_property.cc',
  'globalObject/global_object.cc',
```
This avoid the name conflicts for the object files generated on windows system.

JckXia: Okay thank you! I am just wondering, i noticed that we run the tests per PR on MacOS and ubuntu , should we setup a similar CI for a window systems to catch issues like this?

NickNaso: I agree with you and I just opened a new PR https://github.com/nodejs/node-addon-api/pull/948 that add the windows platform to CI.

test/globalObject/global_object_get_property.cc
```diff
@@ -0,0 +1,41 @@
...
+  Object globalObject = info.Env().Global();
+  String cppStrKey = info[0].As<String>();
+  return globalObject.Get(cppStrKey.Utf8Value());
+}
+
+void CreateMockTestObject(const CallbackInfo& info) {
+  Object globalObject = info.Env().Global();
+
+  napi_value napi_key;
+  napi_create_int32(info.Env(), 2, &napi_key);
```
NickNaso, line 33: Could you use `Napi::Number` here? 
```cpp
 Number napi_key = Number::New(info.Env(), 2);
```

test/globalObject/global_object_has_own_property.cc
```diff
@@ -0,0 +1,23 @@
...
+  Object globalObject = info.Env().Global();
+  String key = info[0].As<String>();
+  return Boolean::New(info.Env(), globalObject.HasOwnProperty(key.Utf8Value()));
+}
+
+Value HasPropertyWithNapiValueAsKey(const CallbackInfo& info) {
+  Object globalObject = info.Env().Global();
+  Name key = info[0].As<Name>();
+  return Boolean::New(
+      info.Env(), globalObject.HasOwnProperty(static_cast<napi_value>(key)));
```
NickNaso, line 22: I think that here it's not necessary to cast the `key` value.  You can do the same like reported below:
```cpp
return Boolean::New(
      info.Env(), globalObject.HasOwnProperty(key));
```

test/globalObject/global_object_set_property.cc
```diff
@@ -0,0 +1,31 @@
...
+  Number key = info[0].As<Number>();
+  Value value = info[1];
+  globalObject.Set(key.Uint32Value(), value);
+}
+
+void SetPropertyWithNapiValueAsKey(const CallbackInfo& info) {
+  Object globalObject = info.Env().Global();
+  Name key = info[0].As<Name>();
+  Value value = info[1];
+  globalObject.Set(static_cast<napi_value>(key), value);
```
NickNaso, line 30: I think that here it's not necessary to cast the key value. You can do the same like reported below:
```cpp
globalObject.Set(key, value);
```

NickNaso (approved):
LGTM

mhdawson (approved):
LGTM

mhdawson:
Landed as bc5147cc4ab7e79d6c11a08790082c958970b887 Thanks @JckXia

PR #942, opened 2021-03-20
doc: unambiguously mark deprecated signatures

tniessen:
Currently, deprecation notices are always right between two function signatures and it's virtually impossible to be certain whether they refer to the previous signature or the next signature:

```
`void foo()`

This signature is deprecated.

`void bar()`
```

It would be great if someone could confirm that it is indeed always the *above* signature that's deprecated.

### DIFF
--- a/doc/property_descriptor.md
+++ b/doc/property_descriptor.md
@@ -138,7 +138,7 @@ The name of the property can be any of the following types:
 - `napi_value value`
 - `Napi::Name`
 
-**This signature is deprecated. It will result in a memory leak if used.**
+**The above signature is deprecated. It will result in a memory leak if used.**
 
 ```cpp
 static Napi::PropertyDescriptor Napi::PropertyDescriptor::Accessor (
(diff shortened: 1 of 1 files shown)


legendecas (approved):

legendecas:
I believe it is the one above that been deprecated. See https://github.com/nodejs/node-addon-api/pull/343 and https://github.com/nodejs/node-addon-api/blob/main/napi.h#L1465.

/cc @gabrielschulhof

NickNaso (approved):

mhdawson:
Also by looking at at https://github.com/nodejs/node-addon-api/blob/main/napi-inl.deprecated.h looks to confirm that it is the "above" one that is deprecated.

mhdawson (approved):
LGTM

gabrielschulhof (approved):

gabrielschulhof:
Landed in 01278131117c7c81ffec2d0dd5f7e97f9f96f5a1.

PR #969, opened 2021-04-13
doc: correct struct definition

RaisinTen:

### DIFF
--- a/doc/version_management.md
+++ b/doc/version_management.md
@@ -25,8 +25,7 @@ information is stored in the `napi_node_version` structure that is defined as
 shown below:
 
 ```cpp
-using napi_node_version =
-struct {
+typedef struct {
   uint32_t major;
   uint32_t minor;
   uint32_t patch;


doc/version_management.md
```diff
@@ -25,13 +25,12 @@ information is stored in the `napi_node_version` structure that is defined as
 shown below:
 
 ```cpp
-using napi_node_version =
-struct {
+using napi_node_version = struct {
```
legendecas, line 28: I believe the intention of this code block is just showing how `struct napi_node_version` is defined in `node_api_types.h`, but not how users should define the types. As the type `struct napi_node_version` is defined as followings:

```c
typedef struct {
  uint32_t major;
  uint32_t minor;
  uint32_t patch;
  const char* release;
} napi_node_version;
```

I believe the best way to do this here is to keep aligned with `node_api_types.h` or just remove the code block to prevent confusion.

legendecas (changes requested):

doc/version_management.md
RaisinTen: @legendecas I think it would be better to keep the code block. I have a PR open to replace the `typedef`s with `using`s in main repo that also changes this part in the file: https://github.com/nodejs/node/pull/38228/files#diff-9751c3fa7589557e1988160cbf907aa233c3cff86e1320fbc634d80394c918edR37-R42. Maybe we should wait for a confirmation on that PR before changing this?

legendecas: What I really mean is that header files like `node_api.h`, `js_native_api.h`, and so on are C headers but not C++ headers. We are not intended to use C++ features like `using` declarations in them. It's by design.

RaisinTen: @legendecas Thanks for letting me know that it's a C header. Updated the PR. PTAL.

legendecas (approved):

mhdawson (approved):
LGTM

PR #972, opened 2021-04-20
Added test coverage for Symbol class

JckXia:
Added new tests for the Symbol class. Also added a static method to the Symbol class to perform "Symbol.For()" from the C++ side, since the exisiting method (Symbol::Wellknown) , which has a different functionality caused some confusion.

### DIFF
--- a/doc/symbol.md
+++ b/doc/symbol.md
@@ -34,7 +34,7 @@ If an error occurs, a `Napi::Error` will get thrown. If C++ exceptions are not
 being used, callers should check the result of `Napi::Env::IsExceptionPending` before
 attempting to use the returned value.
 
-### Utf8Value
+### WellKnown
 ```cpp
 static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& name);
 ```
@@ -44,5 +44,18 @@ static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& nam
 
 Returns a `Napi::Symbol` representing a well-known `Symbol` from the
 `Symbol` registry.
+
+### For
+```cpp
+static Napi::Symbol Napi::Symbol::For(napi_env env, const std::string& description);
+static Napi::Symbol Napi::Symbol::For(napi_env env, const char* description);
+static Napi::Symbol Napi::Symbol::For(napi_env env, String description);
+static Napi::Symbol Napi::Symbol::For(napi_env env, napi_value description);
+```
+
+- `[in] env`: The `napi_env` environment in which to construct the `Napi::Symbol` object.
+- `[in] description`: The C++ string representing the `Napi::Symbol` in the global registry to retrieve.
+
+Searches in the global registry for existing symbol with the given name. If the symbol already exist it will be returned, otherwise a new symbol will be created in the registry. It's equivalent to Symbol.for() called from JavaScript.
 
 [`Napi::Name`]: ./name.md
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -998,6 +998,27 @@ inline Symbol Symbol::WellKnown(napi_env env, const std::string& name) {
   return Napi::Env(env).Global().Get("Symbol").As<Object>().Get(name).As<Symbol>();
 }
 
+inline Symbol Symbol::For(napi_env env, const std::string& description) {
+  napi_value descriptionValue = String::New(env, description);
+  return Symbol::For(env, descriptionValue);
+}
+
+inline Symbol Symbol::For(napi_env env, const char* description) {
+  napi_value descriptionValue = String::New(env, description);
+  return Symbol::For(env, descriptionValue);
+}
+
+inline Symbol Symbol::For(napi_env env, String description) {
+  return Symbol::For(env, static_cast<napi_value>(description));
+}
+
+inline Symbol Symbol::For(napi_env env, napi_value description) {
+  Object symbObject = Napi::Env(env).Global().Get("Symbol").As<Object>();
+  auto forSymb =
+      symbObject.Get("for").As<Function>().Call(symbObject, {description});
+  return forSymb.As<Symbol>();
+}
+
 inline Symbol::Symbol() : Name() {
 }
 
--- a/napi.h
+++ b/napi.h
@@ -538,6 +538,18 @@ namespace Napi {
    /// Get a public Symbol (e.g. Symbol.iterator).
    static Symbol WellKnown(napi_env, const std::string& name);
 
+   // Create a symbol in the global registry, UTF-8 Encoded cpp string
+   static Symbol For(napi_env env, const std::string& description);
+
+   // Create a symbol in the global registry, C style string (null terminated)
+   static Symbol For(napi_env env, const char* description);
+
+   // Create a symbol in the global registry, String value describing the symbol
+   static Symbol For(napi_env env, String description);
+
+   // Create a symbol in the global registry, napi_value describing the symbol
+   static Symbol For(napi_env env, napi_value description);
+
    Symbol();  ///< Creates a new _empty_ Symbol instance.
    Symbol(napi_env env,
           napi_value value);  ///< Wraps a Node-API value primitive.
--- a/test/symbol.cc
+++ b/test/symbol.cc
@@ -1,0 +1,77 @@
+#include <napi.h>
+using namespace Napi;
+
+Symbol CreateNewSymbolWithNoArgs(const Napi::CallbackInfo&) {
+  return Napi::Symbol();
+}
+
+Symbol CreateNewSymbolWithCppStrDesc(const Napi::CallbackInfo& info) {
+  String cppStrKey = info[0].As<String>();
+  return Napi::Symbol::New(info.Env(), cppStrKey.Utf8Value());
+}
+
+Symbol CreateNewSymbolWithCStrDesc(const Napi::CallbackInfo& info) {
+  String cStrKey = info[0].As<String>();
+  return Napi::Symbol::New(info.Env(), cStrKey.Utf8Value().c_str());
+}
+
+Symbol CreateNewSymbolWithNapiString(const Napi::CallbackInfo& info) {
+  String strKey = info[0].As<String>();
+  return Napi::Symbol::New(info.Env(), strKey);
+}
+
+Symbol GetWellknownSymbol(const Napi::CallbackInfo& info) {
+  String registrySymbol = info[0].As<String>();
+  return Napi::Symbol::WellKnown(info.Env(),
+                                 registrySymbol.Utf8Value().c_str());
+}
+
+Symbol FetchSymbolFromGlobalRegistry(const Napi::CallbackInfo& info) {
+  String registrySymbol = info[0].As<String>();
+  return Napi::Symbol::For(info.Env(), registrySymbol);
+}
+
+Symbol FetchSymbolFromGlobalRegistryWithCppKey(const Napi::CallbackInfo& info) {
+  String cppStringKey = info[0].As<String>();
+  return Napi::Symbol::For(info.Env(), cppStringKey.Utf8Value());
+}
+
+Symbol FetchSymbolFromGlobalRegistryWithCKey(const Napi::CallbackInfo& info) {
+  String cppStringKey = info[0].As<String>();
+  return Napi::Symbol::For(info.Env(), cppStringKey.Utf8Value().c_str());
+}
+
+Symbol TestUndefinedSymbolsCanBeCreated(const Napi::CallbackInfo& info) {
+  Napi::Env env = info.Env();
+  return Napi::Symbol::For(env, env.Undefined());
+}
+
+Symbol TestNullSymbolsCanBeCreated(const Napi::CallbackInfo& info) {
+  Napi::Env env = info.Env();
+  return Napi::Symbol::For(env, env.Null());
+}
+
+Object InitSymbol(Env env) {
+  Object exports = Object::New(env);
+
+  exports["createNewSymbolWithNoArgs"] =
+      Function::New(env, CreateNewSymbolWithNoArgs);
+  exports["createNewSymbolWithCppStr"] =
+      Function::New(env, CreateNewSymbolWithCppStrDesc);
+  exports["createNewSymbolWithCStr"] =
+      Function::New(env, CreateNewSymbolWithCStrDesc);
+  exports["createNewSymbolWithNapi"] =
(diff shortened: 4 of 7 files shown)


JckXia:
I can't seem to recreate this test failure. It seems intermittent as the CI running against this commit is successful in my forked repo. https://github.com/JckXia/node-addon-api/actions/runs/765661855.

napi.h
```diff
@@ -538,6 +538,9 @@ namespace Napi {
    /// Get a public Symbol (e.g. Symbol.iterator).
    static Symbol WellKnown(napi_env, const std::string& name);
 
+   // Create a symbol in the global registry;
+   static Symbol For(napi_env env, const std::string& name);
```
NickNaso, line 542: Do we need some other overloads for this methods?
```cpp
static Symbol For(napi_env env,  const char* description = nullptr);
static Symbol For(napi_env env,  String description);
static Symbol For(napi_env env, napi_value description);
```
What do you think about?

NickNaso, line 542: Please remember to add documentation for this new method.

test/symbol.cc
```diff
@@ -0,0 +1,51 @@
+#include <napi.h>
+
+using namespace Napi;
+
+Symbol CreateNewSymbolWithNoArgs(const Napi::CallbackInfo& info) {
+  (void)info;
```
NickNaso, line 6: Why does it necessary do this?
```cpp
(void)info;
```

test/symbol.js
```diff
@@ -0,0 +1,65 @@
...
+    for(const wellknownProperty of wellKnownSymbolFunctions)
+    {
+      assertSymbolIsWellknown(wellknownProperty);
+    }
+
+    assertCanCreateOrFetchGlobalSymbols("keys");
+    
+}
+
+
```
NickNaso, line 65: Please remove one empty new line.

test/symbol.cc
```diff
@@ -0,0 +1,51 @@
...
+      Function::New(env, CreateNewSymbolWithCppStrDesc);
+  exports["createNewSymbolWithCStr"] =
+      Function::New(env, CreateNewSymbolWithCStrDesc);
+  exports["createNewSymbolWithNapi"] =
+      Function::New(env, CreateNewSymbolWithNapiString);
+  exports["getWellKnownSymbol"] = Function::New(env, GetWellknownSymbol);
+  exports["getSymbolFromGlobalRegistry"] =
+      Function::New(env, FetchSymbolFromGlobalRegistry);
+  return exports;
+}
```
NickNaso, line 51: PLease add one empty new line.

NickNaso (commented):
@JckXia well done. Just some comments.

NickNaso:
> I can't seem to recreate this test failure. It seems intermittent as the CI running against this commit is successful in my forked repo. https://github.com/JckXia/node-addon-api/actions/runs/765661855.

Don't worry I restarted the jobs and now it's all green.

test/symbol.cc
JckXia: I think this was to suppress compiler warnings for unused parameter, since the default Symbol constructor doesn't take in any values. But i can remove this function since it's not being called from js.

napi.h
JckXia: Hmm yeah I think that will be a good idea. Will add them in soon.

test/symbol.cc
NickNaso: To avoid compiler complains maybe you can try the following solution (comment the formal parameter of the function):
```cpp
Symbol CreateNewSymbolWithNoArgs(const Napi::CallbackInfo& /*info*/) {
```
Let me know if it worked.

JckXia: Yep it works. Thanks!

doc/symbol.md
```diff
@@ -34,7 +34,7 @@ If an error occurs, a `Napi::Error` will get thrown. If C++ exceptions are not
 being used, callers should check the result of `Napi::Env::IsExceptionPending` before
 attempting to use the returned value.
 
-### Utf8Value
+### Wellkown
```
NickNaso, line 37: Please fix the typo it should be **WellKnown**

doc/symbol.md
```diff
@@ -45,4 +45,14 @@ static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& nam
 Returns a `Napi::Symbol` representing a well-known `Symbol` from the
 `Symbol` registry.
 
+### For
+```cpp
+static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& name);
+```
+
```
NickNaso, line 52: Please add the documentation for all the overloads of `Napi::Symbol Napi::Symbol::WellKnown()` methos.

doc/symbol.md
```diff
@@ -45,4 +45,14 @@ static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& nam
...
+- `[in] env`: The `napi_env` environment in which to construct the `Napi::Symbol` object.
+- `[in] name`: The C++ string representing the `Napi::Symbol` to retrieve.
+
+Register `Napi::Symbol` in the global registry. If symbol already exist retrieve said symbol. Equivalent to `Symbol.for("symb")` called from JS.
```
NickNaso, line 56: Maybe the following description is better: 
Searches in the global registry for existing symbol with the given name. If the symbol already exist it will be returned, otherwise a new symbol will be created in the registry. It's equivalent to `Symbol.for()` called from JavaScript.

napi-inl.h
```diff
@@ -998,6 +998,30 @@ inline Symbol Symbol::WellKnown(napi_env env, const std::string& name) {
...
+
+inline Symbol Symbol::For(napi_env env, const char* description) {
+  napi_value descriptionValue = description != nullptr
+                                    ? String::New(env, description)
+                                    : static_cast<napi_value>(nullptr);
+  return Symbol::For(env, descriptionValue);
+}
+
+inline Symbol Symbol::For(napi_env env, String description) {
+  napi_value descriptionValue = description;
```
NickNaso, line 1014: You can directory pass `description` to `Symbol::For()` because `Napi::String` is a `Napi::Value` that has the cast operator `napi_value`.

NickNaso (commented):
Hi @JckXia,
I added som e comments about you last work.

doc/symbol.md
```diff
@@ -45,4 +45,14 @@ static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& nam
 Returns a `Napi::Symbol` representing a well-known `Symbol` from the
 `Symbol` registry.
 
+### For
+```cpp
+static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& name);
```
KevinEady, line 50: You are adding the `For` method... should this be `For`...?

doc/symbol.md
```diff
@@ -34,7 +34,7 @@ If an error occurs, a `Napi::Error` will get thrown. If C++ exceptions are not
 being used, callers should check the result of `Napi::Env::IsExceptionPending` before
 attempting to use the returned value.
 
-### Utf8Value
+### Wellknown
```
NickNaso, line 37: It shlould be `WellKnown` insteand of `Wellknown`. The `k` should be in uppercase.

doc/symbol.md
```diff
@@ -45,4 +45,17 @@ static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& nam
 Returns a `Napi::Symbol` representing a well-known `Symbol` from the
 `Symbol` registry.
 
+### For
+```cpp
+static Napi::Symbol Napi::Symbol::For(napi_env env, const std::string& description);
+static Napi::Symbol For(napi_env env, const char* description = nullptr);
+static Napi::Symbol For(napi_env env, String description);
+static Napi::Symbol For(napi_env env, napi_value description);
```
NickNaso, line 53: Please use the complete namespace there:

```cpp
static Napi::Symbol Napi::Symbol::For(napi_env env, const char* description = nullptr);
static Napi::Symbol Napi::Symbol::For(napi_env env, String description);
static Napi::Symbol Napi::Symbol::For(napi_env env,napi_value description);

```

NickNaso (commented):
Hi @JckXia,
just something to fix.

NickNaso (commented):
Hi @JckXia,
could you propose the last change on the package.json on a different PR? I'm asking for this because the change it's not correlated with the test and in case we will need to revert the change it will be more simpler to do. For the rest it's all ok.

JckXia:
Hey @NickNaso 
Thanks for the review! I think I accidentally made some changes to package.json and checked it into git for some reason and was trying to revert it.  There was no changes needed to be made to package.json and I will try to revert the changes made to it. Sorry about that.

NickNaso (approved):
LGTM

napi.h
```diff
@@ -538,6 +538,18 @@ namespace Napi {
...
+
+   // Create a symbol in the global registry, C style string (null terminated)
+   static Symbol For(napi_env env, const char* description = nullptr);
+
+   // Create a symbol in the global registry, String value describing the symbol
+   static Symbol For(napi_env env, String description);
+
+   // Create a symbol in the global registry, napi_value describing the symbol
+   static Symbol For(napi_env env, napi_value description);
+
```
gabrielschulhof, line 552:
```suggestion
   static Symbol For(Env env, const std::string& description);

   // Create a symbol in the global registry, C style string (null terminated)
   static Symbol For(Env env, const char* description = nullptr);

   // Create a symbol in the global registry, String value describing the symbol
   static Symbol For(Env env, String description);

   // Create a symbol in the global registry, napi_value describing the symbol
   static Symbol For(Env env, napi_value description);

```

napi-inl.h
```diff
@@ -998,6 +998,29 @@ inline Symbol Symbol::WellKnown(napi_env env, const std::string& name) {
   return Napi::Env(env).Global().Get("Symbol").As<Object>().Get(name).As<Symbol>();
 }
 
+inline Symbol Symbol::For(napi_env env, const std::string& description) {
```
gabrielschulhof, line 1001:
```suggestion
inline Symbol Symbol::For(Napi::Env env, const std::string& description) {
```

... and a similar change for all APIs below.

gabrielschulhof (changes requested):
Blocking until we figure out what happens when you pass `nullptr` to `napi_create_string_utf8()` as its `const char*` parameter.

gabrielschulhof:
@JckXia thanks for catching this!

gabrielschulhof:
We must never pass `nullptr` into `napi_create_string_*()` because that will cause a crash. Thus, we should actually change *all* the APIs (even existing ones) that end up calling `napi_create_string_*()` to return `napi_invalid_arg` when passed in what evaluates to `(const char *)nullptr`.

I'll make a PR upstream to fix this in core.

JckXia:
Thanks for the review @gabrielschulhof . Just making sure, are we converting all function signature from For(Napi::Env env, const std::string & val),etc to For(napi_env env, const std::string & val), or are we keeping both?

gabrielschulhof:
I checked napi.h by looking for `[(]napi_env` and it turns out most of our APIs take `napi_env` as the first argument so it's OK to keep the `napi_env` and we need neither replace nor provide overloads. So, LGTM except for the passing-null-to-`napi_create_string_*` part.

gabrielschulhof:
Refs: https://github.com/nodejs/node/pull/38923

mhdawson:
@JckXia just wondering what the next steps are to get this landed?

JckXia:
@mhdawson  I believe that it is ready to be merged.

doc/symbol.md
```diff
@@ -45,4 +45,17 @@ static Napi::Symbol Napi::Symbol::WellKnown(napi_env env, const std::string& nam
 Returns a `Napi::Symbol` representing a well-known `Symbol` from the
 `Symbol` registry.
 
-[`Napi::Name`]: ./name.md
+### For
+```cpp
+static Napi::Symbol Napi::Symbol::For(napi_env env, const std::string& description);
+static Napi::Symbol Napi::Symbol::For(napi_env env, const char* description = nullptr);
```
KevinEady, line 51: Hi @JckXia ,

Do we have a test for this `nullptr` value...? I am not sure what happens when using a nullptr `napi_value` for the `description` here:

```cpp
  auto forSymb =
      symbObject.Get("for").As<Function>().Call(symbObject, {description});
```

and I think a test would be beneficial

JckXia:
@KevinEady Thanks for catching that! Apparently if we simply pass the C++ nullptr into that function it will cause a SegFault. I updated the Symbol::For method impl to use the napi_value we get from calling the napi_get_null function instead and it seems to work.

KevinEady:
> @KevinEady Thanks for catching that! Apparently if we simply pass the C++ nullptr into that function it will cause a SegFault. I updated the Symbol::For method impl to use the napi_value we get from calling the napi_get_null function instead and it seems to work.

@JckXia  Aahahh... I am inclined then to remove the `std::nullptr` default and the "check if nullptr, assume napi_get_null" ... Because that is not something we have in other locations. For example,

```js
{
  [null]: "3"
}
```
is a valid JSON object, but we do not check in `Object::Get` for using `const char* utf8name` with `std::nullptr`:

https://github.com/nodejs/node-addon-api/blob/ad76ad07f914fab02be5778ec67485916c4626d9/napi-inl.h#L1227-L1232

The application would just crash (I assume). I would think, for consistency purposes, we should remove the nullptr default as well as the runtime check (which is additional overhead anyway).

If the user wanted to explicitly use `null`, they could use `Symbol::For(env, env.Null())`.

Thoughts?

JckXia:
@KevinEady Oh okay I see. So I think we can simply pass the description argument into String::New(), and pass the newly constructed Napi string the Symbol::For function?  Also, I think this ` Symbol::For(env, env.Null())` is a good approach, as I believe they could do something similar for `undefined`. Will update the PR and later today. :)

KevinEady:
This depends on landing #1015 first.

gabrielschulhof:
Please feel free to dismiss my change request after #1015 has landed and the `nullptr` default for the `char *` argument was removed!

KevinEady:
As discussed in today's meeting, please remove the `std::nullptr` default value in `Symbol::For`.

mhdawson:
1015 landed, dismissing @gabrielschulhof review as suggested above.

mhdawson (approved):
LGTM

JckXia:
@mhdawson Thanks, it's been fixed.

mhdawson:
Landed in e02e8a4ce3c24f9e833b771cff7aac51658cf2ee

PR #975, opened 2021-04-22
src: fix a crashing issue in Error::ThrowAsJavaScriptException via NODE_API_SWALLOW_UNTHROWABLE_EXCEPTIONS

KevinEady:
When terminating an environment (e.g., by calling worker.terminate), napi_throw, which is called by Error::ThrowAsJavaScriptException, returns napi_pending_exception, which is then incorrectly treated as a fatal error resulting in a crash. In order to not terminate the Node process, define the preprocessor directive `NODE_API_SWALLOW_UNTHROWABLE_EXCEPTIONS`.

### DIFF
--- a/common.gypi
+++ b/common.gypi
@@ -15,7 +15,6 @@
       }
     }]
   ],
-  'defines': [ 'NODE_API_SWALLOW_UNTHROWABLE_EXCEPTIONS' ],
   'include_dirs': ["<!(node -p \"require('../').include_dir\")"],
   'cflags': [ '-Werror', '-Wall', '-Wextra', '-Wpedantic', '-Wunused-parameter' ],
   'cflags_cc': [ '-Werror', '-Wall', '-Wextra', '-Wpedantic', '-Wunused-parameter' ]
--- a/common.gypi
+++ b/common.gypi
@@ -15,6 +15,7 @@
       }
     }]
   ],
+  'defines': [ 'NODE_API_SWALLOW_UNTHROWABLE_EXCEPTIONS' ],
   'include_dirs': ["<!(node -p \"require('../').include_dir\")"],
   'cflags': [ '-Werror', '-Wall', '-Wextra', '-Wpedantic', '-Wunused-parameter' ],
   'cflags_cc': [ '-Werror', '-Wall', '-Wextra', '-Wpedantic', '-Wunused-parameter' ]
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2393,7 +2393,8 @@ inline void Error::ThrowAsJavaScriptException() const {
     // new one as that is not allowed/possible
     napi_status status = napi_is_exception_pending(_env, &pendingException);
 
-    if ((status == napi_ok) && (pendingException == false)) {
+    if ((status != napi_ok) ||
+        ((status == napi_ok) && (pendingException == false))) {
       // We intentionally don't use `NAPI_THROW_*` macros here to ensure
       // that there is no possible recursion as `ThrowAsJavaScriptException`
       // is part of `NAPI_THROW_*` macro definition for noexcept.
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2418,11 +2418,6 @@ inline void Error::ThrowAsJavaScriptException() const {
     napi_status status = napi_throw(_env, Value());
 #endif
 
-    if (status == napi_pending_exception) {
-      // The environment could be terminating.
-      return;
-    }
-
 #ifdef NAPI_CPP_EXCEPTIONS
     if (status != napi_ok) {
       throw Error::New(_env);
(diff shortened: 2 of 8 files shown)


napi-inl.h
```diff
@@ -2386,12 +2386,42 @@ inline const std::string& Error::Message() const NAPI_NOEXCEPT {
...
     // that there is no possible recursion as `ThrowAsJavaScriptException`
     // is part of `NAPI_THROW_*` macro definition for noexcept.
 
     napi_status status = napi_throw(_env, Value());
+#endif
+
+    if (status == napi_pending_exception) {
+      // The environment could be terminating.
+      return;
+    }
```
mhdawson, line 2424: I don't think this block should be added here.

common.gypi
```diff
@@ -15,6 +15,7 @@
       }
     }]
   ],
+  'defines': [ 'NODE_API_SWALLOW_UNTHROWABLE_EXCEPTIONS' ],
```
mhdawson, line 18: This should not be here by default right ?

KevinEady: Hi @mhdawson ,

So this relates to my comment on https://github.com/nodejs/node-addon-api/pull/902#issuecomment-817291009 where:
> Not having the define in binding.gyp and using #define prior to #include <napi.h>: I was not able to opt-in.

You can see in commit https://github.com/nodejs/node-addon-api/pull/975/commits/a089697d9e1e37d83b834a110c585fed06d8538b the change to move the setting from `binding.gyp` to `#define` and now _all_ of the builds fail.

I do not think it is possible to have a per-compilation unit setting for this feature, because the `Error::ThrowAsJavaScriptException` is a compiled, exported symbol:

```
➜  test git:(swallow-unthrowable-exceptions) ✗ nm -C build/Release/binding_noexcept.node | grep ThrowAsJavaScriptException
00000000000024f0 t Napi::Error::ThrowAsJavaScriptException() const
```

So, when the first compilation unit uses it, it will be compiled with whatever preprocessor definitions are currently in scope.

napi-inl.h
KevinEady: Addressed.

KevinEady:
Hi @mhdawson and team. This PR is ready for review. As discussed in last week's meeting, the last commit contains (1) the implementation change and (2) creation of new test target.  Just struggling with the CI.

KevinEady:
@legendecas I believe you are working on build system changes. Since this introduces new target, will possibly need to collaborate between the two PRs.

mhdawson:
@KevinEady, if the build changes were to refactor the testing so that it is easier to add a new config that we run the tests  under, unless I'm remembering wrong, I think I already landed that.

KevinEady:
> @KevinEady, if the build changes were to refactor the testing so that it is easier to add a new config that we run the tests under, unless I'm remembering wrong, I think I already landed that.

Hi @mhdawson , no the build changes were to create a new target that only compiles the single file with the `NODE_API_SWALLOW_UNTHROWABLE_EXCEPTIONS` define.

mhdawson:
@KevinEady got it.

legendecas (approved):

mhdawson:
@KevinEady I'm wondering if this is still waiting on the other PR mentioned or if its ready to go?

KevinEady:
Hi @mhdawson ,

This PR is ready to go and indeed addresses the issue in #902. I am wondering though if there is any real-world code that we can see if this fix actually solves the problem, cc: @rudolftam ?

Thanks, Kevin

mhdawson (approved):
LGTM

mhdawson:
I'll just hold off landing this until the 4.0 release goes out.

mhdawson:
Actually thinking about it, I don't see the harm of it being in 4.x so landing now.

mhdawson:
Landed as 028107f686f71101abd4794f29c82f25841d5a55

mhdawson:
@KevinEady I've landed but I am wondering about testing with the additional define. Should I see additional output when I run npm test or is that all hidden behind the scenes and I should only see one line for each test?

KevinEady:
Hi @mhdawson ,

There are two additional build targets ([1](https://github.com/nodejs/node-addon-api/runs/2450198938#step:5:754), [2](https://github.com/nodejs/node-addon-api/runs/2450198938#step:5:758)) and one test ([1](https://github.com/nodejs/node-addon-api/runs/2450198938#step:5:791)) that runs everything in the background via a separate node process:

https://github.com/nodejs/node-addon-api/blob/3f291f71bc1e083f87930f770cd4d0f412b63cc6/test/error_terminating_environment.js#L47-L64

rudolftam:
@KevinEady re: I am wondering though if there is any real-world code that we can see if this fix actually solves the problem

It was in one of my unreleased hobby projects in which I ran into this issue. It uses Node.js as an embedded library to provide scripting support for the end-user. As the scripts must be terminable (mainly for the same reasons as in web browsers), everything around them must be capable of handling it without crashing.

I'm glad that this is now possible when using node-addon-api. I'm worried, though, that the opt-in nature of this fix will mean that most addons will needlessly keep crashing in the future as well.

```javascript
while (true) {
	console.log(`I hope you're not planning on terminating me. :)`)
	
	do_something_with_addon_a_fix_enabled()
	do_something_with_addon_b_fix_disabled()
	do_something_with_addon_c_fix_disabled()
	do_something_with_addon_d_fix_disabled()
}
```

PR #976, opened 2021-04-23
test: common test runners

legendecas:
Collects test entries so that we can run tests one by another. Also, it makes adding new test cases and test suites easier, i.e. we don't have to add the addon path to every test case file.


mhdawson (approved):
LGTM

mhdawson:
Landed as b6f5eb15e6cdf378e4c83db4cce85650c0e12ba1

PR #977, opened 2021-04-23
src: return bool on object set and define property

legendecas:
If c++ exception is disabled, there is no ergonomic way to detect
failtures on void returning.

### DIFF
--- a/doc/object.md
+++ b/doc/object.md
@@ -72,7 +72,7 @@ Creates a new `Napi::Object` value.
 ### Set()
 
 ```cpp
-void Napi::Object::Set (____ key, ____ value);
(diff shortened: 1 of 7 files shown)


legendecas:
@nodejs/node-api it will be great for this PR to have reviews, thanks a lot :D

KevinEady (approved):
Hi @legendecas , just a clarification that changing return type from `void` to `bool` is not breaking, correct....? I don't see it as a problem, but perhaps there is something I am missing...

legendecas:
@KevinEady yes, this is not a breaking change. Existing code will continue to compile and works (and there are still lots of them in the test suites), like https://github.com/nodejs/node-addon-api/blob/main/test/binding.cc#L76.

PR #991, opened 2021-05-05
src: return bool on object freeze and seal

legendecas:
These operations can call into JavaScript by using Proxy handlers,
hence JavaScript exceptions might pending at the end of the operations.

This is not a breaking change.


mhdawson (approved):
LGTM

PR #992, opened 2021-05-05
src: remove const qualifier on Object::GetPropertyNames and Object::InstanceOf

legendecas:
These operations can call into JavaScript and result in arbitrary
side effects, hence shall not be considered as `const` member functions.

E.g. the following examples intercepts the operations by Proxy handlers:

```javascript
const obj = new Proxy({
  foo: 0,
}, {
  ownKeys(target) {
    target.foo++;
    return [ 'foo' ];
  },
  getPrototypeOf(target) {
    target.foo++;
    return Object.prototype;
  },
});

Object.getOwnPropertyNames(obj);
obj.foo // => 1
obj instanceof Object;
obj.foo // => 2
```

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1315,7 +1315,7 @@ inline bool Object::Delete(uint32_t index) {
   return result;
 }
 
-inline Array Object::GetPropertyNames() const {
+inline Array Object::GetPropertyNames() {
   napi_value result;
   napi_status status = napi_get_property_names(_env, _value, &result);
   NAPI_THROW_IF_FAILED(_env, status, Array());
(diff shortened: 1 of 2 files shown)


mhdawson:
@legendecas we might need to consider this breaking, as code may no longer compile...

mhdawson:
There has been some discussion as to what const means. Maybe we should discuss in the next team meeting.

legendecas:
@mhdawson Yes, this is a breaking change. Apart from the const semantics, do we have any policies on node-addon-api like semver-major changes or breaking changes?

legendecas:
Labeled as "do not land" until we concluded with a policy regarding the breaking changes and find a consensus on `const` qualifier semantics on node-addon-api.

legendecas:
Discussed at today's node-api meeting. This is quite similar to const pointers and pointer to const objects. These functions did not change the `Napi::*` c++ objects so it seems reasonable to keep them `const`.

Issue #998, opened 2021-05-13
Reference count of ObjectWrap changed in Node.js 14.17.0


oyyd:
Not sure if this is an issue of `napi` or `node-addon-api`, but we found this when using `node-addon-api` so I post it here.

The code below:

```cpp
#include <napi.h>

class MyObject : public Napi::ObjectWrap<MyObject> {
 public:
  static Napi::Object Init(Napi::Env env, Napi::Object exports);
  MyObject(const Napi::CallbackInfo& info);
  ~MyObject();
};

Napi::Object MyObject::Init(Napi::Env env, Napi::Object exports) {
  Napi::Function func =
      DefineClass(env,
                  "MyObject",
                  {});

  Napi::FunctionReference* constructor = new Napi::FunctionReference();
  *constructor = Napi::Persistent(func);
  env.SetInstanceData(constructor);

  exports.Set("MyObject", func);
  return exports;
}

MyObject::MyObject(const Napi::CallbackInfo& info)
    : Napi::ObjectWrap<MyObject>(info) {
  this->Ref(); // Nothing special except that we Ref() here to keep this object from been garbage collected while other resources still alive
}

MyObject::~MyObject() {
  uint32_t count = -1; 
  count = this->Unref(); // And Unref() in destructor.
  fprintf(stderr, "count %d\n", count); // Log the count.
}

Napi::Object InitAll(Napi::Env env, Napi::Object exports) {
  return MyObject::Init(env, exports);
}

NODE_API_MODULE(addon, InitAll)
```

```js
var addon = require('bindings')('addon');

new addon.MyObject(10); // Create an instance and wait the process to exit.
```

Running the code in Node.js 14.16 will get:

```
count 0
```

In Node.js 14.17, the result becomes:

```
count 1
```

~~Though the code snippet doesn't throw here, in our real programs, running similar codes in Node.js 14.17 causes:~~

When enable cpp exceptions i.e. `NAPI_CPP_EXCEPTIONS`, run the script will result in:
```
// Mac
libc++abi.dylib: terminating with uncaught exception of type Napi::Error
// Debian
terminate called after throwing an instance of 'Napi::Error'
  what():  Unknown failure
Aborted (core dumped)
```

legendecas:
@oyyd can you share the core dump backtrace in your realm programs? (you can wipe out sensitive data from the text backtrace) It will be helpful to track down the problem since you said the code snippet doesn't crash.

oyyd:
[...]

PR #1004, opened 2021-05-23
src: set default return value of Reference Ref/Unref to 0

legendecas:
Update the default value to reflect the most possible values when there are any errors occurred on `napi_reference_unref`.

See: https://github.com/nodejs/node/blob/master/src/js_native_api_v8.cc#L2535
Related: https://github.com/nodejs/node-addon-api/issues/998#issuecomment-846593458

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2583,7 +2583,7 @@ template <typename T>
 inline uint32_t Reference<T>::Ref() {
   uint32_t result;
   napi_status status = napi_reference_ref(_env, _ref, &result);
-  NAPI_THROW_IF_FAILED(_env, status, 1);
+  NAPI_THROW_IF_FAILED(_env, status, 0);
   return result;
 }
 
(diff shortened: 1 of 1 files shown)


mhdawson:
I'm not sure returning 1 instead of 0 when an error occurs really helps.  What would make more sense to me is -1 to indicate an error.   If we are going to change the return value, which would affect code that is not properly checking for an exception. I think -1 would  be a better value going forward.

From the original post where an issue was reported:

I agree that calling count = this->Unref(); in the destructor does not make sense since if you count on that decrementing the ref count it will never happen except during env teardown and in that case it is not needed anyway.

So the fix for the original issue reported I think is to just remove the Unref() in the ~MyObject() destructor.

legendecas:
The reference count type is `uint32_t` so that we could not return -1 in the case of error.

mhdawson:
@legendecas right did not see it being uint32_t :(

mhdawson (approved):
LGTM

mhdawson:
Landed as 37a9b8e75384b6251decd090e0a0bb08072c3c20

Issue #1007, opened 2021-06-03
Build fails with gcc 11 / C++20


urp:
# Problem
Compiler error when including `node.h` with `NAPI_VERSION=8`:

```
napi.h:2477:60: error: expected unqualified-id before ‘)’ token
 2477 |     TypedThreadSafeFunction<ContextType, DataType, CallJs>();
      |                                                            ^
/home/adiletkutmanov/Development/lauda_measurement_napi/node_modules/node-addon-api/napi.h:2479:33: error: expected ‘)’ before ‘tsFunctionValue’
 2479 |         napi_threadsafe_function tsFunctionValue);
      |                                 ^~~~~~~~~~~~~~~~
      |                                 )
/home/adiletkutmanov/Development/lauda_measurement_napi/node_modules/node-addon-api/napi.h:2478:59: note: to match this ‘(’
 2478 |     TypedThreadSafeFunction<ContextType, DataType, CallJs>(
      |                                                           ^
In file included from /home/adiletkutmanov/Development/lauda_measurement_napi/node_modules/node-addon-api/napi.h:2725,
                 from /home/adiletkutmanov/Development/lauda_measurement_napi/src/lauda_measurement.cpp:1:
/home/adiletkutmanov/Development/lauda_measurement_napi/node_modules/node-addon-api/napi-inl.h:4764:8: error: no declaration matches ‘Napi::TypedThreadSafeFunction<ContextType, DataType, CallJs>::TypedThreadSafeFunction()’
 4764 | inline TypedThreadSafeFunction<ContextType, DataType, CallJs>::
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/adiletkutmanov/Development/lauda_measurement_napi/node_modules/node-addon-api/napi-inl.h:4764:8: note: no functions named ‘Napi::TypedThreadSafeFunction<ContextType, DataType, CallJs>::TypedThreadSafeFunction()’
In file included from /home/adiletkutmanov/Development/lauda_measurement_napi/src/lauda_measurement.cpp:1:
/home/adiletkutmanov/Development/lauda_measurement_napi/node_modules/node-addon-api/napi.h:2347:9: note: ‘class Napi::TypedThreadSafeFunction<ContextType, DataType, CallJs>’ defined here
 2347 |   class TypedThreadSafeFunction {​​​​​​​
[...]

PR #1008, opened 2021-06-11
doc: add g++11 incompatibility with c++20

KevinEady:
Closes: #1007


KevinEady:
This can actually be corrected in the code, as seen in the gcc not-a-bug [bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101032)

PR #1009, opened 2021-06-11
src: fix gcc-11 c++20 compilation

KevinEady:
As mentioned in [this comment](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101032#c2), the as-is code is not valid C++20.  This fix is valid in all C++ standards.

Closes: #1007


mhdawson (approved):
LGTM

mhdawson:
Landed as fed13534c595c52fdd2cbf64d7bffe83de74ef35

Issue #1011, opened 2021-06-16
"Class property and descriptor" example crashes in Electron render process on reload


ToadKing:
PoC Project: https://github.com/ToadKing/electron-native-module-poc (The native module part is an exact copy of the example in the node-addon-api docs.)

The [class example](https://github.com/nodejs/node-addon-api/blob/main/doc/class_property_descriptor.md) in the docs makes the constructor reference persistent using `Napi::Persistent` on module initialization. However, in Electron when a render process reloads a page, the native module is re-initialized without reloading the shared library itself. When a new constructor reference is loaded into the `constructor` variable it tries to delete the former reference, which causes a crash of the Electron render process.

KevinEady:
Hi @ToadKing ,

Thanks for pointing this out. The example in the documentation predates some modern APIs for handling context-sensitivity, particularly [`SetInstanceData`](https://github.com/nodejs/node-addon-api/blob/main/doc/env.md#setinstancedata-1) and [`GetInstanceData`](https://github.com/nodejs/node-addon-api/blob/main/doc/env.md#getinstancedata). You can update https://github.com/ToadKing/electron-native-module-poc/blob/76d02d8c7cb151f3a657734836012629fc2f4f8c/native/src/binding.cc#L23-L24 with:


```cpp
    Napi::FunctionReference *constructor = new Napi::FunctionReference();
    *constructor = Napi::Persistent(func);
    env.SetInstanceData(constructor);
```


https://user-images.githubusercontent.com/8634912/122254029-060c0280-cecd-11eb-9e29-d9fcd3a7cded.mp4

[NB: SetInstanceData will overwrite any existing instance data set. See https://github.com/nodejs/node-addon-api/issues/744#issuecomment-859193621 for some examples on how to store multiple constructors on the instance data.]

ToadKing:
Thank you, this change fixes my PoC app and the app I found it in. Should this issue be left open and be changed to update the documentation or is it good to close?

KevinEady:
@ToadKing I am not sure... We will leave it open to discuss at this week's Node-API meeting. We'll need to discuss a hopefully context-sensitive approach for Node-API version <= 5. I will leave this issue open until we discuss it.

Thanks, Kevin

PR #1013, opened 2021-06-21
doc: update examples for context sensitivity

KevinEady:
Closes: https://github.com/nodejs/node-addon-api/issues/1011

### DIFF
--- a/doc/class_property_descriptor.md
+++ b/doc/class_property_descriptor.md
@@ -17,7 +17,6 @@ class Example : public Napi::ObjectWrap<Example> {
     Example(const Napi::CallbackInfo &info);
 
   private:
-    static Napi::FunctionReference constructor;
     double _value;
     Napi::Value GetValue(const Napi::CallbackInfo &info);
     void SetValue(const Napi::CallbackInfo &info, const Napi::Value &value);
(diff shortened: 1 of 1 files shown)


doc/class_property_descriptor.md
```diff
@@ -31,8 +30,9 @@ Napi::Object Example::Init(Napi::Env env, Napi::Object exports) {
         InstanceAccessor<&Example::GetValue>("readOnlyProp")
     });
 
-    constructor = Napi::Persistent(func);
-    constructor.SuppressDestruct();
+    Napi::FunctionReference *constructor = new Napi::FunctionReference();
+    *constructor = Napi::Persistent(func);
+    env.SetInstanceData(constructor);
```
legendecas, line 35: A finalizer is required for this FunctionReference pointer right?

KevinEady: Hi @legendecas ,

Good question... I was under the impression that the instance data would be `delete`d, so there would be no need for a Finalizer.
- In the case of multiple calls to `Example::Init`, the previous instance data would be overwritten via most recent call to `env.SetInstanceData`. https://github.com/nodejs/node/blob/d615aeb7583b15bb5a8d1ec666ea29b8c7377455/src/js_native_api_v8.cc#L3201-L3205 does this delete the data?
- The final data would be `delete`d on environment cleanup.

Is my understanding correct?

legendecas: Eh, IIUC, it's node-addon-api's default finalizer deleted the data pointer: https://github.com/nodejs/node-addon-api/blob/main/napi.h#L208. Not quite intuitive tho... So no finalizer is required here.

legendecas: The `Delete` you point out only deleting the `RefBase` structure allocated in Node.js core, which would invoke the finalizer provided by node-addon-api.

legendecas (approved):

mhdawson (approved):
LGTM

mhdawson:
Landed as 627dbf3c37ab64e33ad4431c6b569f6a36dd1723

PR #1014, opened 2021-06-25
src: add Env::AddCleanupHook

KevinEady:
Adds wrapping for `napi_add_env_cleanup_hook` and `napi_remove_env_cleanup_hook`.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -5692,6 +5692,64 @@ Addon<T>::DefineProperties(Object object,
 }
 #endif  // NAPI_VERSION > 5
 
+template <typename Hook, typename Hint>
+void EnvHookCleanupData<Hook, Hint>::Wrapper(void* data) NAPI_NOEXCEPT {
+  EnvHookCleanupData* cleanupData = static_cast<EnvHookCleanupData*>(data);
+  cleanupData->callback();
+  delete cleanupData;
+}
+
+template <typename Hook, typename Hint>
+void EnvHookCleanupData<Hook, Hint>::WrapperWithHint(void* data) NAPI_NOEXCEPT {
+  EnvHookCleanupData* cleanupData = static_cast<EnvHookCleanupData*>(data);
+  cleanupData->callback(static_cast<Hint*>(cleanupData->hint));
+  delete cleanupData;
+}
+
+template <typename Hook, typename Hint>
+EnvCleanupHook<Hook, Hint>::EnvCleanupHook(void (*wrapper)(void* arg),
+                                           EnvHookCleanupData<Hook, Hint>* data)
+    : wrapper(wrapper), data(data) {}
+
+template <typename Hook, typename Hint>
+void EnvCleanupHook<Hook, Hint>::Remove(Napi::Env env) {
+  napi_status status = napi_remove_env_cleanup_hook(env, wrapper, data);
+  NAPI_FATAL_IF_FAILED(status,
+                       "EnvCleanupHook<Hook, Hint>::Remove",
+                       "napi_remove_env_cleanup_hook");
+  delete data;
+}
+
+template <typename Hook, typename Hint>
+EnvCleanupHook<Hook, Hint> AddCleanupHook(Napi::Env env,
+                                          Hook finalizeCallback,
+                                          Hint* arg) {
+  EnvHookCleanupData<Hook, Hint>* cleanupData =
+      new EnvHookCleanupData<Hook, Hint>({std::move(finalizeCallback), arg});
+
+  napi_status status = napi_add_env_cleanup_hook(
+      env, EnvHookCleanupData<Hook, Hint>::WrapperWithHint, cleanupData);
+  NAPI_FATAL_IF_FAILED(status,
+                       "EnvCleanupHook<Hook, Hint>::AddCleanupHook",
+                       "napi_add_env_cleanup_hook");
+  return EnvCleanupHook<Hook, Hint>(
+      EnvHookCleanupData<Hook, Hint>::WrapperWithHint, cleanupData);
+}
+
+template <typename Hook>
+EnvCleanupHook<Hook> AddCleanupHook(Napi::Env env, Hook finalizeCallback) {
+  EnvHookCleanupData<Hook>* cleanupData =
+      new EnvHookCleanupData<Hook>({std::move(finalizeCallback), nullptr});
+
+  napi_status status = napi_add_env_cleanup_hook(
+      env, EnvHookCleanupData<Hook>::Wrapper, cleanupData);
+  NAPI_FATAL_IF_FAILED(status,
+                       "EnvCleanupHook<Hook>::AddCleanupHook",
+                       "napi_add_env_cleanup_hook");
+
+  return EnvCleanupHook<Hook>(EnvHookCleanupData<Hook>::Wrapper, cleanupData);
+}
+
 } // namespace Napi
 
 #endif // SRC_NAPI_INL_H_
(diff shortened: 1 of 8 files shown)


napi.h
```diff
@@ -221,6 +229,21 @@ namespace Napi {
     napi_env _env;
   };
 
+  template <typename Hook, typename Hint>
+  class CleanupHook {
+   public:
```
KevinEady, line 234: Some way to make this private...?

napi.h
```diff
@@ -221,6 +229,21 @@ namespace Napi {
...
+    struct CleanupData {
```
KevinEady, line 235: Some way to make this private...?

KevinEady:
Hi team. This is ready for review. I have fixed the API privatization, see https://github.com/nodejs/node-addon-api/pull/1014/files#diff-6294f8ee1d8cd7e4e94cfc22a083380fb80058e34f53100d895b9b3b295dd801 

- `Env::CleanupHook` is a private class in `Env` with public constructors, so only `Env` can construct an instance and eg. return it to the caller in `Env::AddCleanupHook`.
- `Env::CleanupHook` exposes one public method, `Remove`.
- `Env::CleanupHook` has the private the static wrapper methods (one for calling `hook()` and one for `hook(arg)`). They are referenced inside the constructor.

napi.h
```diff
@@ -219,6 +227,24 @@ namespace Napi {
 
   private:
     napi_env _env;
+
+    template <typename Hook, typename Arg>
+    class CleanupHook {
+     public:
+      CleanupHook(Env env, Hook hook, Arg* arg);
```
mhdawson, line 234: I think these need to be guarded by NAPI_VERSION > 3 as the cleanup hooks were only added in version 3 as per: https://nodejs.org/api/n-api.html#n_api_napi_add_env_cleanup_hook

napi.h
```diff
@@ -201,6 +203,12 @@ namespace Napi {
     Value RunScript(const std::string& utf8script);
     Value RunScript(String script);
 
+    template <typename Hook>
+    CleanupHook<Hook> AddCleanupHook(Hook hook);
+
+    template <typename Hook, typename Arg>
+    CleanupHook<Hook, Arg> AddCleanupHook(Hook hook, Arg* arg);
+
```
mhdawson, line 211: I think these need to be guarded by NAPI_VERSION > 3 as the cleanup hooks were only added in version 3 as per: https://nodejs.org/api/n-api.html#n_api_napi_add_env_cleanup_hook

napi-inl.h
```diff
@@ -5692,6 +5709,46 @@ Addon<T>::DefineProperties(Object object,
...
+Env::CleanupHook<Hook> Env::AddCleanupHook(Hook hook) {
+  return CleanupHook<Hook>(*this, hook);
+}
+
+template <typename Hook, typename Arg>
+Env::CleanupHook<Hook, Arg>::CleanupHook(Napi::Env env, Hook hook)
+    : wrapper(Env::CleanupHook<Hook, Arg>::Wrapper) {
+  data = new CleanupData{std::move(hook), nullptr};
+  napi_status status = napi_add_env_cleanup_hook(env, wrapper, data);
+  NAPI_FATAL_IF_FAILED(status,
```
mhdawson, line 5727: What would cause napi_add_env_cleanup_hook to fail ?   I'm wondering if we can let the caller trying add a hook know it failed versus causing a FATAL error.

napi.h
KevinEady: I thought about this, but we dropped support for node 8, and Node-API version is present in 10+. Do we still need a guard?

napi-inl.h
KevinEady: Looking at the implementation, it looks like it should always return `napi_ok` ... https://github.com/nodejs/node/blob/663d7e9fb27f017f38b79382a95fec10e5f66f9a/src/node_api.cc#L656-L676

According to the docs it can just abort the process with invalid arguments: 
> Providing the same fun and arg values multiple times is not allowed and will lead the process to abort.

But we would never provide the same `arg` value (as we always create a `new CallbackData`)

Maybe we don't need the check at all...? However `napi_add_env_cleanup_hook` always returning `napi_ok` is an implementation detail, maybe we should keep it blackboxed 🤷 

How would you recommend "add a hook know it failed" ...?

Thanks, Kevin

napi.h
mhdawson: I think so, people can still ask for a particular Node-API version and I think we should respect that.

napi-inl.h
mhdawson: How would you recommend "add a hook know it failed" . What I had in mind was some kind of return value the caller could check. For example if it was documented that the callers should check that the result from AddCleanupHook was non-null would that make any sense?

napi.h
KevinEady: Addressed.

napi-inl.h
KevinEady: Addressed. Added a `bool IsEmpty()` method on `Env::CleanupHook` https://github.com/nodejs/node-addon-api/blob/36ef600b196f1876cf3390610960488e2dae9be5/napi-inl.h#L5756-L5759

`data` is nullptr if adding the hook failed:
https://github.com/nodejs/node-addon-api/blob/36ef600b196f1876cf3390610960488e2dae9be5/napi-inl.h#L5737-L5746

napi.h
KevinEady: Note I did not add guards to the tests -- we will always test Node-API 3+ right?

napi-inl.h
```diff
@@ -5692,6 +5712,53 @@ Addon<T>::DefineProperties(Object object,
...
+
+template <typename Hook>
+Env::CleanupHook<Hook> Env::AddCleanupHook(Hook hook) {
+  return CleanupHook<Hook>(*this, hook);
+}
+
+template <typename Hook, typename Arg>
+Env::CleanupHook<Hook, Arg>::CleanupHook(Napi::Env env, Hook hook)
+    : wrapper(Env::CleanupHook<Hook, Arg>::Wrapper) {
+  data = new CleanupData{std::move(hook), nullptr};
```
mhdawson, line 5729: Could this version just call the other one and pass nullptr for data ?

napi.h
mhdawson: In terms of guarding the tests its only a 1 line addition here: https://github.com/nodejs/node-addon-api/blob/6697c51d1d08ced8d1c1c40715ceea002613fa6b/test/index.js#L84 so I think we might as well.

mhdawson: Thinking more I guess there is more to it than that. You also need to guard the compilation but that is pretty easy too. Unlikely anybody will test with napiVersion <3 but for consistency I'd still prefer if we guard since it's not a lot of effort.

napi-inl.h
KevinEady: @mhdawson no because each constructor initializes the instance member `wrapper` to a different value, `Env::CleanupHook<Hook, Arg>::Wrapper` (this one) or `Env::CleanupHook<Hook, Arg>::WrapperWithArg`

napi.h
KevinEady: Addressed.

napi-inl.h
mhdawson: Thanks, nervermind.

mhdawson (approved):
LGTM

mhdawson:
Landed as 10564a43c62d4b9a72c28e1b0d35c032e0aa8990

PR #1015, opened 2021-06-28
Add error checking in String::New function

JckXia:
Added an if statement to String::New(env, const char * str) to make sure that str pointer is not null

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -901,8 +901,9 @@ inline String String::New(napi_env env, const std::u16string& val) {
 
 inline String String::New(napi_env env, const char* val) {
   if (val == nullptr) {
-    NAPI_THROW(TypeError::New(env, "A null pointer has been passed in"),
-               Napi::String());
+    NAPI_THROW(
+        TypeError::New(env, "String::New recieved a nullpointer as a value"),
+        Napi::String());
   }
   napi_value value;
   napi_status status = napi_create_string_utf8(env, val, std::strlen(val), &value);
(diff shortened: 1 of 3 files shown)


JckXia:
Should I also add this if statement to the other` String::New `overloads that accept a C char pointer? Also  for the overload `String::New(env, const char *, length)`, it doesn't seem like there are any checks in place to make sure that the length param is indeed the length of the string passed in. If I pass a value that exceeds the length of the string ( in this case String::New(env, "abcde",46)), I get this result: 

stringLengthCheck
Is this expected behavior? Thank you! 
To reproduce the above result, I defined a C++ function returning String::New(env, "abcde",46). From the JS side, I printed out the return value of this Cpp function using `console.log()`

test/name.js
```diff
@@ -6,7 +6,13 @@ module.exports = require('./common').runTest(test);
...
 function test(binding) {
   const expected = '123456789';
-
+  
+  try{
+    binding.name.nullStringShouldThrow();
+  }catch(e)
+  {
+    assert.strictEqual(e.message, 'String::New recieved a nullpointer as a value');
+  }
```
KevinEady, line 15: You can use `assert.throws()`

napi-inl.h
```diff
@@ -900,6 +900,11 @@ inline String String::New(napi_env env, const std::u16string& val) {
 }
 
 inline String String::New(napi_env env, const char* val) {
+  if (val == nullptr) {
+    NAPI_THROW(
+        TypeError::New(env, "String::New recieved a nullpointer as a value"),
```
KevinEady, line 905: typo 😉

KevinEady (changes requested):

KevinEady (approved):
LGTM

mhdawson (approved):
LGTM

mhdawson:
Landed in da50b51398789185ab63ece534890f131703ffbb

Issue #1022, opened 2021-07-14
Intermittent crashes in CI


mhdawson:
There have been a few recent crashes (maybe 3 in last 15 or so days) in the CI  https://ci.nodejs.org/view/x%20-%20Abi%20stable%20module%20API/job/node-test-node-addon-api-LTS%20versions/

A number of the failures look like the following just before the crash

```bash
Running test 'objectwrap_constructor_exception'
Running test 'objectwrap_multiple_inheritance'
Running test 'objectwrap_worker_thread'
Running test 'promise'
Running test 'reference'
Running test 'run_script'
Running test 'symbol'
Running test 'threadsafe_function/threadsafe_function'
Running test 'threadsafe_function/threadsafe_function_ctx'
Tests aborted with SIGSEGV
napiVersion:8
Testing with Node-API Version '8'.
Starting test suite
```

mhdawson:
I think I've managed to recreate a similar instance and the stack trace looks like:

```shell
sing host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/home/midawson/newpull/land/node/out/Release/node --expose-gc --no-concurrent-a'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000e9ffe1 in v8::internal::GlobalHandles::Destroy(unsigned long*) ()
[Current thread is 1 (Thread 0x7fda22e537c0 (LWP 2218825))]
Missing separate debuginfos, use: yum debuginfo-install glibc-2.28-127.el8.x86_64 libgcc-8.3.1-5.1.el8.x86_64 libstdc++-8.3.1-5.1.el8.x86_64
(gdb) bt
#0  0x0000000000e9ffe1 in v8::internal::GlobalHandles::Destroy(unsigned long*) ()
#1  0x0000000000ad98dd in napi_resolve_deferred ()
#2  0x00007fda20085ebd in Napi::Promise::Deferred::Resolve (this=0x5935dc0, value=0x5945090) at ../../napi-inl.h:2170
#3  0x00007fda200b272c in (anonymous namespace)::TSFNWrap::<lambda(Napi::Env, Napi::Reference<Napi::Value>*)>::operator()(Napi::Env, Napi::Reference<Napi::Value> *) const (__closure=0x5939dc8, env=..., ctx=0x5a09dc0)
    at ../threadsafe_function/threadsafe_function_ctx.cc:51
[...]

PR #1024, opened 2021-07-20
doc: update tests to avoid running in parallel

mhdawson:
fixes: https://github.com/nodejs/node-addon-api/issues/1022

The objectwrap_worker_thread and symbol tests
were not waiting for the test to complete before the
subsequent tests were started. This caused intermittent
crashes in the CI.

Updated both tests so that they complete before the
next test runs.

### DIFF
--- a/test/common/index.js
+++ b/test/common/index.js
@@ -102,7 +102,7 @@ exports.runTestWithBindingPath = async function(test, buildType) {
   }
 }
 
-exports.runTestNoBinding = async function(test, buildType) {
+exports.runTestWithBuildType = async function(test, buildType) {
   buildType = buildType || process.config.target_defaults.default_configuration || 'Release';
 
    await Promise.resolve(test(buildType))
--- a/test/common/index.js
+++ b/test/common/index.js
@@ -101,3 +101,10 @@ exports.runTestWithBindingPath = async function(test, buildType) {
     await test(item);
   }
 }
+
+exports.runTestNoBinding = async function(test, buildType) {
+  buildType = buildType || process.config.target_defaults.default_configuration || 'Release';
+
+   await Promise.resolve(test(buildType))
+     .finally(exports.mustCall());
+}
--- a/test/objectwrap_worker_thread.js
+++ b/test/objectwrap_worker_thread.js
@@ -1,14 +1,19 @@
 'use strict';
+const path = require('path');
 const { Worker, isMainThread, workerData } = require('worker_threads');
 
-if (isMainThread) {
-    const buildType = process.config.target_defaults.default_configuration;
-    new Worker(__filename, { workerData: buildType });
-} else {
-    const test = binding => {
-        new binding.objectwrap.Test();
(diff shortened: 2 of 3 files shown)


mhdawson:
@nodejs/node-api  anybody have a windows machine who can check what's up with the symbol test on windows?  I'm not sure how the change would have broken just for windows?

NickNaso:
Hi @mhdawson,
it seems that when we call the function `assertSymbolIsWellknown` (https://github.com/nodejs/node-addon-api/blob/main/test/symbol.js#L30) for the  symbol `matchAll` we get `undefined` and the test fail.
If I do the same in native JavaScript all works well:
```js
'use strict'

const assert = require('assert')

const symbOne = Symbol.for('matchAll');
const symbTwo = Symbol.for('matchAll');

assert(symbOne && symbTwo);
assert(symbOne === symbTwo);
```

mhdawson:
@NickNaso thanks! Strange that it is found on linux but not windows.

test/common/index.js
```diff
@@ -101,3 +101,10 @@ exports.runTestWithBindingPath = async function(test, buildType) {
     await test(item);
   }
 }
+
+exports.runTestNoBinding = async function(test, buildType) {
```
legendecas, line 105: The above two helpers are named with their parameters:

- `runTest(test: (binding) => Promise<void>)`
-  `runTestWithBindingPath(test: (bindingPath) => Promise<void>)`

Maybe we can name this as `runTestWithBuildType` so that we can assume the first parameter of the test function is `buildType`?

mhdawson:
@NickNaso  this is what I see on linux

```shell
[midawson@drx-hemera node]$ cat test.js
console.log(global.Symbol.matchAll);
console.log(global.Symbol['matchAll']);
[midawson@drx-hemera node]$ ./node test.js
Symbol(Symbol.matchAll)
Symbol(Symbol.matchAll)
```

From what I understand you get undefined on windows instead, right ?

test/common/index.js
mhdawson: @legendecas that makes sense to me will update.

mhdawson:
@NickNaso, I thinkI understand you now, sounds like we should only test for that symbol on 12.x and above

NickNaso:
> @NickNaso this is what I see on linux
> 
> ```shell
> [midawson@drx-hemera node]$ cat test.js
> console.log(global.Symbol.matchAll);
> console.log(global.Symbol['matchAll']);
> [midawson@drx-hemera node]$ ./node test.js
> Symbol(Symbol.matchAll)
> Symbol(Symbol.matchAll)
> ```
> 
> From what I understand you get undefined on windows instead, right ?

Yes on Windows, but only with Node.js 10.x I get `undefined`.

legendecas (approved):

mhdawson:
Landed as a459f5cc8f68065dd3aaed2a45f775859af36e43

PR #1026, opened 2021-07-23
Add Function::Call Napi::Value override

rgerd:
Adds an override to Napi::Function::Call so that you can call it with a c-style array of Napi::Value's

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2287,6 +2287,11 @@ inline MaybeOrValue<Value> Function::Call(
   return Call(Env().Undefined(), args);
 }
 
+inline MaybeOrValue<Value> Function::Call(
+    const std::vector<Value>& args) const {
+  return Call(Env().Undefined(), args);
+}
+
 inline MaybeOrValue<Value> Function::Call(size_t argc,
                                           const napi_value* args) const {
   return Call(Env().Undefined(), argc, args);
@@ -2300,6 +2305,27 @@ inline MaybeOrValue<Value> Function::Call(
 inline MaybeOrValue<Value> Function::Call(
     napi_value recv, const std::vector<napi_value>& args) const {
   return Call(recv, args.size(), args.data());
+}
+
+inline MaybeOrValue<Value> Function::Call(
+    napi_value recv, const std::vector<Value>& args) const {
+  const size_t argc = args.size();
+  const size_t stackArgsCount = 6;
+  napi_value stackArgs[stackArgsCount];
+  std::vector<napi_value> heapArgs;
+  napi_value* argv;
+  if (argc <= stackArgsCount) {
+    argv = stackArgs;
+  } else {
+    heapArgs.resize(argc);
+    argv = heapArgs.data();
+  }
+
+  for (size_t index = 0; index < argc; index++) {
+    argv[index] = static_cast<napi_value>(args[index]);
+  }
+
+  return Call(recv, argc, argv);
 }
 
 inline MaybeOrValue<Value> Function::Call(napi_value recv,
--- a/napi.h
+++ b/napi.h
@@ -1350,11 +1350,14 @@ namespace Napi {
    MaybeOrValue<Value> Call(
        const std::initializer_list<napi_value>& args) const;
    MaybeOrValue<Value> Call(const std::vector<napi_value>& args) const;
+   MaybeOrValue<Value> Call(const std::vector<Value>& args) const;
    MaybeOrValue<Value> Call(size_t argc, const napi_value* args) const;
    MaybeOrValue<Value> Call(
        napi_value recv, const std::initializer_list<napi_value>& args) const;
    MaybeOrValue<Value> Call(napi_value recv,
                             const std::vector<napi_value>& args) const;
+   MaybeOrValue<Value> Call(napi_value recv,
+                            const std::vector<Value>& args) const;
    MaybeOrValue<Value> Call(napi_value recv,
                             size_t argc,
                             const napi_value* args) const;
--- a/test/function.cc
+++ b/test/function.cc
@@ -69,6 +69,16 @@ Value CallWithVector(const CallbackInfo& info) {
    return MaybeUnwrap(func.Call(args));
 }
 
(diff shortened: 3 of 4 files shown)


mhdawson:
Note from discussion in Node-api team meeting today, we should validate this does not add any performance overhead to the existing cases.

napi-inl.h
```diff
@@ -277,7 +277,7 @@ template <typename ContextType, typename DataType, typename CallJs, CallJs call>
 typename std::enable_if<call == nullptr>::type static inline CallJsWrapper(
     napi_env env, napi_value jsCallback, void* /*context*/, void* /*data*/) {
   if (jsCallback != nullptr) {
-    Function(env, jsCallback).Call(0, nullptr);
+    Function(env, jsCallback).Call(0, static_cast<const napi_value*>(nullptr));
```
gabrielschulhof, line 280: Why is this change needed?

rgerd: The compiler flagged it as an ambiguous case with the new overload, indicating the nullptr's type fixed the error.

rgerd:
> Note from discussion in Node-api team meeting today, we should validate this does not add any performance overhead to the existing cases.

@mhdawson Is there something I should do on my end to validate performance?

mhdawson:
@rgerd not at this point it was more of a reminder to reviewers of the PR

mhdawson:
@gabrielschulhof wondering if you have more review questions/comments?

gabrielschulhof (approved):

mhdawson:
@rgerd thanks for your patience, discussed again in the Node-API team meeting. We agreed this can land once rebased to address the conflicts.

One additional suggestion was that it would make sense to add another overload to accept a std:vector of Value object either in this or a follow on PR.

mhdawson (approved):
LGTM

napi-inl.h
legendecas: Does this mean that user might need to change their code if they are already using this pattern?

legendecas:
Commented with a question about compatibility issue https://github.com/nodejs/node-addon-api/pull/1026#discussion_r713564873. We may need to come to a conclusion about the usage pattern compatibility until we can proceed.

gabrielschulhof:
In today's Node-API team meeting @vmoroz brought up a good point, namely that it may not be a good idea to have an override passing a C-style array of `napi_value` items because having a well-formed array description is a safer choice. For example, C++20 will have `std::span<T>`.

He proposed that we create a simple, say, `Napi::ValueSpan` class which will be constructed with `napi_value` and `size_t` and create an override that accepts such an instance.

```C++
my_function.Call(Napi::ValueSpan(count, values));
```

We can use C++ templating to create a constructor override that can deduce the length of the array without having to pass it in explicitly as a parameter. We should structure the class definition to match that of `std::span<T>` exactly, so that, once we start using C++20, we can simply convert the definition to an alias of `std::span<T>`.

mhdawson:
We discussed again in today's Node-API team meeting. The conclusion we came to was that it would be better for signature of the new method to use std::vector. For example:

```C++
Value Call(napi_value recv, const std::vector<Value>& args) ;
```

It is possible to get a Value* array into a std::vector in a few ways (assign, constructor etc.)

@rgerd is this something you'd still like to move forward?

rgerd:
Hi, thank you very much for the feedback. Yes, I'd still like to move it forward, sorry for disappearing. I should be able to make the changes and validate within the next week or so.

rgerd:
Rebased with the suggested changes. The style checks are failing on `function.js` for parts of the file unrelated to my changes, but I also noticed eslint was added after `function.js` was last touched, so it seems like it hasn't been linted by the CI until now.

I could fix the linting problems myself, but just wanted to check if you have any opinions on introducing a bunch of unrelated diffs in this PR, or if we should open another PR to go in first to address those style changes.

mhdawson:
@rgerd its a good question about the linter changes.  I think in this case what might makes sense is to have 2 commits in the PR, one for the linter changes and one for the change your are making.  When we land we can keep both commits instead of squashing. That seems better than using 2 PR to accomplish the same thing.

mhdawson:
@NickNaso will take a look/test on ci and then will go ahead and land.

NickNaso:
First pass on CI:
| Version | Job | Status |
|---:|---|:---:||
| main | https://ci.nodejs.org/job/node-test-node-addon-api-new/5358/ |  |
| v17.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5359/ |  |
| v16.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5360/ |  |
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5361/ |  |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5362/ |  |

NickNaso:
Second pass on CI:
| Version | Job | Status |
|---:|---|:---:|
| main | https://ci.nodejs.org/job/node-test-node-addon-api-new/5363/ |   |
| v17.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5364/ |   |
| v16.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5365/ |   |
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5366/ |   |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5367/ |   |

NickNaso:
Third pass on CI:
| Version | Job | Status |
|---:|---|:---:|
| main | https://ci.nodejs.org/job/node-test-node-addon-api-new/5369/ | ✓ |
| v17.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5370/ | ✓ |
| v16.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5371/ | ✓ |
| v14.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5372/ | ✓ |
| v12.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/5373/ | ✓ |

PR #1051, opened 2021-08-25
[WIP] Add unit test helper to asyncworker

JckXia:
Add a test class to aid testing different async worker overloads


napi-inl.h
```diff
@@ -3232,6 +3232,19 @@ inline MaybeOrValue<Object> FunctionReference::New(
...
+////////////////////////////////////////////////////////////////////////////////
+// AsyncWorkerUnitTest
+////////////////////////////////////////////////////////////////////////////////
+
+#if defined(IS_TESTING)
+inline AsyncWorkerUnitTest::AsyncWorkerUnitTest(AsyncWorker& worker)
+    : _worker(worker) {}
+
+inline bool AsyncWorkerUnitTest::isSuppressDest() {
+  return this->_worker._suppress_destruct;
```
legendecas, line 3244: Maybe we should test this with asan build rather than checking it explicitly?

Issue #85, opened 2017-07-20
MSVC needs -D _HAS_EXCEPTIONS=1 for exceptions to work properly


mika-fischer:
This should be added to the instructions in the README.

```
'conditions': [
    ['OS=="win"', { 'defines': [ '_HAS_EXCEPTIONS=1' ] }]
]
```

jasongin:
Why? Can you explain what problem you're seeing without that define? There are [tests ](https://github.com/nodejs/node-addon-api/blob/master/test/error.cc) of the exception behavior that seem to work fine without it.

As far as I know, `_HAS_EXCEPTIONS=1` is the default for MSVC. Some projects may override it to 0. But in that case, they'd probably want `'NAPI_DISABLE_CPP_EXCEPTIONS` instead.

mika-fischer:
Thing work to some degree without it but there are subtle memory errors. I don't understand exactly what this flag causes but the effect is that objects are freed to early when handling an exception.

For instance try this with `_HAS_EXCEPTIONS=0` and `_HAS_EXCEPTIONS=1`:
```
try
{
    throw std::runtime_error(std::string("abc"));
}
catch (std::exception& e)
{
    std::cout << "'" << e.what() << "'" << std::endl;
}
```

For me with MSVC 2015 this reliably prints `''` with `_HAS_EXCEPTIONS=0`. But I've also encountered garbage memory in other cases. Needless to say this was a huge pain to debug and thus I think it should be much more prominently documented.

The define comes from nodejs itself: https://github.com/nodejs/node/blob/master/common.gypi#L273

See also https://github.com/nodejs/node-gyp/issues/939 which led me to the solution.

jasongin:
Interesting. Thanks for the details. And I didn't realize node `common.gypi` set `_HAS_EXCEPTIONS=0`. I agree then it should be added to the README.

DBJDBJ:
### Abstract
[The author measures size costs of error handling approaches. Exceptions are big, `std::abort` and integer error codes are small.](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1640r1.html)

NickNaso:
Hi @DBJDBJ,
thanks for reporting the paper about performance on error handling. `node-addon-api` allow you to enable or disable the exception mechanism. 
It is documented on the following sections:
- **[Setup](https://github.com/nodejs/node-addon-api/blob/master/doc/setup.md)**
[...]

PR #1059, opened 2021-09-01
doc: fix documentation about how to enable C++ exception

NickNaso:
This PR fixes the documentation about how to enbale C++ exception. The problem was identified the first time in this issue: https://github.com/nodejs/node-addon-api/issues/85. In Windows system is not enough add the following configuration on `binding.gyp`:
```
'msvs_settings': {
    'VCCLCompilerTool': { 'ExceptionHandling': 1 },
 },
```
but it's necessary override the defined `_HAS_EXCEPTIONS` macro and set it to 1. 
In the `common.gypi` (see:  https://github.com/nodejs/node/blob/master/common.gypi#L273) the `_HAS_EXCEPTIONS` has been set to 0 to make sure the STL doesn't try to use exceptions, but in the addons with the support for C++ it needs to be changed to 1 otherwise the exceptions system could be catch the wrong exception.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -44,7 +44,7 @@ To use **Node-API** in a native module:
       "defines": [
         "_HAS_EXCEPTIONS=1"
       ],
-        "msvs_settings": {
+      "msvs_settings": {
         "VCCLCompilerTool": {
           "ExceptionHandling": 1
         },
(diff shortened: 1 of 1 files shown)


mhdawson (approved):
LGTM

PR #1075, opened 2021-09-21
Add logic to handle graceful error handling when primitive errors are thrown by JS functions

JckXia:
Added some logic to override Value() function on the Error object to do the unwrapping, as well wrapping logic to the Object(env, value) function accordingly.


doc/error_handling.md
```diff
@@ -14,6 +14,10 @@ If C++ exceptions are enabled (for more info see: [Setup](setup.md)), then the
 `Napi::Error` class extends `std::exception` and enables integrated
 error-handling for C++ exceptions and JavaScript exceptions.
 
+Note, that due to limitations of the N-API, if one attempt to cast the error object thrown as a primitive, an
+wrapped object will be received instead. (With properties ```isWrapObject``` and ```errorVal``` containing the primitive value thrown)
```
mhdawson, line 18: Maybe just add some additional clarification that this is when the node-addon-api Error C++ object is used within the addon.

napi-inl.h
```diff
@@ -2498,12 +2498,61 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+      status = napi_create_reference(env, wrappedErrorObj, 1, &_ref);
+    }
+
     // Avoid infinite recursion in the failure case.
     // Don't try to construct & throw another Error instance.
     NAPI_FATAL_IF_FAILED(status, "Error::Error", "napi_create_reference");
   }
 }
 
+inline Object Error::Value() const {
```
mhdawson, line 2530: I think it might make sense for this to return Value instead of Object (Values can be objects), the main concern is that would potentially be breaking. 

The reason is that we wrapping so that we can have a value versus just objects.  If its possible we might want two methods one which returns a Value and one which returns an Object in order to avoid breaking existing code.

JckXia: I tried setting the return type to Value and it seems to pass all the test suites.  Although I am not too sure if that breaks any conventions. (i.e, In the base Reference<T> class, ```Reference<T>::Value()``` will return an instance of type T, and Error is of type ```Object``` since it inherits ```ObjectReference```. Therefore people might expect Error::Value() to return an ```Object``` also) 

For the second part then, can I declare two public methods ```Napi::Value GetUnWrappedValue()``` and ```Object Value()``` ?

mhdawson: If is possible to have 2 methods both called Value()  which return different types. If so that might make sense assuming the compiler will select the most appropriate one based on how you use the return value?

JckXia: Hmm, I am not too sure if it's possible to overload base on return value in C++ (without using templates) according to this post (https://stackoverflow.com/questions/9568852/overloading-by-return-type).

mhdawson: After discussing in Node-API team meeting I think this is ok as  is.

mhdawson (approved):
LGTM

doc/error_handling.md
```diff
@@ -14,6 +14,10 @@ If C++ exceptions are enabled (for more info see: [Setup](setup.md)), then the
 `Napi::Error` class extends `std::exception` and enables integrated
 error-handling for C++ exceptions and JavaScript exceptions.
 
+Note, that due to limitations of the N-API, if one attempts to cast the error object wrapping a primitive inside a C++ addon, the wrapped object
+will be received instead. (With properties ```4b3d96fd-fb87-4951-a979-eb4f9d2f2ce9-isWrapObject``` and ```errorVal``` containing the primitive value thrown)
```
NickNaso, line 18: It's inline code so please use one back tick  es. `4b3d96fd-fb87-4951-a979-eb4f9d2f2ce9-isWrapObject`

NickNaso (commented):
Just some notes about documentation.

doc/error_handling.md
```diff
@@ -14,6 +14,10 @@ If C++ exceptions are enabled (for more info see: [Setup](setup.md)), then the
...
+
+
```
gabrielschulhof, line 20: We don't need this extra blank line.

napi-inl.h
```diff
@@ -2498,12 +2498,66 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+  if (_ref == nullptr) {
+    return Object(_env, nullptr);
+  }
```
gabrielschulhof, line 2535: When would this ever be the case?

napi-inl.h
```diff
@@ -2498,12 +2498,66 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+
+  napi_value refValue;
+  napi_status status = napi_get_reference_value(_env, _ref, &refValue);
+  NAPI_THROW_IF_FAILED(_env, status, Object());
+
+  // We are checking if the object is wrapped
+  bool isWrappedObject = false;
+  napi_has_property(
```
gabrielschulhof, line 2543: Maybe add a typecheck first to see that it's not a Symbol. If the reference is a Symbol it's definitely not wrapped. In new Node.js versions, the reference can be a symbol.

napi.h
```diff
@@ -8,7 +8,6 @@
 #include <mutex>
 #include <string>
 #include <vector>
-
```
gabrielschulhof, line 11: This line should remain.

test/binding.cc
```diff
@@ -31,6 +31,7 @@ Object InitDate(Env env);
 Object InitDataView(Env env);
 Object InitDataViewReadWrite(Env env);
 Object InitEnvCleanup(Env env);
+Object InitErrorHandlingPrim(Env env);
```
gabrielschulhof, line 34: Are these names consistent with the file names being used? This will help @deepakrkris with the unit test filtering.

gabrielschulhof: @deepakrkris can you double-check this please?

test/errorHandlingForPrimitives.cc
```diff
@@ -0,0 +1,13 @@
...
+void Test(const Napi::CallbackInfo& info) {
+  info[0].As<Napi::Function>().Call({});
+}
+
+}  // namespace
+Napi::Object InitErrorHandlingPrim(Napi::Env env) {
+  Napi::Object exports = Napi::Object::New(env);
+  exports.Set("errorHandlingPrim", Napi::Function::New<Test>(env));
+  return exports;
+}
```
gabrielschulhof, line 13: Please add a newline to the end of the file!

napi-inl.h
```diff
@@ -2595,12 +2595,110 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+    // create a reference to this wrapper
+    if (status != napi_ok) {
+      napi_value wrappedErrorObj;
+      status = napi_create_object(env, &wrappedErrorObj);
+
+      NAPI_FATAL_IF_FAILED(status, "Error::Error", "napi_create_object");
+
+      napi_property_descriptor errValDesc = {
+          "errorVal",                     // const char* utf8Name
+          String::From(env, "errorVal"),  //  napi_value name
```
gabrielschulhof, line 2608: This can be `nullptr`. `utf8name` is sufficient for specifying the name of the property.

napi-inl.h
```diff
@@ -2595,12 +2595,110 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+                                              nullptr,
+                                              nullptr,
+                                              nullptr,
+                                              nullptr,
+                                              Value::From(env, value),
+                                              napi_enumerable,
+                                              nullptr};
+
+#ifdef NODE_ADDON_API_ENABLE_MAYBE
+      Symbol uniqueSymb;
```
gabrielschulhof, line 2630: 1. Move this declaration outside the `#ifdef`.

napi-inl.h
```diff
@@ -2595,12 +2595,110 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+      if (result.IsJust()) {
+        uniqueSymb = result.Unwrap();
+      }
+
+      wrapObjFlag.name = uniqueSymb;
+      status = napi_define_properties(env, wrappedErrorObj, 1, &wrapObjFlag);
+      NAPI_FATAL_IF_FAILED(status, "Error::Error", "napi_define_properties");
+#else
+
+      wrapObjFlag.name = result;
```
gabrielschulhof, line 2640:
```suggestion
      uniqueSymb = result;
      wrapObjFlag.name = result;
```

napi-inl.h
```diff
@@ -2595,12 +2595,110 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+      status = napi_define_properties(env, wrappedErrorObj, 1, &wrapObjFlag);
+      NAPI_FATAL_IF_FAILED(status, "Error::Error", "napi_define_properties");
```
gabrielschulhof, line 2642: 3. Move these two statements below the `#endif`. That way, you are not duplicating them in both bodies of the `#if...#else...#endif`, thereby reducing the size of the the `#if...#else...#endif` block.

napi-inl.h
```diff
@@ -2595,12 +2595,110 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+
+#ifdef NODE_ADDON_API_ENABLE_MAYBE
+      Symbol uniqueSymb;
+      if (result.IsJust()) {
+        uniqueSymb = result.Unwrap();
+      }
+
+      wrapObjFlag.name = uniqueSymb;
+      status = napi_define_properties(env, wrappedErrorObj, 1, &wrapObjFlag);
+      NAPI_FATAL_IF_FAILED(status, "Error::Error", "napi_define_properties");
```
gabrielschulhof, line 2637: 2. Remove these two lines.

napi-inl.h
```diff
@@ -2595,12 +2595,110 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+
+  napi_valuetype type;
+  status = napi_typeof(_env, refValue, &type);
+  NAPI_THROW_IF_FAILED(_env, status, Object());
+
+  if (type != napi_symbol) {
+    // We are checking if the object is wrapped
+    bool isWrappedObject = false;
+
+    MaybeOrValue<Symbol> result = Symbol::For(_env, "isWrapObject");
```
gabrielschulhof, line 2671: Should this also be guarded by `#ifdef NODE_ADDON_API_ENABLE_MAYBE` since it's `MaybeOrValue<Symbol>`?

napi-inl.h
```diff
@@ -2595,12 +2595,110 @@ inline Error::Error(napi_env env, napi_value value) : ObjectReference(env, nullp
...
+
+#ifdef NODE_ADDON_API_ENABLE_MAYBE
+    Symbol uniqueSymb;
+    if (result.IsJust()) {
+      uniqueSymb = result.Unwrap();
+    }
+    status = napi_has_property(_env, refValue, uniqueSymb, &isWrappedObject);
```
gabrielschulhof, line 2678: Please use the same pattern here as I mentioned before. Move `Symbol uniqueSymb` outside the `#ifdef ... #else ... #endif` block and place the common code after the block so you don't have to write it twice.

JckXia:
@mhdawson @gabrielschulhof @KevinEady @NickNaso Hello everyone. I updated this pull request to opt using hard-coded strings as the property key on the error object, due to issues that came with using `Symbol::For`. Thanks!

mhdawson:
@mhdawson @gabrielschulhof @KevinEady @NickNaso do you want to take a look before I land this, if so let me know before the end of the week. If I don't get an ask for more time to review I'll go ahead and land on Friday.

mhdawson:
@JckXia  I see one comment from @gabrielschulhof that I don't think you addressed the one about lines 2620 and 2621

JckXia:
@mhdawson  I believe Gabriel's comment originally was addressing the fact that I was using a separate flag to denote the error object is wrapped on top of having an `errorVal` property. I removed one of the property descriptors and now we only attach one property `4bda9e7e-4913-4dbc-95de-891cbf66598e-errorVal` to show that the err obj is wrapped

JckXia:
Hey guys @mhdawson @gabrielschulhof @NickNaso @KevinEady. Could you take a look at this PR again? Thanks!

mhdawson (approved):
LGTM, looks good to me with all the update.  @JckXia thanks for your work on this.

KevinEady (approved):
LGTM! Thank you for clarifying the need for the string key in today's Node.JS API meeting.

JckXia:
Landed as `293c732`

PR #1079, opened 2021-09-28
test: update compiler used for testing

mhdawson:
Looks like github no longer has ubuntu 16 so we'll
need to update the compiler and remove 10.x testing.


mhdawson:
Seems like 16.04 was removed Sep 20 - https://github.com/actions/virtual-environments/issues/3287

NickNaso (approved):
LGTM

Issue #811, opened 2020-09-08
InstanceMember creates read only, not enumerable and not configurable properties on default


Flarna:
`InstanceMember` uses `napi_default` for `attributes` if nothing else is specified.

`napi_default` [means](https://nodejs.org/dist/latest-v14.x/docs/api/n-api.html#n_api_napi_property_attributes) read only, not enumerable and not configurable.

The corresponding NAN API `SetPrototypeMethod` calls `PrototypeTemplate()->Set()` and to my understanding of the v8 API they use `PropertyAttribute::None` which is defined [inverted](https://v8docs.nodesource.com/node-14.1/d2/dc3/namespacev8.html#aa944500afca8000fffef41951b9f2660):

```c
enum PropertyAttribute {
  /** None. **/
  None = 0,
  /** ReadOnly, i.e., not writable. **/
  ReadOnly = 1 << 0,
  /** DontEnum, i.e., not enumerable. **/
  DontEnum = 1 << 1,
  /** DontDelete, i.e., not configurable. **/
  DontDelete = 1 << 2
};
```

Is this difference intended? In general non configurable properties are quite uncommon in JS world.

FWIW: If I define a `class` in javascript members are configureable, writable and not enumerable.

NickNaso:
Hi @Flarna,
today I tried to define a class and then print how his attributes has been set. With JavaScript on `Node.js`, `Chrome` and `Firefox` I obtained that the instance method is set `writable` and `configurable` by default.
```js
'use strict'

class Class {
  method() {}
}

console.log(Object.getOwnPropertyDescriptors(Class.prototype))
```
```
{
 constructor: {
    value: [class Class],
    writable: true,
    enumerable: false,
    configurable: true
  },
  method: {
    value: [Function: method],
    writable: true,
    enumerable: false,
    configurable: true
  }
}
```
I did the same experiment for native add-on and I obtained the following result:

```
{
 method: {
    value: [Function: method],
    writable: false,
    enumerable: false,
    configurable: false
  },
  constructor: {
    value: [Function: Class],
    writable: true,
    enumerable: false,
    configurable: true
  }
}
```
I want to discuss about this in tomorrow meeting. Thanks for reporting.

mhdawson:
We discussed in the N-API team meeting and its a result of history and trying to match the JavaScript spec but seems like we did not end up with the default (in both the C and C++ wrapper) we might have chosen thinking about it now.

At this point though we don't want to change the default as that will affect existing addons.

[...]

PR #1094, opened 2021-10-19
Update object_wrap.md

alexanderfloh:
Add the default attributes for methods to the example to make them writeable by default.

See https://github.com/nodejs/node-addon-api/issues/811 for discussion.


mhdawson:
I don't think  `napi_default_method` is available in Node.js 12 or 14. Until all LTS versions have `napi_default_method` it might be better to update the example to add `napi_writable | napi_configurable` instead.

My assessment is based on looking at the tags shown for https://github.com/nodejs/node/commit/c9506a8f3e9bc5c679151feb39198023154464ab

mhdawson (approved):
LGTM

PR #1118, opened 2021-12-30
  Update index.js

Kailee01:


Kailee01:
added

mhdawson:
@Kailee01  there are a number of linter errors that need to be fixed up. Can you also add some context as what the purpose is (adding a test or something else).

mhdawson:
I'm going to close this as there has been not response from the opener. @Kailee01 please re-open if you feel that was not the right thing to do.

Issue #1123, opened 2022-01-11
Class constructors cannot be invoked without 'new'


p120ph37:
Currently, `ObjectWrap<T>::ConstructorCallbackWrapper` has a guard against calling the constructor as a javascript function without the "new" keyword (`TypeError: Class constructors cannot be invoked without 'new'`).  In a pure Javascript project, or in a non `node-addon-api` native project, it is possible to trap this sort of scenario and do something special other than abort -- often simply re-call the function as a constructor and return the new object, but potentially other clever things too.

If we could change the `ObjectWrap<T>::ConstructorCallbackWrapper` implementation to delegate to an overridable public static method when `!isConstructCall` rather than always aborting (e.g. `Value T::NonConstructFunction(const Napi::CallbackInfo& info)` with a default implementation that throws the usual `TypeError`), that would allow for more flexibility in porting older modules to use `node-addon-api` without changing their existing semantics.

legendecas:
Calling a constructor created from `class {}` is throwing:

```js
(class {})(); // TypeError: Class constructors cannot be invoked without 'new'
```
I think this is expected behavior. As you can still create normal functions with `Napi::Function::New`.

p120ph37:
Sure you can create normal functions with `Napi::Function::New`, but I don’t see a way in which you can create a dual-purpose function which can, depending on invocation context, act either as a normal function, or as an `ObjectWrap` constructor and `instanceof`-testable prototype.  This sort of thing is possible and even common in classic (pre-ES2015) js, and it is possible with `napi_wrap` directly, just not with the current `ObjectWrap` abstraction. For new development it is probably not a great idiom, but I have an older Node module I’m trying to port, and without this functionality, I can’t maintain backward compatibility with existing code.

It should be possible to support this feature with a fairly trivial hook.  I’ll see if I can whip up a PR in the morning…

p120ph37:
I've opened PR #1125 to add this functionality, and tests which illustrate the use-case.

KevinEady:
PR #1125 has landed in https://github.com/nodejs/node-addon-api/commit/cee899ade542d8d19429d2c32ac41f0cc5328d3a , closing issue.

PR #1125, opened 2022-01-11
Add `NonConstructor` to allow customization of ObjectWrap behavior

p120ph37:
Addresses #1123 by moving the error-throw into an overridable static method.  Default behavior remains exactly the same, but now a developer can override `NonConstructor` in their `ObjectWrap` subclass in order to define alternate behavior.  As a trivial example of behavior a developer might implement, this code in an `ObjectWrap` subclass would support the old (ugly) idiom of instantiating a class even when the function is called without the `new` keyword:
```
static Napi::Value MyObject::NonConstructor(const Napi::CallbackInfo& info) {
  std::vector<napi_value> args;
  for(size_t i = 0; i < info.Length(); i++) args.push_back(info[i]);
  return GetConstructor(info.Env()).New(args);
}
```

Which then allows this usage:
```
const MyObject = binding.MyObject;
const newConstructed = new MyObject();
const factoryConstructed = MyObject();
assert(newConstructed instanceof MyObject);
assert(factoryConstructed instanceof MyObject);
```

I've also included a new test case, which amounts to basically the above, and also tests that exceptions thrown inside the non-constructor are handled appropriately.


test/objectwrap_nonconstructor.cc
```diff
@@ -0,0 +1,46 @@
...
+      NAPI_THROW(Napi::Error::New(info.Env(), "an exception"), Napi::Value());
+    }
+    // Otherwise, act as a factory.
+    std::vector<napi_value> args;
+    for(size_t i = 0; i < info.Length(); i++) args.push_back(info[i]);
+    return MaybeUnwrap(GetConstructor(info.Env()).New(args));
+  }
+
+  static std::unordered_map<napi_env, Napi::FunctionReference*> constructors;
+
```
p120ph37, line 23: I know this is leaky without an env-shutdown-hook, but it's good enough to run the tests without clobbering `Napi::Addon<T>` via `env.SetInstanceData()` as would be the typical idiom.

p120ph37: Should probably add a comment to the source so nobody thinks it’s a good idea to copy paste this into a real project…

mhdawson:
We discussed in the node-api team meeting today. 

We agreed giving developers the option makes sense.  We would like the PR to include updating the documentation as well.

We would prefer that instead of using  NonConstructor that we use CreateViaFunctionCall

p120ph37:
Sounds reasonable to me - I'll make the requested adjustments.

p120ph37:
I'm happy to name it `CreateViaFunctionCall` if that's the consensus, but I do feel like that's a little bit misleading because the actual action may be something other than creating an object -- that's just one possible implementation for the method.  All I intended to convey with my original `NonConstructor` name was that it handles the case where the `new` keyword was not used.  Absent the `new` keyword, the function is technically allowed to do anything it wants, including returning an object of some completely different type, or returning a "builder" object, or an object-factory, etc.  So calling it `Create...` seems a little presumptive ;-)

vmoroz:
> I'm happy to name it `CreateViaFunctionCall` if that's the consensus, but I do feel like that's a little bit misleading because the actual action may be something other than creating an object -- that's just one possible implementation for the method. All I intended to convey with my original `NonConstructor` name was that it handles the case where the `new` keyword was not used. Absent the `new` keyword, the function is technically allowed to do anything it wants, including returning an object of some completely different type, or returning a "builder" object, or an object-factory, etc. So calling it `Create...` seems a little presumptive ;-)

+@mhdawson and @gabrielschulhof

@p120ph37, it is a great point! There were a few points for the alternative name:
- `NonConstructor` sounds a little bit alien and does not communicate the function purpose clearly.
- What if we use the same terms as in the JS docs? E.g. in [MDN Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) they use the phrase `When called as a function, returns a string representation of the current date and time ...`. Thus, we thought that the name should have something from the `called as function` phrase.
- The method name is better start with a verb.

Since `Create...` seems to be wrong here (even the `Date()` call does not create anything), could we use something like `HandleWhenCalledAsFunction`, `OnCalledAsFunction`, or something else along these lines? E.g. to communicate that this method handles the situation when the JS function representing the native class was called as a function rather than as a constructor.

p120ph37:
I like `OnCalledAsFunction` - it's not too verbose, and the `On` prefix is already familiar and implies the mechanism of action: the developer should define the method in order to change the default behavior.

vmoroz:
> I like `OnCalledAsFunction` - it's not too verbose, and the `On` prefix is already familiar and implies the mechanism of action: the developer should define the method in order to change the default behavior. I'll go ahead with that name first, and we can see what the doc looks like.

I like it too! Though, the final decision is from @mhdawson and @gabrielschulhof.

test/objectwrap_nonconstructor.cc
p120ph37: I've added an env cleanup-hook so this test-code shouldn't leak anymore and should be adaptable to real-world projects.

mhdawson:
adding my +1 for OnCalledAsFunction

mhdawson (approved):
LGTM

mhdawson:
Landed as cee899ade542d8d19429d2c32ac41f0cc5328d3a

PR #1134, opened 2022-02-18
src: do not use non-static class member for constant value

addaleax:
Non-static class members need to be set for each instantiated
object and take up extra space, which is why constant data
is generally provided through static members or static functions
(and static functions are a bit easier to use in header-only C++11).


santigimeno (approved):

NickNaso (approved):
LGTM

PR #1142, opened 2022-02-25
Enable running pre-commit ClangFormat on Windows

vmoroz:
When we do a Git commit locally, we automatically run clang-format to ensure proper code formatting.
Unfortunately, the current code doesn't work correctly on Windows.

This PR addresses the issue by detecting the Window OS and adjusting the target binary for clang-format.

There is also a small adjustment to the `.gitignore` file to ignore VS Code local settings.

### DIFF
--- a/tools/clang-format.js
+++ b/tools/clang-format.js
@@ -20,7 +20,10 @@ function main (args) {
   }
 
   const clangFormatPath = path.dirname(require.resolve('clang-format'));
-  const options = ['--binary=node_modules/.bin/clang-format', '--style=file'];
+  const binary = process.platform === 'win32'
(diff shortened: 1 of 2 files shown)


tools/clang-format.js
```diff
@@ -20,7 +20,10 @@ function main (args) {
   }
 
   const clangFormatPath = path.dirname(require.resolve('clang-format'));
-  const options = ['--binary=node_modules/.bin/clang-format', '--style=file'];
+  const binary = process.platform === 'win32'
+    ? 'node_modules\\.bin\\clang-format.cmd'
```
mhdawson, line 24: LGTM

mhdawson:
Landed as 57c212e15f6933fe27a85689f7c9bc964c832e64

PR #1152, opened 2022-03-19
build: run Windows CI on all nondeprecated build configurations

RaisinTen:
The windows-2016 runner has been deprecated on March 15, 2022. That is
why those runs are getting cancelled recently. This change fixes the
issue by removing such runners and adding some more nondeprecated build
configurations to the matrix.

Refs: https://github.blog/changelog/2021-10-19-github-actions-the-windows-2016-runner-image-will-be-removed-from-github-hosted-runners-on-march-15-2022
Signed-off-by: Darshan Sen <raisinten@gmail.com>

cc @mhdawson


mhdawson:
@RaisinTen thanks for the change. @richardlau I think you dealt with a similar issue in other tests. Can you remind me how you handled that. Just want to confirm removing is the way we want to go.

richardlau:
> @RaisinTen thanks for the change. @richardlau I think you dealt with a similar issue in other tests. Can you remind me how you handled that. Just want to confirm removing is the way we want to go.

For Node.js core we had Node.js 12 being tested on "windows-2016" and nothing else for which I bumped that to "windows-2019". Node.js 14 was using "windows-latest" and I had to pin that to "windows-2019" as Node.js 14 currently cannot detect Visual Studio 2022. Neither of those cases appear to apply here, so this change LGTM.

mhdawson (approved):
LGTM

Issue #1150, opened 2022-03-17
Regular hang in thread safe function in CI testing


mhdawson:
Seems consistent on Debian 8 and Node.js 12 but I've seen it intermittently on other platforms as well

https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=debian8-64/5608

mhdawson:
```
Running test 'threadsafe_function/threadsafe_function_unref'
Running test 'typed_threadsafe_function/typed_threadsafe_function'
Running test 'typed_threadsafe_function/typed_threadsafe_function_ctx'
Running test 'typed_threadsafe_function/typed_threadsafe_function_existing_tsfn'
Running test 'typed_threadsafe_function/typed_threadsafe_function_ptr'
Running test 'typed_threadsafe_function/typed_threadsafe_function_sum'
Build timed out (after 60 minutes). Marking the build as failed.
Tests aborted with SIGTERM
npm ERR! Test failed.  See above for more details.
```

mhdawson:
This time in s390 - https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=rhel8-s390x/5615/console

In past runs it seemed to fail every time on debian but not last night :).

There is also an issue on the core side which is likely related - https://github.com/nodejs/reliability/issues/224 (4 failures shown in that reliabiliity report)

KevinEady:
Hi @mhdawson ,

I installed a debian 8 VM and again I am unable to reproduce any failures locally after 100 full test runs (not just threadsafe tests). I think we need to go with the next approach, getting access to the build server?

Thanks, Kevin

mhdawson:
Created https://github.com/nodejs/build/issues/2919 to request access. Hopefully early next week we can get you access to the machine.

PR #1154, opened 2022-03-20
build: increase ci timeout

RaisinTen:
Some builds seem to run for more than 30 minutes, so this changes the
timeout-minutes value to 45.

Refs: https://github.com/nodejs/node-addon-api/runs/5612157844?check_suite_focus=true
Signed-off-by: Darshan Sen <raisinten@gmail.com>

cc @mhdawson

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   test:
-    timeout-minutes: 30
+    timeout-minutes: 45
     strategy:
       matrix:
         node-version:


mhdawson:
@RaisinTen I think we have an intermittent hang in that test. The test suite should not take nearly that long to run.

mhdawson (changes requested):
This is the PR tracking the intermittent hang - https://github.com/nodejs/node-addon-api/issues/1150

RaisinTen:
Ah, okay. My bad, closing! 😅

PR #1155, opened 2022-03-25
doc: clarify ObjectWrap weak ref behavior

mildsunrise:
Refs: https://github.com/nodejs/node/pull/42461

We currently defer finalizer callbacks until the loop is idle.

Warn users that the weak reference on `ObjectWrap` isn't guaranteed to be valid just because the object hasn't been finalized (destructed) yet.


NickNaso (approved):
LGTM

PR #1162, opened 2022-04-13
src: fix -Wconversion msg/C++17 UB from unknown_array_type

### LINKED ISSUE #1066: GCC warning with -Wconversion and undefined behavior for unknown_array_type
This line:
https://github.com/nodejs/node-addon-api/blob/da2e754a021e418a3ca36f3c10ea134c9541959f/napi.h#L1075
causes a warning with `-Wconversion` enabled:
```
 warning: the result of the conversion is unspecified because ‘-1’ is outside the range of type ‘napi_typedarray_type’ [-Wconversion]
  911 |     static const napi_typedarray_type unknown_array_type = static_cast<napi_typedarray_type>(-1);
      |
```

I don't think this is really an issue, but it would be nice to get rid of the warning.

(https://wiki.sei.cmu.edu/confluence/display/cplusplus/INT50-CPP.+Do+not+cast+to+an+out-of-range+enumeration+value)

zbjornson:
Opening gambit to fix #1066. Only lightly tested (I'll fix the errors, no need to comment on those). Seeking feedback on the approach before adding tests. 

Breaking change: this makes code like `TypedArrayOf<bool>::New()` fail at compile-time instead of run-time. The other changes I think are non-breaking.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2005,11 +2005,11 @@ inline void DataView::WriteData(size_t byteOffset, T value) const {
 ////////////////////////////////////////////////////////////////////////////////
 
 inline TypedArray::TypedArray()
-  : Object(), _type(TypedArray::unknown_array_type), _length(0) {
+  : Object(), _type_is_known(false), _length(0) {
 }
 
 inline TypedArray::TypedArray(napi_env env, napi_value value)
-  : Object(env, value), _type(TypedArray::unknown_array_type), _length(0) {
+  : Object(env, value), _type_is_known(false), _length(0) {
 }
 
 inline TypedArray::TypedArray(napi_env env,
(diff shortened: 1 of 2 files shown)


mhdawson:
> Breaking change: this makes code like TypedArrayOf<bool>::New() fail at compile-time instead of run-time

I'm not sure we want a breaking change if your earlier comment:

> I don't think this is really an issue, but it would be nice to get rid of the warning.

is correct. Is there no way to fix the warning without breaking existing code?

mhdawson:
We looked at this in detail today. The thought was that the right approach might be to remove the concept of an unknown type completely.

In the contructor with no options, then just use a default type (maybe int8)

In the contructor that passes in an napi_value then do the lookup immediately versus defering.

We can't think of any way this should affect working existing code.

mhdawson:
@KevinEady is going to take a look at implementing the suggestion to remove concept of an unknown type completely.

mhdawson:
Closing as this will be fixed by https://github.com/nodejs/node-addon-api/pull/1209 instead.

Issue #1175, opened 2022-05-08
eslint issue on Windows


egg-bread:
When running `npm run lint:fix` on Windows bash, lint doesn't actually work:
![image](https://user-images.githubusercontent.com/38633553/167314968-147d74c6-d8e7-42eb-9c3a-662038dcc3c9.png)

But when the eslint path is changed to `node_modules\\.bin\\eslint.cmd` in tools/eslint-format.js, lint works.

JckXia:
Fixed with https://github.com/nodejs/node-addon-api/pull/1176

PR #1176, opened 2022-05-08
Add condition for window to find eslint

JckXia:
A small fix for `eslint-format` based on the issue https://github.com/nodejs/node-addon-api/issues/1175. At the moment `npm run lint/fix` is failing silently on windows because it cannot find `eslint` cmd on windows and we do not check for that.


KevinEady (approved):
LGTM! Tested on my windows machine

JckXia:
Thanks for the review @KevinEady!

JckXia:
Landed as 33e402971e91ca423ce6c4b631e345531b1a746c

PR #1186, opened 2022-06-04
Update LICENSE.md

chineduG:
Corrections of misspelled words on the License file.

 a


Trott:
I don't think we'd consider changing a single character of the MIT license boilerplate text.

PR #1196, opened 2022-08-13
test: Add test coverage for "TSFN::Ref()"

JckXia:
We currently do not have a test for testing the `Ref(Env env)` function attached to the TSFN class. But I am not too sure if this is the best way to test this functionality. From what I understand this method is a wrapper around `napi_ref_threadsafe_function` which prevents the event loop from exiting until the tsfn gets destroyed by a call to `Release()`. 

My thought behind this approach is that we first reference this TSFN function, before the callback in `setTimeout`  Unref's the TSFN which causes the event loop to exit. Though it appears that this tsfn should already be referenced prior to calling `Ref`, and `napi_ref_threadsafe_function` is idempotent so it has no effect.  Is there a better approach to test this method? Also in what scenarios would `napi_ref_threadsafe_function` or `TSFN::Ref()` be used? Thanks!

### DIFF
--- a/test/typed_threadsafe_function/typed_threadsafe_function_unref.cc
+++ b/test/typed_threadsafe_function/typed_threadsafe_function_unref.cc
@@ -63,6 +63,7 @@ static Value TestRef(const CallbackInfo& info) {
       static_cast<FinalizerDataType*>(nullptr));
 
   tsfn->BlockingCall();
+  tsfn->Unref(info.Env());
   tsfn->Ref(info.Env());
 
   return info.Env().Undefined();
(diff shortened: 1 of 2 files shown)


test/typed_threadsafe_function/typed_threadsafe_function_unref.cc
```diff
@@ -32,6 +32,7 @@ static Value TestUnref(const CallbackInfo& info) {
       static_cast<FinalizerDataType*>(nullptr));
 
   tsfn->BlockingCall();
+  tsfn->Ref(info.Env());
```
mhdawson, line 35: Is this change still needed?

mhdawson (approved):
LGTM

PR #1202, opened 2022-08-29
test: Add test case for canceling async worker tasks

JckXia:
Adding a small test case for Async workers' `Cancel`  functionality, where if the cancellation is successful neither `OnOk` or `OnError` should be invoked.

### DIFF
--- a/test/async_worker.cc
+++ b/test/async_worker.cc
@@ -125,7 +125,9 @@ class CancelWorker : public AsyncWorker {
   static void DoWork(const CallbackInfo& info) {
     Function cb = info[0].As<Function>();
     std::string echo = info[1].As<String>();
-    for (int i = 1; i < MAX_CANCEL_THREADS; i++) {
+    int threadNum = info[2].As<Number>().Uint32Value();
+
+    for (int i = 0; i < threadNum; i++) {
       AsyncWorker* worker = new EchoWorker(cb, echo);
       worker->Queue();
       assert(worker->Env() == info.Env());
@@ -133,7 +135,17 @@ class CancelWorker : public AsyncWorker {
 
     AsyncWorker* cancelWorker = new CancelWorker(cb);
     cancelWorker->Queue();
+
+#ifdef NAPI_CPP_EXCEPTIONS
+    try {
+      cancelWorker->Cancel();
+    } catch (Napi::Error& e) {
+      Napi::Error::New(info.Env(), "Unable to cancel async worker tasks")
+          .ThrowAsJavaScriptException();
+    }
+#else
     cancelWorker->Cancel();
+#endif
   }
 
   void Execute() override {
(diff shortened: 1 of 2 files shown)


test/async_worker.cc
```diff
@@ -95,12 +98,65 @@ class TestWorkerNoCallback : public AsyncWorker {
...
+    // Simulate cpu heavy task
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
+
+  void OnOK() override {
+    NAPI_THROW_IF_FAILED_VOID(this->Env(), napi_generic_failure);
+  }
+
+  void OnError(const Error&) override {
+    NAPI_THROW_IF_FAILED_VOID(this->Env(), napi_generic_failure);
```
legendecas, line 149: It can be more intuitive if we can throw an error with an explicit message about the reason here:

```suggestion
    Napi::Error::New(env, "Unreachable").ThrowAsJavaScriptException();
```

test/async_worker.cc
```diff
@@ -95,12 +98,65 @@ class TestWorkerNoCallback : public AsyncWorker {
...
+    cancelWorker->Cancel();
+  }
+
+  void Execute() override {
+    // Simulate cpu heavy task
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
+
+  void OnOK() override {
+    NAPI_THROW_IF_FAILED_VOID(this->Env(), napi_generic_failure);
```
legendecas, line 145: Ditto

test/async_worker.cc
```diff
@@ -95,12 +98,65 @@ class TestWorkerNoCallback : public AsyncWorker {
...
+
+class CancelWorker : public AsyncWorker {
+ public:
+  CancelWorker(Function& cb) : AsyncWorker(cb) {}
+  ~CancelWorker() {}
+
+  static void DoWork(const CallbackInfo& info) {
+    Function cb = info[0].As<Function>();
+    std::string echo = info[1].As<String>();
+    for (int i = 1; i < MAX_CANCEL_THREADS; i++) {
```
legendecas, line 128: Maybe we can get the cpu number with JS API and fill the worker threads with that number of workers? In that case, the test can be more stable.

With JS specified worker filling number, we can also test that `cancelWorker->Cancel()` throws if the cancel worker has been scheduled.

legendecas (commented):
This is a great job!

test/async_worker.cc
JckXia: Hey Legendecas!  Thank you for the review. By cpu numbers do you mean the size of the Libuv's thread pool? I tested this by setting `UV_THREADPOOL_SIZE` to smaller numbers such as 2 or 3 and it seems to work reliably in that the CancelWorker is able to cancel the queue'd task. 

Also, I am running into some difficulties testing that `Cancel()` throws if the worker has started the task. Initially, I just passed the value 0 as the threadNum to queue a task for the Cancel worker to cancel, expecting an error to be thrown. For some reason, this exception is only thrown once by `Cancel()`. 

After doing some investigation, it looks like we run this test using `common.runTest` where we run the test function using 4 different node-gyp configs in rapid succession. Could it be that by the second time the `test` function is invoked, the thread pool is busy and cancellation is successful? If that's the case is there a way where we could check to make sure the threads are idle before running the test suite? Thanks.

legendecas: Looks like we already have a similar test case in core (https://github.com/nodejs/node/blob/main/test/node-api/test_async/test_async.c#L140) and it has been running in a quite stable way for a long time. If that's the case, I'm fine with the current pattern as it is the simplest way to test the API.

Yeah, if the test case is mixed up with other test suites, we can not guarantee that the threadpool is ready to execute the work immediately. But I believe we can still facilitate mutexes and conditional variables to notify the JS thread when the Worker is started to execute. However, this increases the complexity of this test case and I assume it is fine to do it in a follow up PR.

legendecas (approved):

JckXia:
Landed as 793268c59f8c14a56f44ef00a7711c18d1c35b95

PR #1208, opened 2022-09-20
test: remove update to process.config

mhdawson:
process.config is now read-only. Remove
attempt to update it. This may break the ability
to run tests in debug mode but is needed to get ci running again.


mhdawson:
going to push ones a few more of the CI check pass, so that we get the jenkins ci back to green.

NickNaso (approved):

Issue #1198, opened 2022-08-18
Building with clang-16 error `-Wenum-constexpr-conversion`


ammarfaizi2:
Hi,

Building with clang-16 error `-Wenum-constexpr-conversion`:

```
napi.h:1173:7: error: integer value -1 is outside the valid range of values [0, 15] for this enumeration type [-Wenum-constexpr-conversion]
      static_cast<napi_typedarray_type>(-1);
      ^
1 error generated.
```

NickNaso:
Hi @ammarfaizi2,
I think that this issue is correlate with PR https://github.com/nodejs/node-addon-api/issues/1066 and issue https://github.com/nodejs/node-addon-api/issues/1066 wea re working on solve it. Thanks for reporting.

NickNaso:
I'm closing the issue because it has been solved with PR https://github.com/nodejs/node-addon-api/pull/1209

PR #1209, opened 2022-09-21
src: remove TypedArray::unknown_array_type

KevinEady:
Removes `TypedArray::unknown_array_type` and makes the default (if not specified) as `napi_int8_array`.

Fixes: #1066 
Fixes: #1198 
Closes: #1162


mhdawson (approved):
LGTM

mhdawson:
Marking this as SemVer major out of an abundanc of caution. We don't believe this will affect any code that could be working, but might as well signal as SemVer major just in case.

mhdawson:
Landed as a8afb2d73c368a23ab9ed157b67e1cf8587d39c4

PR #1214, opened 2022-09-29
test: Add test cased for failed task cancellations

JckXia:
A follow up to https://github.com/nodejs/node-addon-api/pull/1202#discussion_r974497323. Adding a test for checking when a task cancellation is unsuccessful (task is already queued and executing).

### DIFF
--- a/test/async_worker.cc
+++ b/test/async_worker.cc
@@ -119,22 +119,35 @@ class EchoWorker : public AsyncWorker {
   std::string echo;
 };
 
-std::condition_variable cv;
-std::mutex m;
-bool ready = false;
 class FailCancelWorker : public AsyncWorker {
+ private:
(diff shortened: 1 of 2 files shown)


test/async_worker.cc
```diff
@@ -117,6 +119,49 @@ class EchoWorker : public AsyncWorker {
   std::string echo;
 };
 
+std::condition_variable cv;
```
legendecas, line 122: Might be worth moving these variables to be members of `FailCancelWorker` instead.

JckXia: Yep I agree. Made new changes

legendecas (approved):

JckXia:
Landed as 257a52f82321a344dbb04f1dc6b4774ca40ee95e

Issue #1095, opened 2021-10-22
AsyncProgressQueueWorker::Signal() does not trigger the TSFN


KevinEady:
As reported in https://github.com/nodejs/node-addon-api/pull/1086#issue-1017797535

KevinEady:
Hi @legendecas , can you elaborate some more on the problem? Maybe we can discuss in today's call.

KevinEady:
😅  The above fix might be a bit premature. Let's discuss in the call.

legendecas:
@KevinEady the problem introduced in #853 is that it unconditionally suppressed the calls when the data is nullptr, for timing details you can checkout the graph here https://github.com/nodejs/node-addon-api/issues/821#issuecomment-723626374. So I believe we should try to find a way to distinguish `Signal`s (a user-initiated action) from duplicated activation from second uv idle callback (not a user-initiated action).

KevinEady:
Current state: I have an implementation that solves this issue, but it suffers from the "not all of the async `Signal()` calls (which calls the TSFN `NonBlockingCall()`) get executed. We discussed in 28 Jan meeting to port the implementation over to Node-API directly (without using the node-addon-api wrapper) to see if it's an issue with the underlying thread-safe functions APIs or with the wrapper. This is currently ongoing.

KevinEady:
Fixed by #1216

Issue #1081, opened 2021-09-29
`const` error with `AsyncProgressWorker<T>::ExecutionProgress::Signal()`


JosephusPaye:
Hi 👋 

I'm getting the following error trying to call `.Signal()` on `AsyncProgressWorker::ExecutionProgress`:

<details>
<summary>View error</summary>

```
In file included from /home/jpaye/nuclearnet.js/node_modules/node-addon-api/napi.h:2905,
                 from ../src/NetworkBinding.hpp:22,
                 from ../src/NetworkListener.hpp:21,
                 from ../src/NetworkListener.cpp:18:
/home/jpaye/nuclearnet.js/node_modules/node-addon-api/napi-inl.h: In instantiation of ‘void Napi::AsyncProgressWorker<T>::Signal() const [with T = char]’:
/home/jpaye/nuclearnet.js/node_modules/node-addon-api/napi-inl.h:5797:12:   required from ‘void Napi::AsyncProgressWorker<T>::ExecutionProgress::Signal() const [with T = char]’
../src/NetworkListener.cpp:83:22:   required from here
/home/jpaye/nuclearnet.js/node_modules/node-addon-api/napi-inl.h:5792:3: error: passing ‘const Napi::AsyncProgressWorker<char>’ as ‘this’ argument discards qualifiers [-fpermissive]
 5792 |   this->NonBlockingCall(static_cast<T*>(nullptr));
      |   ^~~~
In file included from /home/jpaye/nuclearnet.js/node_modules/node-addon-api/napi.h:2905,
                 from ../src/NetworkBinding.hpp:22,
                 from ../src/NetworkListener.hpp:21,
                 from ../src/NetworkListener.cpp:18:
/home/jpaye/nuclearnet.js/node_modules/node-addon-api/napi-inl.h:5640:20: note:   in call to ‘napi_status Napi::AsyncProgressWorkerBase<DataType>::NonBlockingCall(DataType*) [with DataType = void]’
 5640 | inline napi_status AsyncProgressWorkerBase<DataType>::NonBlockingCall(DataType* data) {
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make: *** [nuclearnet.target.mk:119: Release/obj.target/nuclearnet/src/NetworkListener.o] Error 1
make: Leaving directory '/home/jpaye/nuclearnet.js/build'
```
</details>

From reading the code, this is what I've been able to piece together this sequence of calls:

```cpp
[...]

PR #1216, opened 2022-10-07
src: fix implementation of AsyncProgressWorker::Signal

KevinEady:
Fix implementation of `AsyncProgressWorker::Signal` and `AsyncProgressQueueWorker::Signal`

Fixes: #1095 
Fixes: #1081

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -5950,7 +5950,8 @@ inline AsyncProgressWorker<T>::AsyncProgressWorker(const Object& receiver,
                                                    const Object& resource)
     : AsyncProgressWorkerBase(receiver, callback, resource_name, resource),
       _asyncdata(nullptr),
-      _asyncsize(0) {}
+      _asyncsize(0),
+      _signaled(false) {}
 
 #if NAPI_VERSION > 4
 template <class T>
@@ -5990,12 +5991,15 @@ template <class T>
 inline void AsyncProgressWorker<T>::OnWorkProgress(void*) {
   T* data;
   size_t size;
+  bool signaled;
   {
     std::lock_guard<std::mutex> lock(this->_mutex);
     data = this->_asyncdata;
     size = this->_asyncsize;
+    signaled = this->_signaled;
     this->_asyncdata = nullptr;
     this->_asyncsize = 0;
+    this->_signaled = false;
   }
 
   /**
@@ -6005,7 +6009,7 @@ inline void AsyncProgressWorker<T>::OnWorkProgress(void*) {
    * the deferring the signal of uv_async_t is been sent again, i.e. potential
    * not coalesced two calls of the TSFN callback.
    */
-  if (data == nullptr) {
+  if (data == nullptr && !signaled) {
     return;
   }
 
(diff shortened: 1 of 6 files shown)


mhdawson:
@KevinEady could you add a short explanation of how using SendProgress instead of Signal resolves the problem?  I think I know but not 100% sure.

@legendecas would be good if you reviewed as well since it seems like you were in the loop on the discussions of the original problem.

KevinEady:
Hi @mhdawson ,

The `AsyncProgress[Queue]Worker::Signal()` method is supposed to call the `OnProgress` worker with `nullptr` for `data` and `0` for `count`. The main change is not using `NonBlockingCall()` but instead using `SendProgress_()`.

I went about this PR by looking at the existing test we have:

https://github.com/nodejs/node-addon-api/blob/a8afb2d73c368a23ab9ed157b67e1cf8587d39c4/test/async_progress_worker.cc#L82-L83

Since sending `nullptr` with `0` data is allowed, I wrote `Signal()` to behave in this manner, by basically making `progress.Signal()` behave as `progress.Send(nullptr, 0)`, which means using the `SendProgress_()` method.

Hope this clarifies!

Thanks, Kevin

mhdawson (approved):
LGTM

mhdawson:
@KevinEady many thanks for the explanation

napi-inl.h
```diff
@@ -6032,12 +6032,12 @@ inline void AsyncProgressWorker<T>::SendProgress_(const T* data, size_t count) {
 
 template <class T>
 inline void AsyncProgressWorker<T>::Signal() const {
-  this->NonBlockingCall(static_cast<T*>(nullptr));
+  this->SendProgress_(static_cast<T*>(nullptr), 0);
```
legendecas, line 6035: This will override saved progress and make unexpected side effects. We should save a bit of the signal and proceed with null data at https://github.com/nodejs/node-addon-api/blob/main/napi-inl.h#L5998 instead.

legendecas (commented):
Sorry for the delay. 

`AsyncProgressWorker::Signal` should not invalidate sent progress not handled, similar to https://github.com/nodejs/nan/blob/main/nan.h#L2088. Would you mind adding a test case for this?

napi-inl.h
KevinEady: Hi @legendecas , please take a look at [1f2cd71](https://github.com/nodejs/node-addon-api/pull/1216/commits/1f2cd71a281c393ae7663b3c9513030bf3fe1b5a). Is this what you are talking about with 'save a bit of the signal'?

I also added a test case were an `OnProgress` and `Signal` are performed in sequence, and checks that the progress data is present.

KevinEady:
Another thing @legendecas , i did not make any changes to `AsyncProgressQueueWorker` regarding a state member like [1f2cd71](https://github.com/nodejs/node-addon-api/pull/1216/commits/1f2cd71a281c393ae7663b3c9513030bf3fe1b5a). Since the queue variety of the progress worker stores all progress updates, I don't think it is necessary to update the queue worker. Does this sound right?

napi-inl.h
legendecas: Thanks, this looks good!

legendecas (approved):
LGTM. I think it would be great to add a caveat on https://github.com/nodejs/node-addon-api/blob/main/doc/async_worker_variants.md#onprogress about the nullability of the data parameter if the method `Signal` is used.

legendecas (approved):

mhdawson:
Landed as edf630cc79eeef1029d3c1e7e47dcbcf2ab6d47b

Issue #1207, opened 2022-09-20
Testing broken on Node.js main branch


mhdawson:
Fails with:

```shell
/home/iojs/build/workspace/node-test-node-addon-api-new/nodes/debian10-x64/node-addon-api/test/index.js:93
process.config.target_defaults.default_configuration =
                                                     ^

TypeError: Cannot assign to read only property 'default_configuration' of object '#<Object>'
    at Object.<anonymous> (/home/iojs/build/workspace/node-test-node-addon-api-new/nodes/debian10-x64/node-addon-api/test/index.js:93:54)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
    at Module._load (node:internal/modules/cjs/loader:878:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12)
    at node:internal/main/run_main_module:23:47
```

mhdawson:
Likely related to https://github.com/nodejs/node/commit/e0ab8dd63731d8eb92053776c5160fb190d5aea5

mhdawson:
This is the code in the test script

```
process.config.target_defaults.default_configuration =
  fs
    .readdirSync(path.join(__dirname, process.env.REL_BUILD_PATH || '', 'build'))
    .filter((item) => (item === 'Debug' || item === 'Release'))[0];

```

mhdawson:
I'm going to remove that code so that the ci tests start running again, that will break the ability to run in debug I think, so I'll leave this issue open until we figure that out.

mhdawson:
PR https://github.com/nodejs/node-addon-api/pull/1208 to remove and get ci working again.

mhdawson:
CI is green again.

mhdawson:
Places where we read `default_configuration`

https://github.com/nodejs/node-addon-api/search?q=default_configuration

mhdawson:
@NickNaso is going to work on doing updates the replace the usage of default_configuration in the existing test files.

Discussion around how to choose. 
  * Command line flag (preferred, by default use Release by default)
  * File system check

NickNaso:
I'm closing the issue because the PR https://github.com/nodejs/node-addon-api/pull/1226 has been merged.

PR #1226, opened 2022-10-28
test: removed the usage of default_configuration.

NickNaso:
This PR remove the usage of  `process.config.target_defaults.default_configuration` in all test suite.
A new function `whichBuildType` has been introduced and it has the reponsability to check which type of build `node-gyp` made so the test can refer to the right folders.

Refs:
- https://github.com/nodejs/node-addon-api/issues/1207

### DIFF
--- a/test/common/index.js
+++ b/test/common/index.js
@@ -91,8 +91,8 @@ async function checkBuildType (buildType) {
 }
 
 async function whichBuildType () {
-  let buildType;
-  const envBuildType = process.env.BUILD_TYPE;
+  let buildType = 'Release';
+  const envBuildType = process.env.NODE_API_BUILD_CONFIG;
   if (envBuildType) {
     if (Object.values(buildTypes).includes(envBuildType)) {
       if (await checkBuildType(envBuildType)) {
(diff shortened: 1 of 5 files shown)


test/common/index.js
```diff
@@ -75,9 +76,35 @@ exports.mustNotCall = function (msg) {
...
+async function checkBuildType (buildType) {
+  try {
+    await access(path.join(path.resolve('./test/build'), buildType));
+    return true;
+  } catch {
+    return false;
+  }
+}
 
+async function whichBuildType () {
```
legendecas, line 93: What about determining with a simple environ like `BUILD_TYPE`?

legendecas: Note that if people incremental build with `node-gyp build -C test --debug`, the build directory will not be cleaned and `build/Release` can still present.

test/common/index.js
```diff
@@ -91,12 +91,25 @@ async function checkBuildType (buildType) {
...
 
 async function whichBuildType () {
-  let buildType = 'Release';
-  if (await checkBuildType(buildTypes.Release)) {
-    buildType = buildTypes.Release;
-  }
-  if (await checkBuildType(buildTypes.Debug)) {
-    buildType = buildTypes.Debug;
+  let buildType;
+  const envBuildType = process.env.BUILD_TYPE;
```
mhdawson, line 95:
```suggestion
  const envBuildType = process.env.BUILD_TYPE || 'Release';
```

mhdawson: Suggestion in meeting from @vmoroz how about `process.env.NODE_API_BUILD_CONFIG`

legendecas (approved):

mhdawson (approved):
LGTM

mhdawson:
Landed as 10ad762807925c88109e45db728bc0ed87153a2d

Issue #1238, opened 2022-11-23
Class Env::CleanupHook should be public


julianmesa-gitkraken:
I have created a Cleanup hook using `env.AddCleanupHook(cleanup, this)` but I cannot store the hook in a member var because the class Env::CleanupHook is private.
So I cannot remove the hook calling to Remove function.
I make it public to test my app and works perfect the only issue is that I think it will be a good idea to create an empty default contructor so we can create a member class var without issue.
I can create a PR making this class public, but not sure why this class was designed to be private.

Also found an error in [line 361](https://github.com/julianmesa-gitkraken/node-addon-api/blob/main/napi.h#L361)
Then `#endif` should be betwen lines 359 - 360

julianmesa-gitkraken:
PR made: [1240](https://github.com/nodejs/node-addon-api/pull/1240)

NickNaso:
I'm closing the issue beacuse the PR https://github.com/nodejs/node-addon-api/pull/1240 has been merged.

PR #1240, opened 2022-11-23
Make CleanupHook public

julianmesa-gitkraken:
Make CleanupHook public.
Added default constructor to make easier store the hook in a member class var.
Also fixed a bad #endif position.

This PR fix [issue 1238](https://github.com/nodejs/node-addon-api/issues/1238)

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -6234,6 +6234,11 @@ Env::CleanupHook<Hook> Env::AddCleanupHook(Hook hook) {
 }
 
 template <typename Hook, typename Arg>
+Env::CleanupHook<Hook, Arg>::CleanupHook() {
+  data = nullptr;
+}
+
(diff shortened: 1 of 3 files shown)


julianmesa-gitkraken:
> Overall LGTM. Thank you for working on this!
> 
> Would you mind adding a test that verifies the `CleanupHook` is available to be used as a class member?

@legendecas Added a very simple test, not sure if this is enough

legendecas (approved):
Thank you for working on this!

mhdawson:
Seems like since the docs do identify the type which is returned it should be public so agree this makes sense.

mhdawson (approved):
LGTM

mhdawson:
Landed as 39267baf1ba985275f74991fef6faac2ccf8e495

PR #1245, opened 2022-12-04
src, test: fix errors and warnings in VS 2017

vmoroz:
This PR has fixes for errors and warnings when running `npm test` with Visual Studio 2017.
- `napi-inl.h` has changes to address the same issue as PR #1242
  - fixed use of `static` keyword for standalone functions. Each `static` function is duplicated in every cpp file that includes the header. Thus, it produces code that is unnecessary bigger.
- `addon_data.cc` - warning about incorrect use of format specifier
- `async_worker.cc` - unused variable warning
- `binding.gyp` - no warning - just using the same Python quotes as the rest of the code
- `function_reference.cc` - warning about possible buffer overflow. The code is changed to avoid memory leak.
- `threadsafe_function.cc`, `typed_threadsafe_function.cc` - warning about `napi-inl.h` shadowing the `tsfn` global variable declared in these files.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -30,11 +30,11 @@ namespace details {
 // TODO: Replace this code with `napi_add_finalizer()` whenever it becomes
 // available on all supported versions of Node.js.
 template <typename FreeType>
-static inline napi_status AttachData(napi_env env,
-                                     napi_value obj,
-                                     FreeType* data,
-                                     napi_finalize finalizer = nullptr,
-                                     void* hint = nullptr) {
+inline napi_status AttachData(napi_env env,
+                              napi_value obj,
+                              FreeType* data,
+                              napi_finalize finalizer = nullptr,
+                              void* hint = nullptr) {
   napi_status status;
   if (finalizer == nullptr) {
     finalizer = [](napi_env /*env*/, void* data, void* /*hint*/) {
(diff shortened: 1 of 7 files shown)


KevinEady:
Hi @vmoroz , can you clarify a bit on:

> threadsafe_function.cc, typed_threadsafe_function.cc - warning about napi-inl.h shadowing the tsfn global variable declared in these files.

Which `tsfn` declaration in napi-inl.h is being shadowed?

vmoroz:
> Hi @vmoroz , can you clarify a bit on:
> 
> > threadsafe_function.cc, typed_threadsafe_function.cc - warning about napi-inl.h shadowing the tsfn global variable declared in these files.
> 
> Which `tsfn` declaration in napi-inl.h is being shadowed?

For example, [here](https://github.com/nodejs/node-addon-api/blob/39267baf1ba985275f74991fef6faac2ccf8e495/napi-inl.h#L4937) we define a local variable `tsfn`. The issue is that the tests define a global variable with the same name. Then the VS2017 issues a warning that the local variable is shadowing the global variable defined in the unit tests. The fix is to change the name of the global variables in tests to `s_tsfn`.

KevinEady (approved):
LGTM!

mhdawson:
Kicked off jenkins run here - https://ci.nodejs.org/job/node-test-node-addon-api-new/6750/

Set to only run for windows and Node.js 14.x since I don't think this should affect any other platforms.

mhdawson (approved):
LGTM

mhdawson:
Landed in 97736c93f47d2271fb91255d633d7031b533a964

PR #1252, opened 2022-12-16
chore: add Windows with VS 2022 and Node.JS 19.x to the CI matrix

vmoroz:
In this PR we are adding `windows-2022` OS image to the CI to test Node-API with Visual Studio 2022.
We are also adding verification for Node.JS version 19.x.

The new verification matrix for Windows is:
- `windows-2022` - Node,JS [ `16.x`, `18.x`, `19.x` ]
- `windows-2019` - Node,JS [ `14.x`, `16.x`, `18.x`, `19.x` ]
Note that VS 2022 is not supported by Node.JS 14.x GYP and it is not in the matrix.

For the MacOS and Ubuntu we have added printing of compiler version.
It helped to identify that:
- On MacOS the `gcc` is an alias for `clang`. So, we are removing the `MacOS + gcc` from the test matrix.
- On Ubuntu the condition `"${{ matrix.os }}" = ubuntu-*` seems to always fail and we never install the GCC version 6.5. Fixing this condition caused the script error because the package could not be found. CI uses preinstalled GCC version with the Ubuntu image which is currently `11.3`. Thus, the custom logic to install compiler is removed.

The new verification matrix for MacOS and Ubuntu is:
- `ubuntu-latest` - compiler: [ `gcc`, `clang` ] - Node,JS [ `14.x`, `16.x`, `18.x`, `19.x` ]
- `macos-latest` - compiler:  [ `clang` ] - Node,JS [ `14.x`, `16.x`, `18.x`, `19.x` ]

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -10,6 +10,7 @@ jobs:
         node-version:  [14.x, 16.x, 18.x]
         os:
           - windows-2019
+          - windows-2022
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -7,10 +7,11 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version:  [14.x, 16.x, 18.x]
-        os:
-          - windows-2019
-          - windows-2022
+        node-version: [14.x, 16.x, 18.x]
+        os: [ windows-2019 ]
+        include:
+          - node-version: 18.x
+            os: windows-2022
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -8,7 +8,9 @@ jobs:
     strategy:
       matrix:
         node-version: [ 16.x, 18.x, 19.x ]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-win.yml
```diff
@@ -7,9 +7,11 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version:  [14.x, 16.x, 18.x]
-        os:
-          - windows-2019
+        node-version: [14.x, 16.x, 18.x]
```
mhdawson, line 10: we should probably ad 19.x here as well

vmoroz: Good point! Let me try adding it,

vmoroz: @mhdawson , I have added Node.JS 19.x and updated the PR description.
Should I add Node.JS 19.x for non-Windows platforms in this PR too?

mhdawson: @vmoroz yes, we should have it for all platforms. We probably should have done it in Oct when 19.x was cut.

vmoroz: OK, I will do it then, while I am editing these files anyway. :)

vmoroz: @mhdawson , I have added Node.JS 19x to the MacOS and Ubuntu runners. I also cleaned up the script a bit.

legendecas (approved):

mhdawson (approved):
LGTM

Issue #1234, opened 2022-11-15
Get napi_callback_info from Napi::CallbackInfo?


erikjalevik:
Is there no way to get hold of the wrapped `napi_callback_info` from a `Napi::CallbackInfo`?

I need to call a function that's been written to accept the C NAPI type from one written with node-addon-api's C++ classes.

legendecas:
It is not supported yet. But I believe it is reasonable to have a conversion method for `CallbackInfo` to `napi_callback_info` as that's how we dealt with other wrapper types.

mhdawson:
@erikjalevik are you interested in submitting a PR?

NickNaso:
I'm closing the issue because it has been solved by the PR https://github.com/nodejs/node-addon-api/pull/1253.

Issue #1048, opened 2021-08-22
[Test] Unit test coverage for CallbackInfo


JckXia:
| class                   | methods                                                                                                                          |
|-------------------------|----------------------------------------------------------------------------------------------------------------------------------|                           
 |CallbackInfo |     |
 |Covered indirectly via call to Function::New|  CallbackInfo(napi_env env, napi_callback_info info)|
 |Covered indreictly via call to create array buffers, etc| Napi::Env Env() const|
 |Covered |Value NewTarget() const |
 |Covered (see function.cc) |bool IsConstructCall() const |
 |Covered, function used by many tests |size_t Length() const |
 |Covered, function used by various tests |const Value operator[] (size_t index) const |
 |Covered, (see addon_data.cc)|Value This() const |
 |Covered, see function.cc |void * Data() const |
 ||void SetData(void * Data) | 
 
Just wondering if it's worth writing seperate unit tests for this class?

JckXia:
Completed following  https://github.com/nodejs/node-addon-api/pull/1253

PR #1253, opened 2022-12-20
src: Add an operator to retrieve raw `napi_callback_info` from `CallbackInfo`

JckXia:
Resolves https://github.com/nodejs/node-addon-api/issues/1234, and also complete unit test coverage for  https://github.com/nodejs/node-addon-api/issues/1048

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3451,6 +3451,10 @@ inline CallbackInfo::~CallbackInfo() {
   }
 }
 
+inline CallbackInfo::operator napi_callback_info() const {
+  return _info;
+}
+
 inline Value CallbackInfo::NewTarget() const {
   napi_value newTarget;
   napi_status status = napi_get_new_target(_env, _info, &newTarget);
--- a/napi.h
+++ b/napi.h
@@ -1791,6 +1791,7 @@ class CallbackInfo {
   Value This() const;
   void* Data() const;
   void SetData(void* data);
+  operator napi_callback_info() const;
 
  private:
   const size_t _staticArgCount = 6;
--- a/test/binding.cc
+++ b/test/binding.cc
@@ -28,6 +28,7 @@ Object InitCallbackScope(Env env);
 #if (NAPI_VERSION > 4)
 Object InitDate(Env env);
 #endif
+Object InitCallbackInfo(Env env);
 Object InitDataView(Env env);
 Object InitDataViewReadWrite(Env env);
 Object InitEnvCleanup(Env env);
@@ -108,6 +109,7 @@ Object Init(Env env, Object exports) {
 #if (NAPI_VERSION > 2)
   exports.Set("callbackscope", InitCallbackScope(env));
 #endif
+  exports.Set("callbackInfo", InitCallbackInfo(env));
   exports.Set("dataview", InitDataView(env));
   exports.Set("dataview_read_write", InitDataView(env));
   exports.Set("dataview_read_write", InitDataViewReadWrite(env));
--- a/test/binding.gyp
+++ b/test/binding.gyp
@@ -17,6 +17,7 @@
         'basic_types/number.cc',
         'basic_types/value.cc',
         'bigint.cc',
+        'callbackInfo.cc',
         'date.cc',
         'binding.cc',
         'buffer.cc',
--- a/test/callbackInfo.cc
+++ b/test/callbackInfo.cc
@@ -1,0 +1,27 @@
+#include <assert.h>
+#include "napi.h"
+using namespace Napi;
+
+struct TestCBInfoSetData {
+  static void Test(napi_env env, napi_callback_info info) {
+    Napi::CallbackInfo cbInfo(env, info);
+    int valuePointer = 1220202;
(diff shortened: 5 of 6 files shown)


legendecas (approved):

mhdawson (approved):
LGTM

mhdawson:
@JckXia thanks! Landed in 55bd08ee26fded907a9e291c3a901e4b000c70f0

lovell:
There's a new compilation error with `sharp` that occurs on Windows only when using `node-addon-api` v5.1.0 that does not happen with v5.0.0, and this PR seems to be the most relevant change that might have triggered it.
```
D:\a\sharp\sharp\src\metadata.cc(279,32): error C2666: 'Napi::CallbackInfo::operator []': 2 overloads have similar conversions [D:\a\sharp\sharp\build\sharp-win32-ia32.vcxproj]
D:\a\sharp\sharp\node_modules\node-addon-api\napi-inl.h(3477,34): message : could be 'const Napi::Value Napi::CallbackInfo::operator [](size_t) const' (compiling source file ..\src\metadata.cc) [D:\a\sharp\sharp\build\sharp-win32-ia32.vcxproj]
D:\a\sharp\sharp\src\metadata.cc(279,32): message : or       'built-in C++ operator[(napi_callback_info, int)' [D:\a\sharp\sharp\build\sharp-win32-ia32.vcxproj]
D:\a\sharp\sharp\src\metadata.cc(279,32): message : while trying to match the argument list '(const Napi::CallbackInfo, int)' [D:\a\sharp\sharp\build\sharp-win32-ia32.vcxproj]
```

https://github.com/lovell/sharp/actions/runs/3924810417/jobs/6709314064

The code that triggers this is relatively straightforward (and hasn't changed):

```cc
Napi::Value metadata(const Napi::CallbackInfo& info) {
  ...
  Napi::Object options = info[0].As<Napi::Object>(); // compilation error occurs here
```

https://github.com/lovell/sharp/blob/flow/src/metadata.cc#L279

I'm still investigating this, but can anyone spot anything wrong with the way `sharp` is using `node-addon-api` here that this change might have uncovered?

JckXia:
Hi @lovell, I believe this particular issue is indeed related to this PR. I dug a little around C2666  [C2666 errors] (https://stackoverflow.com/questions/1726740/c-error-operator-2-overloads-have-similar-conversions) on Stackoverflow. 

From what I understand, by introducing this conversion operator for `napi_callback_info`, the MSVC compiler needs to choose between `info.operator napi_callback_info()[int]` (built-in), and `info.operator[](size_t)` (custom operator overload).   In either variants, 1 step of implicit conversion is required. In the first variant, we need to implicitly convert `Napi::CallbackInfo` into `napi_callback_info`, and in the second variant, we need to convert the literal `0` , which is an `int` into an `size_t`, thus the ambiguity.  A workaround could be to cast the literal into a "size_t" before passing it to info, but I think this is a bug that warrants a change from our side.

JckXia:
Yep I agree. We have a windows ci pipeline running but for some reason the checks wasn't able to catch this. Will bring this issue up in our weekly meetings.

tobias-klein:
I am getting this exact error in the latest [Windows CI builds of node-sword-interface](https://github.com/ezra-bible-app/node-sword-interface/pull/39).

mhdawson:
@JckXia do you have any suggestions for a good way to fix the issue? Trying to understand if we should think about backing out the change (although that would potentially be SemVer major as well)

JckXia:
@mhdawson I think there might be a few options we can explore. 

1. Remove the implicit conversion operator and declare the removal change a SemVer-major. Then, instead of using an implicit conversion operator, we can create an accessor function like `GetCbInfo`. This way people that need the raw `napi_callback_info` (hopefully not many since it's a fairly new change) still have a means of accessing that pointer.  We also get to keep the existing API for `[]` overload, since a LOT of people (including us) are passing literals into `info`.  The downside is that it seems to break conventions as we have been using implicit conversion operators to translate C++ NAPI objects to C node-API objects. But looking at  `std::string`, there is an explicit getter `c_str()` function to acquire the C string equivalent makes me think if we should reconsider our approach.

2. Change the parameter type of the `CallbackInfo::operator[](size_t index)` to be `int` instead.  I believe this way we should be able to avoid the ambiguity in question. But I don't know if this will be a breaking change for anyone because I don't understand all the nuances between `size_t` and `int` other than that `size_t` can potentially represent a greater range of values.

3. Remove the implicit conversion operator completely, but then people who might need the raw `napi_callback_info` is now out of luck if they had already written code around being able to access this pointer. 

On a side note, I couldn't reproduce this error locally using MSVC 2019  `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\MSBuild.exe`. Don't really know why that is. Going to take a deeper dive tomorrow.

mhdawson:
@JckXia thanks for the outline. I don't know the nuances of the implicit conversions well enough to know of the top of my head but my first thought might be that changing from size_t to int could affect people who had declared index as a size_t.

Let us know what you figure out in the deeper dive.

Issue #990, opened 2021-05-02
[Tests] add documentation for handleScope class and EscapableHandleScope class


JckXia:
| class                   | methods                                                                                                                          |
|-------------------------|----------------------------------------------------------------------------------------------------------------------------------|
 |HandleScope |     |
 |Covered #250|HandleScope(Napi::Env env) |
 ||HandleScope(napi_env env, napi_handle_scope scope) |
 | |Env() | 
 |EscapableHandleScope  |  |
 | |EscapableHandleScope(napi_env env, napi_escapable_handle_scope scope) |
 |Covered #250 |EscapableHandleScope(Napi::Env env) |
 |Covered #250 |Value Escape(napi_value escape) |
 | | Env() |

JckXia:
Completed following #1263

PR #1263, opened 2023-01-04
test: Add tests for escape and handlescope

JckXia:
Adding test coverage for `Escape` and `HandleScope`, closes https://github.com/nodejs/node-addon-api/issues/990

### DIFF
--- a/test/handlescope.cc
+++ b/test/handlescope.cc
@@ -13,11 +13,33 @@ Value createScope(const CallbackInfo& info) {
   return String::New(info.Env(), "scope");
 }
 
+Value createScopeFromExisting(const CallbackInfo& info) {
+  {
+    napi_handle_scope scope;
+    napi_open_handle_scope(info.Env(), &scope);
+    HandleScope scope_existing(info.Env(), scope);
+    String::New(scope_existing.Env(), "inner-existing-scope");
+  }
+  return String::New(info.Env(), "existing_scope");
+}
+
 Value escapeFromScope(const CallbackInfo& info) {
   Value result;
   {
     EscapableHandleScope scope(info.Env());
     result = scope.Escape(String::New(info.Env(), "inner-scope"));
+  }
+  return result;
+}
+
+Value escapeFromExistingScope(const CallbackInfo& info) {
+  Value result;
+  {
+    napi_escapable_handle_scope scope;
+    napi_open_escapable_handle_scope(info.Env(), &scope);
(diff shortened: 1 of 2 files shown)


test/handlescope.cc
```diff
@@ -13,6 +13,16 @@ Value createScope(const CallbackInfo& info) {
   return String::New(info.Env(), "scope");
 }
 
+Value createScopeFromExisting(const CallbackInfo& info) {
+  {
+    napi_handle_scope scope;
+    napi_open_handle_scope(info.Env(), &scope);
```
mhdawson, line 19: Could you have used a second HandleScope declaration here instead of making the node-api call directly?  I think it would then look like

HandleScope scope_one(info.Env())
HandleScope scope_two(info.Env, scope_one.napi_handle_scope())

JckXia: I think I've tried that approach before, but ran into some issues with `~HandleScope` being called on the same underly `napi_handle_scope` twice when they go out of scope.

mhdawson: @JckXia if that's the case then I think we might have a real problem?

JckXia: Hmm, I believe the `HandleScope`constructor where we provide an existing `napi_handle_scope` simply saves a reference of the `napi_handle_scope` passed in, as opposed to something like a move constructor where we set the `other` to null. 

I think this means when `scope_two` goes out of scope, its dtor gets ran,  we successfully close the referenced handle scope. But when `scope_one` goes out of scope, its dtor will attempt to call `napi_close_handle_scope` on what I assume is an already closed handle scope, resulting in a `napi_handle_scope_mismatch` error. 

The reason I created the `napi_handle_scope` via a direct call to node-api is such that only scope_two's dtor will be ran, closing it successfully.

mhdawson:
I looked a little more carefully and I understand  now. It's not that we have two scopes where one is beeing used in the second. Instead the API is to create an node-addon-api object from an existing node-api C handle.  In that case creating the handle through a node-addon-api object would result in a double free since there would be two node-addon-api objects pointing to the same node-api C handle.

Based on that I think this looks good :)

mhdawson (approved):
LGTM

mhdawson:
Landed as 35d9d669b37c04130df6d3d9027ad9bad0a60095

JckXia:
Thanks! @mhdawson :)

Issue #1046, opened 2021-08-22
[Test] Unit test coverage report for Maybe<T>


JckXia:
| class                   | methods                                                                                                                          |
|-------------------------|----------------------------------------------------------------------------------------------------------------------------------|                           
 |Maybe |     |
 ||  IsNothing() |
 |Covered|  IsJust() |
 |Covered|  void Check() |
 |Defined as a macro which is used by most tests|T UnWrap() |
 |Defined as a macro, which is used by most tests|T UnWrapOr(const T&default_value) const |
 |Defined as a macro which is used by most tests |bool UnwrapTo(T* out) const |
 ||bool operator==(const Maybe& other) const |
 ||bbool operator!=(const Maybe& other) const |
 

See test/maybe/check.cc and test/maybe/index.js

JckXia:
Closed following #1270

PR #1270, opened 2023-01-13
test: Add test covg for Maybe<T>

JckXia:
Add test coverage for https://github.com/nodejs/node-addon-api/issues/1046

### DIFF
--- a/test/maybe/check.cc
+++ b/test/maybe/check.cc
@@ -8,20 +8,20 @@ namespace {
 
 void VoidCallback(const CallbackInfo& info) {
   Napi::Function fn = info[0].As<Function>();
-  Maybe<Value> fn_m = fn.Call({});
+  Maybe<Value> ret = fn.Call({});
 
-  assert(fn_m.IsNothing() == true);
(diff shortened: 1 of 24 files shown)


test/maybe/check.cc
```diff
@@ -6,18 +7,63 @@ using namespace Napi;
 namespace {
 
 void VoidCallback(const CallbackInfo& info) {
-  Function fn = info[0].As<Function>();
+  Napi::Function fn = info[0].As<Function>();
+  Maybe<Value> fn_m = fn.Call({});
```
legendecas, line 11: This variable represents the return value of the fn invocation. Might be worth to name it with prefix like `ret`, rather than `fn`.

```suggestion
  Maybe<Value> ret_m = fn.Call({});
```

legendecas (approved):

mhdawson (approved):
LGTM

test/maybe/check.cc
JckXia: Updated the PR to fix this.

Issue #1257, opened 2022-12-29
Provide a new `Buffer::NewOrCopy` method


Julusian:
Background:
* https://github.com/nodejs/abi-stable-node/issues/441
* https://github.com/nodejs/node/pull/45181
* https://github.com/electron/electron/issues/35801

After https://github.com/nodejs/node/pull/45181 it is possible for `Buffer::new` to fail with `napi_no_external_buffers_allowed`. A helper method `Buffer::NewOrCopy` which would try to do a `Buffer::new` and fallback to doing a `Buffer::copy` would be helpful and encourage users to code in a way which provides compatibility for both electron and nodejs, without them having to implement the method themselves.

The npm package sharp has implemented this themselves already https://github.com/lovell/sharp/commit/584807b4f5e1b831785c79eef5367f737589f29b#diff-53a01d31aa585021f8f316c6613dadef86bf1f9ac2070d52daa68adbefbb6ef4.

mhdawson:
We discussed in the team meeting today, and we agreed it would make sense to add. @legendecas volunteered to put together a PR.

legendecas:
Landed in https://github.com/nodejs/node-addon-api/pull/1273.

PR #1273, opened 2023-01-16
src: handle no support for external buffers

legendecas:
Fixes https://github.com/nodejs/node-addon-api/issues/1257.

- When `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` introduced in https://github.com/nodejs/node/pull/45181 is defined, hides the methods that can create external buffers. 
- Introduce `Napi::Buffer::NewOrCopy` to create buffer from external buffers that are compatible with runtimes like electron conveniently.

### DIFF
--- a/doc/buffer.md
+++ b/doc/buffer.md
@@ -22,6 +22,9 @@ Returns a new `Napi::Buffer` object.
 
 ### New
 
+> When `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` is defined, this method is not available.
+> See [External Buffer][] for more information.
+
 Wraps the provided external data into a new `Napi::Buffer` object.
 
 The `Napi::Buffer` object does not assume ownership for the data and expects it to be
@@ -46,6 +49,9 @@ static Napi::Buffer<T> Napi::Buffer::New(napi_env env, T* data, size_t length);
 Returns a new `Napi::Buffer` object.
 
 ### New
+
+> When `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` is defined, this method is not available.
+> See [External Buffer][] for more information.
 
 Wraps the provided external data into a new `Napi::Buffer` object.
 
@@ -72,6 +78,9 @@ Returns a new `Napi::Buffer` object.
 
 ### New
 
+> When `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` is defined, this method is not available.
+> See [External Buffer][] for more information.
+
 Wraps the provided external data into a new `Napi::Buffer` object.
 
 The `Napi::Buffer` object does not assume ownership for the data and expects it to be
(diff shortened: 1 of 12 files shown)


lovell:
Thank you for adding this feature. It looks like the approach in this PR can (conditionally) remove the presence of `Buffer<T>::New()` from the API at compile time. Is this intentional? If so, please can this be considered a breaking, semver major change.

legendecas:
@lovell it is intentional and `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` is following the definition in https://github.com/nodejs/node/pull/45181. I don't think it is a breaking change, since user codes will not be affected if the newly introduced macro `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` is not defined.

doc/buffer.md
```diff
@@ -98,6 +107,93 @@ static Napi::Buffer<T> Napi::Buffer::New(napi_env env,
 
 Returns a new `Napi::Buffer` object.
 
+### NewOrCopy
+
+Wraps the provided external data into a new `Napi::Buffer` object. When the
+[external buffer][] is not supported, allocates a new `Napi::Buffer` object and
```
mhdawson, line 113:
```suggestion
[external buffer][] is not supported by the underlying runtime, allocates a new `Napi::Buffer` object and
```

mhdawson (approved):
LGTM other than the suggested wording change to the docs

Julusian:
It looks like when `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` is set, that `Buffer::NewOrCopy` is being defined, with the implementation attempting to use `napi_create_external_buffer` which is missing because of the `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` flag.

When `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` either `Buffer::NewOrCopy` should also be disabled, or the implementation should skip calling `napi_create_external_buffer` and always do the copy?

Ideally I would like to be able to use `Buffer::NewOrCopy` which is able to do `napi_create_external_buffer`, but to also have the offending `Buffer::New` methods removed. That way I can't accidentally use `Buffer::New`, but I can still get the benefits of `Buffer::NewOrCopy` when the runtime allows. I don't know how this could be implemented though

legendecas:
@Julusian thanks for the suggestion! I'll update the PR to include a test to verify `Buffer::NewOrCopy ` is still usable when `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` is defined.

legendecas:
Seems like the test `async_progress_worker` is constantly failing on windows-2019, v16.x.

napi-inl.h
```diff
@@ -2491,6 +2498,94 @@ inline Buffer<T> Buffer<T>::New(napi_env env,
...
+    delete finalizeData;
+    NAPI_THROW_IF_FAILED(env, status, Buffer());
+  }
+  return Buffer(env, value, length, data);
+#endif  // NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED
+}
+
+template <typename T>
+template <typename Finalizer, typename Hint>
+inline Buffer<T> Buffer<T>::NewOrCopy(napi_env env,
```
mhdawson, line 2556: I assume its the templates that prevent us from having one function with all the parameters and the others calling that one with dummy/null parameters, right?

mhdawson (approved):
LGTM on updated version with one question

napi-inl.h
legendecas: Not really. When the template parameter `Hint` is set, the `Finalizer` must accept a third parameter `Hint` too. However, we allow `Finalizer` to be like `void operator()(Env env, T* data)`. So we have to distinguish if `Hint` is present.

legendecas:
@mhdawson is this still looking good to you with the replied question? 

If so, I think we've discussed about the failure of `Node.js CI Windows Platform / test (16.x, x86, windows-2019) (pull_request)` and we can suppress it at the moment and land this PR.

mhdawson:
@legendecas LGTM
Landed as b16c762a190ced903045aa15fbb910eee5c04a97

PR #1275, opened 2023-01-24
Make CallbackInfo cast to napi_callback_info explicit

KevinEady:
Fixes: #1272

Issue #1272, opened 2023-01-16
Possible breaking change in node-addon-api 5.1.0


JckXia:
Hello folks, it looks like we had some compilation issues related to the introduction of the `napi_callback_info` implicit conversion operator. [Issue link](https://github.com/nodejs/node-addon-api/pull/1253#issuecomment-1383239864).  Not too sure why this wasn't caught by our windows CI.

JckXia:
(Just going to move our discussion over here for better visibility)

@mhdawson I think there might be a few options we can explore.

Remove the implicit conversion operator and declare the removal change a SemVer-major. Then, instead of using an implicit conversion operator, we can create an accessor function like GetCbInfo. This way people that need the raw napi_callback_info (hopefully not many since it's a fairly new change) still have a means of accessing that pointer. We also get to keep the existing API for [] overload, since a LOT of people (including us) are passing literals into info. The downside is that it seems to break conventions as we have been using implicit conversion operators to translate C++ NAPI objects to C node-API objects. But looking at std::string, there is an explicit getter c_str() function to acquire the C string equivalent makes me think if we should reconsider our approach.

Change the parameter type of the CallbackInfo::operator[](size_t index) to be int instead. I believe this way we should be able to avoid the ambiguity in question. But I don't know if this will be a breaking change for anyone because I don't understand all the nuances between size_t and int other than that size_t can potentially represent a greater range of values.

Remove the implicit conversion operator completely, but then people who might need the raw napi_callback_info is now out of luck if they had already written code around being able to access this pointer.

On a side note, I couldn't reproduce this error locally using MSVC 2019 C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\MSBuild.exe. Don't really know why that is. Going to take a deeper dive tomorrow.

KevinEady:
Could we make the conversion function an explicit conversion only?

```cpp
explicit operator napi_callback_info() const;
```

Then the user would need to use a `static_cast<napi_callback_info>( cbInfo )` where `cbInfo` is an `Napi::CallbackInfo` object. A little more wordy but might work?

KevinEady:
[...]

PR #1276, opened 2023-01-25
chore: Build node-addon-api against X86

JckXia:
To be merged after #1275 is merged in. Update our windows CI to build node against X86 based on the [observation](https://github.com/nodejs/node-addon-api/issues/1272#issuecomment-1396389182) here. Main source of concern is that this will double the number of runs we will perform for windows CI. But It might be a worthy trade off to avoid issues like https://github.com/nodejs/node-addon-api/issues/1272

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -22,6 +22,7 @@ jobs:
       uses: actions/setup-node@v3
       with:
         node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


mhdawson (approved):
LGTM

mhdawson:
@JckXia does this need a rebase, it still seems to be failing after 1275 landed.

mhdawson:
landed in ff969485ea86b0e2af1bc5ae673eecf6a86593dc

Issue #981, opened 2021-04-27
[Tests] Add test coverage document for typed error


JckXia:
| class                   | methods                                                                                                                          |
|-------------------------|----------------------------------------------------------------------------------------------------------------------------------|
| TypeError                | TypeError()                                                                                                                      |
|                         | TypeError(napi_env env, napi_value value)                                                         |
| |  static TypeError New(napi_env env, const char* message); |
|Covered #24 |  static TypeError New(napi_env env, const std::string& message); |

Issue #1271, opened 2023-01-15
Question about `Napi::TypeError` ctor


JckXia:
Hello everyone, just a question regarding the expected behavior of `TypeError(napi_env env, napi_value value)`. According to the docs, it initializes a `Napi::TypeError` C++ instance from an existing error object, where `value` is a `Napi::Error`.

The first problem is that we do not have a cast operator overload that converts a `Napi::Error` object to a `napi_value` , so we can't call this ctor with a `Napi::Error` object directly.  

Second of all, my assumption was this ctor  accepts an error object and creates a JS `TypeError`  in V8 and a  C++ stack object. However, it looks like we simply call the `Error` ctor with the  `napi_value`. For example, If we give the ctor a `napi_value` passed from JS, where the `napi_value` is a `RangeError`, the resulting `Napi::TypedError`, when `ThrowAsJavaScript`, will yield an RangeError still. 

So my question is, what is the correct interpretation of this particular API, and what should we do to tweak the implementation to comply with that spec? Thank you!

gabrielschulhof:
I think this is a general problem with all our constructors of the form

```cpp
JSValueType::JSValueType(napi_env env, napi_value value): JSValueSuperClass(env, value) {}
```

the purpose of which is to wrap a value coming in from C. You can wrap any JS value using any of our classes, without validation.

In the next major version, we may wanna
1. add a `::Wrap(napi_env env, napi_value value)` static method to these classes, which performs validation, rejecting, for each class, a `napi_value` that does not have the corresponding JS type, and
2. restrict the visibility of the constructor.

Issue #984, opened 2021-04-28
[Tests] Add test coverage for RangeErrors


JckXia:
| class                   | methods                                                                                                                          |
|-------------------------|----------------------------------------------------------------------------------------------------------------------------------|
| RangeError | RangeError::New(Napi::Env env, const char * message)  |
|Covered #52 |RangeError::New(Napi::Env env, const std::string &message) |
| |RangeError() |
| |RangeError(napi_env env, napi_value value) |

JckXia:
Closed following #1280

PR #1280, opened 2023-02-01
[Draft] Add test covg for typed and range err

JckXia:
PR aims to close https://github.com/nodejs/node-addon-api/issues/981, https://github.com/nodejs/node-addon-api/issues/984 and https://github.com/nodejs/node-addon-api/issues/1271.

Adds test for `TypeError` and `RangeError`, as well as some validation logic for the `napi_value` object we pass to the base `TypeError` and `RangeError` ctor. The question is whether we want to crash the application when the `napi_value` passed isn't a js `TypeError`, or if we want to emit some type of warning instead.


legendecas:
I'd suggest avoiding type checks in type constructors for the following reasons:

1. There are error classes and we should avoid throwing in their constructors as they are likely to be used in error handling paths already.
2. We don't do type checks in node-addon-api type wrapper constructors, as it comes with runtime costs, and constructing from napi_value is a common path.

JckXia:
@legendecas  Thank you for your explanation! Will update my changes.

test/error.cc
```diff
@@ -270,7 +359,15 @@ Object InitError(Env env) {
       Function::New(env, LastExceptionErrorCode);
   exports["throwJSError"] = Function::New(env, ThrowJSError);
   exports["throwTypeError"] = Function::New(env, ThrowTypeError);
+  exports["throwTypeErrorCtor"] = Function::New(env, ThrowTypeErrorCtor);
+  exports["throwTypeErrorCStr"] = Function::New(env, ThrowTypeErrorCStr);
+  exports["throwTypeErrorNapiVal"] = Function::New(env, ThrowTypeErrorNapiVal);
```
legendecas, line 364: Seems like `*NapiVal` methods are not called in the test.

JckXia: Yep. Took these out as that overload was tested by https://github.com/nodejs/node-addon-api/pull/1280/files#diff-ac24d807b22ae295c2dcf786090d933b62aa2b6b3c85c16e839d31ad35fc2c13R81

legendecas (approved):

Issue #1246, opened 2022-12-05
Inconsistent return value with enabled NODE_ADDON_API_ENABLE_MAYBE directive


ikokostya:
If preprocessor directives `NAPI_DISABLE_CPP_EXCEPTIONS` and `NODE_ADDON_API_ENABLE_MAYBE` are enabled, `Napi` API calls should return `Maybe` type https://github.com/nodejs/node-addon-api/blob/HEAD/doc/error_handling.md#handling-errors-with-maybe-type-and-c-exceptions-disabled. 

But many Napi methods violate this rule. For example all `Napi::Number` conversion methods (`Int32Value()`, `Uint32Value()`, `FloatValue()`, `DoubleValue()`) and `Napi::BigInt` conversion methods return unwrapped value as is:

https://github.com/nodejs/node-addon-api/blob/39267baf1ba985275f74991fef6faac2ccf8e495/napi-inl.h#L829-L834

https://github.com/nodejs/node-addon-api/blob/39267baf1ba985275f74991fef6faac2ccf8e495/napi.h#L90-L94

In the following example variable `x` will have `0` value if `info[0]` is not a number:

```cpp
void Foo(const Napi::CallbackInfo& info) {
  auto x = info[0].As<Napi::Number>().Int32Value(); // expected x type is Maybe<int32_t>, 
                                                    // actual type is int32_t
}
```

mhdawson:
We discussed in the team meeting today and originally these methods were intentionally left out of those that returns Maybes. @legendecas will take another look and comment.

mhdawson:
@vmoroz suggested we make the macro NAPI_ENABLE_AS_TYPE_CHECK

mhdawson:
From follow on discussion NAPI_ENABLE_TYPE_CHECK_ON_AS was suggested

PR #1281, opened 2023-02-02
src: enforce type checks on Napi::Value::As()

legendecas:
Fixes https://github.com/nodejs/node-addon-api/issues/1246


legendecas:
Not sure why windows builds failed to fetch arm64 `node.lib`. Needs to retry after a while to see if the problem persists.

legendecas:
Windows tests passed! This PR is ready for review.

mhdawson (approved):
LGTM

PR #1306, opened 2023-04-05
Fix exits/exists typo in docs for Env::AddCleanupHook()

RedBeard0531:
I'm 99% sure that this was a typo that was then copy-pasted, but if not, please close.


legendecas (approved):
Thanks for catching this!

mhdawson (approved):
LGTM

Issue #1041, opened 2021-08-19
[Tests] AsyncProgressWorker unit test coverage


JckXia:
| class                   | methods                                                                                                                          |
|-------------------------|----------------------------------------------------------------------------------------------------------------------------------|                           
 |AsyncProgressWorker |     |
 ||  AsyncProgressWorker(const Function& callback); |
 ||  AsyncProgressWorker(const Function& callback, const char* resource_name);|
 |Covered |  AsyncProgressWorker(const Function& callback,const char* resource_name, const Object& resource)|
| |  AsyncProgressWorker(const Object& receiver, const Function& cb)|
| | AsyncProgressWorker(const Object& receiver,const Function& cb, const char* resource_name) |      
|| AsyncProgressWorker(const Object& receiver,const Function& cb, const char* resource_name, const Object& resource) |
| |AsyncProgressWorker(Napi::Env env) |
| |AsyncProgressWorker(Napi::Env env, const char* resource_name) |
| |AsyncProgressWorker(Napi::Env env, const char* resource_name, const Object& resource) | 
|Covered | void Execute(const ExecutionProgress& progress)|
|Covered |void OnProgress(const T* data, size_t count) |
 
  
See asyncprogressworker.cc and asyncprogressworker.js

PR #1307, opened 2023-04-07
[test] Add test coverage for AsyncProgressWorker

JckXia:
Resolves https://github.com/nodejs/node-addon-api/issues/1041, and refactors async hook installation into the common module


test/async_progress_worker.js
```diff
@@ -4,12 +4,146 @@ const common = require('./common');
...
+let asyncHooks;
+function checkAsyncHooks () {
+  if (nodeVersion >= 8) {
+    if (asyncHooks === undefined) {
+      asyncHooks = require('async_hooks');
+    }
+    return true;
+  }
+  return false;
+}
```
gabrielschulhof, line 18: According to the async hooks [doc](https://nodejs.org/api/async_hooks.html) their use is discouraged. Can we use the alternatives mentioned there?

JckXia: Hmm.. From what I understand, we use `async_hooks` here to verify that async resources/objects passed from the C++ side are successfully instantiated, and can be used to track the life cycle of the async call from JS. 

I took a look at `AsyncLocalStorage`, and it doesn't seem to provide any API to inspect the async resources/objects associated with an async call. @legendecas Any ideas?

gabrielschulhof: OK. That's fair.

legendecas (approved):

PR #1313, opened 2023-04-26
src: Avoid calling into C++ with a null this

chearon:
These tests all crash the process because `instance` is null when exceptions are off. If ObjectWrap methods try to access anything on `this`, it's bad news. I don't think the method should get called at all in this case since there's a pending JS exception anyways.

Context: I recently [ported node-canvas](https://github.com/Automattic/node-canvas/pull/2235) to `node-addon-api` and we have tests for this. It was a pain in v8/NAN, and we could remove a lot of code with these changes. I had a positive experience working with `node-addon-api`, great work everyone!

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -165,7 +165,7 @@ napi_value TemplatedInstanceCallback(napi_env env,
   return details::WrapCallback([&] {
     CallbackInfo cbInfo(env, info);
     T* instance = T::Unwrap(cbInfo.This().As<Object>());
-    return (instance->*UnwrapCallback)(cbInfo);
+    return instance ? (instance->*UnwrapCallback)(cbInfo) : Napi::Value();
   });
 }
 
@@ -175,7 +175,7 @@ napi_value TemplatedInstanceVoidCallback(napi_env env, napi_callback_info info)
   return details::WrapCallback([&] {
     CallbackInfo cbInfo(env, info);
     T* instance = T::Unwrap(cbInfo.This().As<Object>());
-    (instance->*UnwrapCallback)(cbInfo);
+    if (instance) (instance->*UnwrapCallback)(cbInfo);
     return nullptr;
   });
 }
@@ -4340,7 +4340,7 @@ inline napi_value InstanceWrap<T>::InstanceVoidMethodCallbackWrapper(
     callbackInfo.SetData(callbackData->data);
     T* instance = T::Unwrap(callbackInfo.This().As<Object>());
     auto cb = callbackData->callback;
-    (instance->*cb)(callbackInfo);
+    if (instance) (instance->*cb)(callbackInfo);
     return nullptr;
   });
 }
@@ -4355,7 +4355,7 @@ inline napi_value InstanceWrap<T>::InstanceMethodCallbackWrapper(
     callbackInfo.SetData(callbackData->data);
     T* instance = T::Unwrap(callbackInfo.This().As<Object>());
     auto cb = callbackData->callback;
-    return (instance->*cb)(callbackInfo);
+    return instance ? (instance->*cb)(callbackInfo) : Napi::Value();
   });
 }
 
@@ -4369,7 +4369,7 @@ inline napi_value InstanceWrap<T>::InstanceGetterCallbackWrapper(
     callbackInfo.SetData(callbackData->data);
     T* instance = T::Unwrap(callbackInfo.This().As<Object>());
     auto cb = callbackData->getterCallback;
-    return (instance->*cb)(callbackInfo);
+    return instance ? (instance->*cb)(callbackInfo) : Napi::Value();
   });
 }
 
@@ -4383,7 +4383,7 @@ inline napi_value InstanceWrap<T>::InstanceSetterCallbackWrapper(
     callbackInfo.SetData(callbackData->data);
     T* instance = T::Unwrap(callbackInfo.This().As<Object>());
     auto cb = callbackData->setterCallback;
-    (instance->*cb)(callbackInfo, callbackInfo[0]);
+    if (instance) (instance->*cb)(callbackInfo, callbackInfo[0]);
     return nullptr;
   });
 }
@@ -4395,7 +4395,7 @@ inline napi_value InstanceWrap<T>::WrappedMethod(
   return details::WrapCallback([&] {
     const CallbackInfo cbInfo(env, info);
     T* instance = T::Unwrap(cbInfo.This().As<Object>());
-    (instance->*method)(cbInfo, cbInfo[0]);
+    if (instance) (instance->*method)(cbInfo, cbInfo[0]);
     return nullptr;
   });
 }
--- a/test/objectwrap.js
+++ b/test/objectwrap.js
@@ -24,6 +24,9 @@ async function test (binding) {
       obj.testSetter = 'instance getter 2';
       assert.strictEqual(obj.testGetter, 'instance getter 2');
       assert.strictEqual(obj.testGetterT, 'instance getter 2');
+
+      assert.throws(() => clazz.prototype.testGetter, /Invalid argument/);
+      assert.throws(() => clazz.prototype.testGetterT, /Invalid argument/);
     }
 
     // read write-only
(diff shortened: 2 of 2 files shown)


test/objectwrap.js
```diff
@@ -98,6 +106,9 @@ async function test (binding) {
     assert.strictEqual(obj.testMethodT(), 'method<>(const char*)');
     obj[clazz.kTestVoidMethodTInternal]('method<>(Symbol)');
     assert.strictEqual(obj[clazz.kTestMethodTInternal](), 'method<>(Symbol)');
+    try { clazz.prototype.testMethod('method'); } catch (e) { }
+    try { clazz.prototype.testMethodT(); } catch (e) { }
+    try { clazz.prototype.testVoidMethodT('method<>(const char*)'); } catch (e) { }
```
gabrielschulhof, line 111: V8 actually protects against these with "Invalid invocation". IOW, it does not allow one to invoke a native instance method where the JavaScript `this` is not an instance of the class. Thus, these checks are unnecessary. Interesting though that it allows property accesses to go through 🤨

test/objectwrap.js
```diff
@@ -24,6 +24,11 @@ async function test (binding) {
       obj.testSetter = 'instance getter 2';
       assert.strictEqual(obj.testGetter, 'instance getter 2');
       assert.strictEqual(obj.testGetterT, 'instance getter 2');
+
+      // eslint-disable-next-line no-unused-expressions
+      try { clazz.prototype.testGetter; } catch (e) { }
```
gabrielschulhof, line 29:
```suggestion
      assert.throws(() => clazz.prototype.testGetter, /Invalid argument/);
```

test/objectwrap.js
```diff
@@ -24,6 +24,11 @@ async function test (binding) {
...
+      // eslint-disable-next-line no-unused-expressions
+      try { clazz.prototype.testGetterT; } catch (e) { }
```
gabrielschulhof, line 31:
```suggestion
      assert.throws(() => clazz.prototype.testGetterT, /Invalid argument/);
```

test/objectwrap.js
```diff
@@ -61,6 +66,9 @@ async function test (binding) {
 
       obj.testGetSetT = 'instance getset 4';
       assert.strictEqual(obj.testGetSetT, 'instance getset 4');
+
+      try { clazz.prototype.testGetSet = 'instance getset'; } catch (e) { }
+      try { clazz.prototype.testGetSetT = 'instance getset'; } catch (e) { }
```
gabrielschulhof, line 71:
```suggestion
      assert.throws(() => clazz.prototype.testGetSet, /Invalid argument/);
      assert.throws(() => clazz.prototype.testGetSetT, /Invalid argument/);
```

napi-inl.h
```diff
@@ -165,7 +165,7 @@ napi_value TemplatedInstanceCallback(napi_env env,
   return details::WrapCallback([&] {
     CallbackInfo cbInfo(env, info);
     T* instance = T::Unwrap(cbInfo.This().As<Object>());
-    return (instance->*UnwrapCallback)(cbInfo);
+    return instance ? (instance->*UnwrapCallback)(cbInfo) : nullptr;
```
gabrielschulhof, line 168:
```suggestion
    return instance ? (instance->*UnwrapCallback)(cbInfo) : Napi::Value();
```

napi-inl.h
```diff
@@ -4355,7 +4355,7 @@ inline napi_value InstanceWrap<T>::InstanceMethodCallbackWrapper(
     callbackInfo.SetData(callbackData->data);
     T* instance = T::Unwrap(callbackInfo.This().As<Object>());
     auto cb = callbackData->callback;
-    return (instance->*cb)(callbackInfo);
+    return instance ? (instance->*cb)(callbackInfo) : nullptr;
```
gabrielschulhof, line 4358:
```suggestion
    return instance ? (instance->*cb)(callbackInfo) : Napi::Value();
```

napi-inl.h
```diff
@@ -4369,7 +4369,7 @@ inline napi_value InstanceWrap<T>::InstanceGetterCallbackWrapper(
     callbackInfo.SetData(callbackData->data);
     T* instance = T::Unwrap(callbackInfo.This().As<Object>());
     auto cb = callbackData->getterCallback;
-    return (instance->*cb)(callbackInfo);
+    return instance ? (instance->*cb)(callbackInfo) : nullptr;
```
gabrielschulhof, line 4372:
```suggestion
    return instance ? (instance->*cb)(callbackInfo) : Nap::Value();
```

gabrielschulhof (commented):
Looking good. Just a few minor comments.

test/objectwrap.js
chearon: Not sure about this one. Don't you want to test the setter? `prototype.testGetSet = 'instance get set'` does actually call the setter with `prototype` as the `this` instead of the constructed object.

```js
class Abc { set v(v) { console.log(this instanceof Abc); } }
const abc = new Abc();
abc.v = 2;
// log: true
Abc.prototype.v = 2;
// log: false
```

chearon:
Thanks for the review! Think I got everything except for the one I commented on.

test/objectwrap.js
legendecas: The suggestion here might be:

```suggestion
      assert.throws(() => { clazz.prototype.testGetSet = 'instance getset'; }, /Invalid argument/);
      assert.throws(() => { clazz.prototype.testGetSetT = 'instance getset'; }, /Invalid argument/);
```

legendecas (approved):

test/objectwrap.js
gabrielschulhof: @legendecas @chearon yes, that's what I meant. The important thing is that we use `assert.throws()` and that we check for the right exception with the regex.

gabrielschulhof (approved):

mhdawson:
Landed in 858942ce31e9324b9c57fcd5854f9e4f286e075b

napi-inl.h
kraenhansen: I would have liked this to throw instead of returning `undefined`.

I'd assume this can only happen if the `InstanceWrap` is setup or called incorrectly (i.e. doesn't have a its instance data set) and I therefore expected it to throw the same way other native APIs throw when I call an unbound member function:

```
> const b = Buffer.alloc(1);
undefined
> const copy = b.copy;
undefined
> copy();
Uncaught:
TypeError [ERR_INVALID_ARG_TYPE]: The "source" argument must be an instance of Buffer or Uint8Array. Received undefined
```

kraenhansen:
I'm posting here to keep the context of the PR and get some early feedback on my thoughts.
I'll happily create a new issue if I'm not overlooking something obvious.

I believe the `instance` checks in [`InstanceMethodCallbackWrapper`](https://github.com/nodejs/node-addon-api/pull/1313/files#diff-cd0b229bcaf532206661e2ee1a603a7f60ac22ecfcb239920f6b42a34d909f34R4358) and [`InstanceGetterCallbackWrapper`](https://github.com/nodejs/node-addon-api/pull/1313/files#diff-cd0b229bcaf532206661e2ee1a603a7f60ac22ecfcb239920f6b42a34d909f34R4358) should throw instead of returning `undefined`.

To me, it seems that the tests are relying on the `ObjectWrap<T>::Unwrap` to throw if failed while `napi_unwrap`'s implementation indeed fails [if the "local value" is not an object](https://github.com/nodejs/node/blob/804d41f9c73c03e670d8e788d14a2237c8c2a057/src/js_native_api_v8.cc#L350).

This is however not the case for `Addon` declaring its methods via `InstanceWrap<T>::InstanceMethod`: When the Addon's `Unwrap` function is called

https://github.com/nodejs/node-addon-api/blob/7c3226fe43d857c0440c9b46ccd9f52d1d533347/napi-inl.h#L4576

which simply defers to `GetInstanceData`
https://github.com/nodejs/node-addon-api/blob/7c3226fe43d857c0440c9b46ccd9f52d1d533347/napi-inl.h#L6795-L6798

Which does report a fatal error if the status indicates a failure

https://github.com/nodejs/node-addon-api/blob/7c3226fe43d857c0440c9b46ccd9f52d1d533347/napi-inl.h#L742-L744

but the call to `napi_get_instance_data` will simply pass through `nullptr` values: https://github.com/nodejs/node/blob/804d41f9c73c03e670d8e788d14a2237c8c2a057/src/js_native_api_v8.cc#L3527

PR #1326, opened 2023-06-10
src: add SyntaxError

KevinEady:
Adds wrapper for `Napi::SyntaxError` from release of Node-API v9.

### DIFF
--- a/doc/syntax_error.md
+++ b/doc/syntax_error.md
@@ -1,0 +1,66 @@
+# SyntaxError
+
+The `Napi::SyntaxError` class is a representation of the JavaScript
+`SyntaxError` that is thrown when the engine encounters tokens or token order
+that does not conform to the syntax of the language when parsing code.
+
+The `Napi::SyntaxError` class inherits its behaviors from the `Napi::Error`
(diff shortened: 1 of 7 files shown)


KevinEady:
|Node Version|Link|Status|
|-|-|-|
|20|https://ci.nodejs.org/job/node-test-node-addon-api-new/7558/||
|21|https://ci.nodejs.org/job/node-test-node-addon-api-new/7559/||

doc/syntax_error.md
```diff
@@ -0,0 +1,66 @@
+# SyntaxError
+
```
mhdawson, line 2: Should it say here that it's only supported in node-api version 9 and higher?

mhdawson: That might be a gap though ?  I go to try to use an API and it's not there?  Either way agree that we don't need to be block this PR on it, but might be worth thinking about wether adding that info makes sense.

legendecas (approved):
LGTM % one nit.

KevinEady:
Added `do not land` label until release of v7.0.0 is complete.

PR #1327, opened 2023-06-10
src: add Env::GetModuleFileName

KevinEady:
Adds wrapper for `Napi::Env::GetModuleFileName()` from release of Node-API v9.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -572,6 +572,14 @@ void Env::DefaultFiniWithHint(Env, DataType* data, HintType*) {
 }
 #endif  // NAPI_VERSION > 5
 
+#if NAPI_VERSION > 8
+inline std::string_view Env::GetModuleFileName() const {
+  const char* result;
+  napi_status status = node_api_get_module_file_name(_env, &result);
+  NAPI_THROW_IF_FAILED(*this, status, std::string_view());
+  return std::string_view(result);
+}
+#endif  // NAPI_VERSION > 8
 ////////////////////////////////////////////////////////////////////////////////
 // Value class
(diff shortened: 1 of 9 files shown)


KevinEady:
|Node Version|Link|Status|
|-|-|-|
|20|https://ci.nodejs.org/job/node-test-node-addon-api-new/7560/||
|21|https://ci.nodejs.org/job/node-test-node-addon-api-new/7561/||

napi-inl.h
```diff
@@ -572,6 +572,14 @@ void Env::DefaultFiniWithHint(Env, DataType* data, HintType*) {
 }
 #endif  // NAPI_VERSION > 5
 
+#if NAPI_VERSION > 8
+inline std::string_view Env::GetModuleFileName() const {
+  const char* result;
+  napi_status status = node_api_get_module_file_name(_env, &result);
```
legendecas, line 578: Node.js v16.x is still compiled with [C++14](https://github.com/nodejs/node/blob/v16.x/common.gypi#L405), however `std::string_view` is available since c++17. I think we should avoid using c++17 features here.

KevinEady: Mhmm mhmm that makes sense ... So then would we prefer the alternative of node-addon-api returning a new `std::string` by copying the `const char* result`?

legendecas: Would it be good to return the `const char* result`? If needed, copying the result with `std::string` is fairly easy.

gabrielschulhof (dismissed):

legendecas (approved):
Thanks

mhdawson (approved):
LGTM

mhdawson:
Landed in b83e453

PR #1340, opened 2023-06-16
Add Node.js version restrictions

### LINKED ISSUE #1339: Add engines limitation into package.json?
I just saw 7.0.0 and a changelog note about dropping Node.js 14 and 19 support ... I would have expected that this is also refelected in the package.json but there is no info at all.

Would it not make sense for less issues on developer and user side to have that in the package.json too?

Apollon77:
Assumed from last Changelog the current version supports Node.js 16, 18, 20 and we should also not block 21+ for now, Right?

fixes #1339

### DIFF
--- a/package.json
+++ b/package.json
@@ -468,5 +468,8 @@
   },
   "pre-commit": "lint",
   "version": "7.0.0",
-  "support": true
+  "support": true,
+  "engines": {
+    "node": "^16 || ^18 || >= 20"
+  }
 }


mhdawson (approved):
LGTM

KevinEady (approved):

legendecas (approved):

legendecas:
Thank you!

espipj:
Why was this added in a minor release? This is a breaking change and should have been released in a major version
This breaks for example installing https://github.com/parcel-bundler/watcher on Node 10 to 16...

legendecas:
> Why was this added in a minor release? This is a breaking change and should have been released in a major version This breaks for example installing https://github.com/parcel-bundler/watcher on Node 10 to 16...

Released a patch version v7.1.1 to remove the engine requirement. We will still put the constraint on v8.x since we don't test older Node.js versions. Disable [`engine-strict`](https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict) if a person really want to use it with untested versions.

PR #1354, opened 2023-07-21
src: remove duplicate buffer info calls

legendecas:
The constructor of the Uint8Array has called `napi_get_typed_array_info`,
which should be identical to `napi_get_buffer_info`. Remove the duplicated
napi_get_buffer_info call.

Refs: https://github.com/nodejs/node/pull/48742


mhdawson (approved):
LGTM

KevinEady (approved):
LGTM

PR #1374, opened 2023-09-03
src: partition Node-API calls by purity

gabrielschulhof:
We introduce a very thin wrapper around Node-API that allows us to annotate the various APIs with whether they are pure or non-pure by having the wrapper accept either a `Napi::PureEnv` or a `Napi::NonPureEnv`.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -18,6 +18,8 @@
 #include <type_traits>
 #include <utility>
 
+#include "node-api-wrap.h"
+
 namespace Napi {
 
 #ifdef NAPI_CPP_CUSTOM_NAMESPACE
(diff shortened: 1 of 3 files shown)


gabrielschulhof:
The next step would be to define `Napi::PureEnv` as a subclass of `Napi::Env` aka `Napi::NonPureEnv`. This will result in a large number of compilation errors, all of which depend on which Node-APIs we consider to be pure or not. Before embarking on a journey to fix all those errors, thereby effectively splitting node-addon-api into pure and non-pure, I would like to double-check that I've correctly split up the APIs. I did so by checking, for each Node-API, whether it runs `CheckGCAccess()`.

So, @vmoroz, when you get a chance, could you please take a look at the new node-api-wrapper.h to make sure that, for each API listed as pure, it really does need to be pure, and conversely, if I've left any APIs in the non-pure category which I should've moved?

mhdawson:
We discussed in the team meeting today. It will be replaced by a PR in core so closing.

Issue #1379, opened 2023-09-14
Enabling C++ exceptions on Windows is hell


mmomtchev:
(this is the suite of https://github.com/nodejs/node-gyp/issues/2903)

`node-addon-api` seems to include an undocumented but very useful file found `node_modules\node-addon-api\except.gypi`.

Alas, besides being undocumented, this file is very hard to use, since when a module is installed in production, its path will be different and `gyp` does not support variable expansion in `includes` directives.

mhdawson:
After discussion in the node-addon-api team meeting today @legendecas is going to take a look at the issue in a bit more depth.

legendecas:
Tried with node-gyp `-I` option and automatic `config.gypi` detection, not working with no luck. However, I found that we can sort of solve the problem with a dummy gyp dependency. As documented at https://gyp.gsrc.io/docs/InputFormatReference.md#processing-order, a dependency's setting is processed after merging the root `target_defaults`, and variable expansion is allowed in dependency specifiers.

This allows the following pattern to enable c++ exception support:

```
'dependencies': [
  "<!(node -p \"require('node-addon-api').gyp\"):node-addon-api-except"
]
```

PR #1389, opened 2023-10-09
gyp: add common targets

legendecas:
Fixes: https://github.com/nodejs/node-addon-api/issues/1379

As documented at https://gyp.gsrc.io/docs/InputFormatReference.md#processing-order, a dependency's setting is processed after merging the root target_defaults, and variable expansion is allowed in dependency specifiers. This allows addons to include node-addon-api and common configs with a single line of gyp config. For example, when the addon needs c++ exception feature, they can simply declare the dependency on `node-addon-api-except`:

```
{
  "targets": [
    {
      "target_name": "my_node_addon",
      "sources": [
        "src/my_node_addon.cc"
      ],
      'dependencies': [
        "<!(node -p \"require('node-addon-api').targets\"):node_addon_api_except"
      ]
    }
  ]
}
```

No need to copy a paragraph of configs into addons' binding.gyp like documented at https://github.com/nodejs/node-addon-api/blob/9aaf3b1/doc/setup.md#installation-and-usage anymore.

- [x] update the document
- [x] add a test

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,8 @@ const includeDir = path.relative('.', __dirname);
 module.exports = {
   include: `"${__dirname}"`, // deprecated, can be removed as part of 4.0.0
   include_dir: includeDir,
-  gyp: path.join(includeDir, 'node_api.gyp:nothing'),
+  gyp: path.join(includeDir, 'node_api.gyp:nothing'), // deprecated.
+  targets: path.join(includeDir, 'node_addon_api.gyp'),
   isNodeApiBuiltin: true,
   needsFlag: false
 };
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@
 /benchmark/src
 /test/addon_build/addons
 /.vscode
+
+# ignore node-gyp generated files outside its build directory
+/test/*.Makefile
+/test/*.mk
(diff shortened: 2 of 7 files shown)


index.js
```diff
@@ -5,7 +5,7 @@ const includeDir = path.relative('.', __dirname);
 module.exports = {
   include: `"${__dirname}"`, // deprecated, can be removed as part of 4.0.0
   include_dir: includeDir,
-  gyp: path.join(includeDir, 'node_api.gyp:nothing'),
+  gyp: path.join(includeDir, 'node_addon_api.gyp'),
```
gabrielschulhof, line 8: Is it all that harmful to leave this as-is and introduce a new key? If there's even a remote chance that it'll break folks, it may be simplest to not reuse this, but introduce a new key, and document what it does.

gabrielschulhof: Couldn't folks simply write the following?

```gyp
"<!(node -p \"require('node-addon-api').include_dir\")/node_addon_api.gyp:node-addon-api-except"
```

Then we would need only document that node_addon_api.gyp is the way to get add node-addon-api to one's binding.gyp.

legendecas: I've updated the PR to name the entrypoint as `require('node-addon-api').targets`.

mhdawson (approved):
LGTM

PR #1479, opened 2024-04-17
Expose version property in public API

lovell:
This PR contains a proposal to expose the package `version` property in the public API.

The `version` value is already used by the conversion tooling and is additionally useful in external contexts too, for example when specifying/detecting `node-addon-api` as an optional and/or peer dependency.

This would bring `node-addon-api` in line with `node-gyp`, which already exports `require('node-gyp').version`. The change in the PR will allow me to improve the install-time detecting/logging for `node-addon-api` within `sharp`.

It also adds a test with expectations for this property as well as all existing exported properties.

### DIFF
--- a/test/exports.js
+++ b/test/exports.js
@@ -1,0 +1,19 @@
+'use strict';
+
+const { strictEqual } = require('assert');
+const { valid } = require('semver');
+
+const nodeAddonApi = require('../');
+
+module.exports = function test () {
+  strictEqual(nodeAddonApi.include.startsWith('"'), true);
+  strictEqual(nodeAddonApi.include.endsWith('"'), true);
+  strictEqual(nodeAddonApi.include.includes('node-addon-api'), true);
+  strictEqual(nodeAddonApi.include_dir, '');
+  strictEqual(nodeAddonApi.gyp, 'node_api.gyp:nothing');
+  strictEqual(nodeAddonApi.targets, 'node_addon_api.gyp');
+  strictEqual(valid(nodeAddonApi.version), true);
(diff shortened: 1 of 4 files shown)


test/exports.js
```diff
@@ -0,0 +1,18 @@
...
+const nodeAddonApi = require('../');
+
+module.exports = function test () {
+  strictEqual(nodeAddonApi.include.startsWith('"'), true);
+  strictEqual(nodeAddonApi.include.endsWith('"'), true);
+  strictEqual(nodeAddonApi.include.includes('node-addon-api'), true);
+  strictEqual(nodeAddonApi.include_dir, '');
+  strictEqual(nodeAddonApi.gyp, 'node_api.gyp:nothing');
+  strictEqual(nodeAddonApi.targets, 'node_addon_api.gyp');
+  strictEqual(valid(nodeAddonApi.version), true);
```
gabrielschulhof, line 15: It may be worth checking the version explicitly here. Can we compare it to the package.json version?

lovell: Thanks for the review. Would something like the following be suitable?
```suggestion
  strictEqual(valid(nodeAddonApi.version), true);
  strictEqual(nodeAddonApi.version, require('../package.json').version);
```

legendecas: +1 on the change.

lovell: Thanks both, I've added this test expectation.

legendecas (approved):

KevinEady (approved):

mhdawson (approved):
LGTM

mhdawson (approved):
LGTM

PR #1492, opened 2024-05-04
fix compilation for Visual Studio 2022

vmoroz:
Fix PR fixes compilation errors and warnings while compiling code in VS 2022.
- async_worker.cc used deprecated `std::allocator` APIs. Their use is is replaced with APIs that is still valid.
- There was a strange error in `ObjectWrap<T>::WrappedMethod`. It was saying that `&` operation requires an l-value. The issue is fixed by copying `method` pointer to a local variable before invocation.
- VS identified that some fields were not initialized. They are changed to initialize with `{}`.
- VS identified that some functions must be `noexcept`. They are augmented with `NAPI_NOEXCEPT`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,14 @@
 # ignore node-gyp generated files outside its build directory
 /test/*.Makefile
 /test/*.mk
+
+# ignore node-gyp generated Visual Studio files
+*.vcxproj
+*.vcxproj.filters
+*.vcxproj.user
+*.vsidx
+*.sln
+*.suo
+/test/.vs/
+/test/Release/
+/test/Debug/
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2988,10 +2988,9 @@ inline Object Error::Value() const {
   return Object(_env, refValue);
 }
 
-inline Error::Error(Error&& other) NAPI_NOEXCEPT
-    : ObjectReference(std::move(other)) {}
+inline Error::Error(Error&& other) : ObjectReference(std::move(other)) {}
 
-inline Error& Error::operator=(Error&& other) NAPI_NOEXCEPT {
+inline Error& Error::operator=(Error&& other) {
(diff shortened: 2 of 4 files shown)


mhdawson:
@vmoroz thanks for the investigation and the PR.  One question about the `noexcept` is that because they are called by other methods that specify noexcept?

I'd have to look closer but adding noexcept I think means it would be breaking to remove those.  I understand that if we did make them throw exceptions we'd have to change some other methods to catch those,  but the way adding them now would  change that to a breaking versus non-breaking change so we should make sure we believe that for those with the additions we believe it would never make sense for them to through an exception. Otherwise we should possibly add a try catch around the calls versus adding the noexcept.

vmoroz:
> @vmoroz thanks for the investigation and the PR. One question about the `noexcept` is that because they are called by other methods that specify noexcept?
> 
> I'd have to look closer but adding noexcept I think means it would be breaking to remove those. I understand that if we did make them throw exceptions we'd have to change some other methods to catch those, but the way adding them now would change that to a breaking versus non-breaking change so we should make sure we believe that for those with the additions we believe it would never make sense for them to through an exception. Otherwise we should possibly add a try catch around the calls versus adding the noexcept.

@mhdawson , it were the recommendations from Visual Studio. The move constructors and assignment operators are typically must be noexcept. In other cases VS saw that methods never through and suggested to make them noexcept.
These changes were just some "good to have". I can remove them from this PR to keep the bare minimum that fixes only compilation errors.

mhdawson:
@vmoroz I'd say lets keep this PR to just what is needed to fix the compile failures, and then we can look more closely at the good to haves :)

vmoroz:
@mhdawson , I have removed all newly added `NAPI_NOEXCEPT`, but I kept the initialization of previously uninitialized variables. They do not cause compilation errors, but I believe they are quite important to avoid undefined behavior.

mhdawson (approved):
LGTM

PR #1503, opened 2024-05-21
Update ci-win.yml

mhdawson:

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -13,7 +13,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        node-version: [ 18.x, 20.x, 21.x ]
+        node-version: [ 18.x, 20.x, 21.x, 22.x ]
         architecture: [x64, x86]
         os:
           - windows-2019
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -18,6 +18,7 @@ jobs:
         os:
           - windows-2019
           - windows-2022
+      fail-fast: false  # Don't cancel other jobs when one job fails
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -154,7 +154,10 @@ napi_value TemplatedCallback(napi_env env,
                              napi_callback_info info) NAPI_NOEXCEPT {
   return details::WrapCallback([&] {
     CallbackInfo cbInfo(env, info);
-    return Callback(cbInfo);
+    // MSVC requires to copy 'Callback' function pointer to a local variable
+    // before invoking it.
+    auto callback = Callback;
+    return callback(cbInfo);
   });
 }
 


mhdawson:
@vmoroz do you know if  VS2019 should still run with Node.js 22.x?  I'm wondering if we dropped it.

vmoroz:
@mhdawson , I will fix them. Is it OK if I do it directly in this PR?

vmoroz (approved):
LGTM

NickNaso (approved):

legendecas:
> do you know if VS2019 should still run with Node.js 22.x? I'm wondering if we dropped it.

Node.js v22 drops support building core with VS2019: see https://github.com/nodejs/node/blob/v22.x-staging/BUILDING.md#prerequisites.

vmoroz:
> > do you know if VS2019 should still run with Node.js 22.x? I'm wondering if we dropped it.
> 
> Node.js v22 drops support building core with VS2019: see https://github.com/nodejs/node/blob/v22.x-staging/BUILDING.md#prerequisites.

Hmm, but we still run VS2019 in CI: https://ci.nodejs.org/job/node-compile-windows/56159/

Dropping VS2019 for node-addon-api sounds good to me.

I would say a much more important question is to decide on which C++ versions we support.
Node.js 22 uses C++20, and we saw the new build breaks.
Since Node-API allows building modules that work across multiple Node.js versions, we should also allow building C++ code that is not necessary matches Node.js C++ version.

I would say that it is reasonable to say that we support C++17, C++20, and latest C++.
Supporting latest C++ may cause build breaks when GitHub VM images are updated, but we can align our code to the latest C++ standards and libraries sooner.

StefanStojanovic:
I've been working on the Windows part of the CI for the past year or so. With that in mind, I'd like to share with you the state of CI regarding which machines are used for what, but before that, this [document](https://github.com/nodejs/build/blob/main/doc/windows-visualstudio-supported-versions.md) is probably a good reference for this discussion.

**Compilation:** yes we have both VS2019 and VS2022 for compiling Node.js. VS2019 is kept because of the LTS branches, so it is only used for v18 and v20 daily builds and PRs to backport something to those branches. VS2022 is used for compiling Node.js v21+, although v21 supported VS2019 compilation as well. IMHO there is no need to support VS2019 on any new version as even the part for finding a suitable VS2019 installation is removed.

**Native add-ons:** They are a different thing, and with them, we accept a broader range of VS versions, the latest change there is that VS2017 was dropped from Node.js v22. This is also the reason we have VS versions as far as VS2015 in the CI as we are running the native test suites with Node.js binaries compiled with either VS2019 (v18 and v20) or v2022 (v21+) and compiling native add-ons with VS2015-VS2022.

Hopefully, I was able to clear up some confusion and if there is something else you'd like to know about it, feel free to ask.

NickNaso:
From the native add-on developer perspective it's a god thing have the support for `VS2019`.

legendecas (approved):

StefanStojanovic:
> From the native add-on developer perspective it's a good thing have the support for `VS2019`.

Yes, and it is supported for native add-ons. We've just dropped VS2017, so VS2019 should be safe for now.

mhdawson:
Ok, so I think current version with @vmoroz fixes does not drop any support but makes sure we can work across the board. Landing.

PR #1507, opened 2024-05-23
src: improve messages on CheckCast

legendecas:
Print actual value type if the check failed.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -9,6 +9,8 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 // Note: Do not include this file directly! Include "napi.h" instead.
+// This should be a no-op and is intended for better IDE integration.
+#include "napi.h"
 
 #include <algorithm>
 #include <cstring>
@@ -336,6 +338,15 @@ struct AccessorCallbackData {
   Setter setterCallback;
   void* data;
 };
+
+// Debugging-purpose C++-style variant of sprintf().
+template <typename... Args>
+inline std::string SPrintF(const char* format, Args&&... args) {
+  char buf[256];
+  int ret = snprintf(buf, 256, format, args...);
+  NAPI_CHECK(ret >= 0, "SPrintF", "Malformed format");
+  return buf;
+}
 
 }  // namespace details
 
@@ -805,8 +816,7 @@ inline void Boolean::CheckCast(napi_env env, napi_value value) {
   napi_valuetype type;
   napi_status status = napi_typeof(env, value, &type);
   NAPI_CHECK(status == napi_ok, "Boolean::CheckCast", "napi_typeof failed");
-  NAPI_CHECK(
-      type == napi_boolean, "Boolean::CheckCast", "value is not napi_boolean");
+  NAPI_INTERNAL_CHECK_EQ(type, napi_boolean, "%d", "Boolean::CheckCast");
 }
 
 inline Boolean::Boolean() : Napi::Value() {}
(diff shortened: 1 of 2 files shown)


napi.h
```diff
@@ -142,6 +142,26 @@ static_assert(sizeof(char16_t) == sizeof(wchar_t),
     }                                                                          \
   } while (0)
 
+#define NAPI_INTERNAL_STRINGIFY_(x) #x
```
vmoroz, line 145: It seems not being used. Did you intend to use it in the code below in places like `#expr`?

napi.h
```diff
@@ -142,6 +142,26 @@ static_assert(sizeof(char16_t) == sizeof(wchar_t),
...
+#define NAPI_INTERNAL_STRINGIFY(x) NAPI_INTERNAL_STRINGIFY_(x)
+
+// Internal check helper. Be careful that the formatted message length should be
+// max 255 size and null terminated.
+#define NAPI_INTERNAL_CHECK(expr, location, ...)                               \
+  do {                                                                         \
+    if (!(expr)) {                                                             \
+      std::string msg = Napi::details::SPrintF(__VA_ARGS__);                   \
+      Napi::Error::Fatal(location " " #expr, msg.data());                      \
```
vmoroz, line 154: Consider using `c_str()` instead of `data()`.

legendecas: Thanks, fixed

napi.h
legendecas: Yes, that part was removed. Updated.

gabrielschulhof (approved):

napi-inl.h
```diff
@@ -337,6 +339,15 @@ struct AccessorCallbackData {
   void* data;
 };
 
+// Debugging-purpose C++-style variant of sprintf().
```
vmoroz, line 342: I would suggest not to use the variadic arguments here. They do not add any additional type safety in this context, but because this is a template we will have a copy of this function for each combination of parameter types. It would be better to use a simple C-style function instead. (I also propose to make the name less cryptic.)

```C++
inline std::string StringFormat(const char* format, ...)
{
	std::string result;
	va_list args;
	va_start(args, format);
	// stringLength does not include the null terminator.
	int stringLength = vsnprintf(nullptr, 0, format, args);
	// std::string adds 1 to the length to store the null terminator.
	result.resize(stringLength);
	vsnprintf(&result[0], stringLength + 1, format, args);
	va_end(args);
	return result;
}
```
I do not think that we need to check that the vsnprintf succeeded - `result.resize` will cause an exception or crash of the provided size is negative. If the expected size is zero, then we are safe because std::string automatically adds space for the null terminator.

napi.h
```diff
@@ -142,6 +142,23 @@ static_assert(sizeof(char16_t) == sizeof(wchar_t),
...
+      Napi::Error::Fatal(location " " #expr, msg.c_str());                     \
+    }                                                                          \
+  } while (0)
+
+#define NAPI_INTERNAL_CHECK_EQ(actual, expected, value_format, location)       \
+  NAPI_INTERNAL_CHECK((actual) == (expected),                                  \
+                      location,                                                \
+                      "Expected " #actual " to be equal to " #expected         \
+                      ", but got " value_format ".",                           \
+                      actual)
```
vmoroz, line 160: We must change this macro in a way that it does not evaluate `actual` twice. For example,
```C++
#define NAPI_INTERNAL_CHECK_EQ(actual, expected, value_format, location)       \
do {                                                                                                                           \
  auto actualValue = (actual);
  NAPI_INTERNAL_CHECK(actualValue == (expected),                                  \
                      location,                                                \
                      "Expected " #actual " to be equal to " #expected         \
                      ", but got " value_format ".",                           \
                      actualValue)
} while(0)
```
It may be also worth to print the expected value.

vmoroz (approved):

Issue #1367, opened 2023-08-16
idea: detect JS calls at compile time


gabrielschulhof:
Recording this idea so it doesn't slip under the radar:

In support of pure finalizers, I'm wondering if it's possible to templetize `napi::Environment` such that one specialization attaches js-execution-capable finalizers whereas the other attaches pure finalizers. We can use type inference to specialize the right class for a given user-provided finalizer.

That is, if the user's finalizer, after inlining and all, contains no calls to functions that execute JS, then we specialize the environment that attaches pure finalizers, otherwise we specialize the environment that does not. That way, add-on maintainers whose finalizers do not really execute javascript could simply recompile and have their memory leak worries go away without any code changes, once https://github.com/nodejs/node/pull/42651 lands.

The reason I chose `napi::Environment` as the "probe" which will tell us whether the user's finalizer ends up calling a Node-API that executes JS is that all Node-APIs require an environment, so the type inference will snap to a non-pure `napi::Environment` if there is even one call to a JS-executing Node-API in the user's code.

We can actually work on this before https://github.com/nodejs/node/pull/42651 lands, but for now we'll attach finalizers the same way whether or not their body calls into JS. Maybe we can use `#warning` to output compiler messages indicating whether the finalizer would have been pure or not.

Another aspect is this: if there are two `napi::Environment` classes available and you pass the wrong one to the API that sets up the finalizer, it will fail to compile if you pass a "pure" environment along with a finalizer that does not accept such a "pure" environment.

gabrielschulhof:
After discussing at today's Node-API meeting, we concluded that, in core, we should guard the new pure env functionality with an additional `NODE_API_EXPERIMENTAL_PURE_ENV` flag. The motivation and policy are as follow:
- An additional flag the name of which contains the feature allows us to have multiple concurrent flags within `NAPI_EXPERIMENTAL`, allowing adopters to independently adopt each experimental feature.
- The flags will be removed along with `NAPI_EXPERIMENTAL` when we make a new Node-API release.
- The motivation for these flags is that we are breaking APIs we have already released and which are already stable, even though we're breaking them without any corresponding ABI breakage. This would cause the type of breakage beyond what we warn about in conjunction with the `NAPI_EXPERIMENTAL` flag.

gabrielschulhof:
[...]

Issue #1213, opened 2022-09-28
Object wrap has a memory leak bug?


ammarfaizi2:
I have got several memory leak issues with this addon library.
I thought it was me who wrote the code wrong. But when I am trying
to run this example, it has the same memory leak too:

https://github.com/nodejs/node-addon-api/blob/main/doc/object_wrap.md

Run the JavaScript code in a `while (1)` loop will reveal the memory 
leak issue.

What is going wrong with this?

## C++ code
```c++
#include <napi.h>

class Example : public Napi::ObjectWrap<Example> {
  public:
    static Napi::Object Init(Napi::Env env, Napi::Object exports);
    Example(const Napi::CallbackInfo& info);
    static Napi::Value CreateNewItem(const Napi::CallbackInfo& info);

  private:
    double _value;
    Napi::Value GetValue(const Napi::CallbackInfo& info);
    Napi::Value SetValue(const Napi::CallbackInfo& info);
};

Napi::Object Example::Init(Napi::Env env, Napi::Object exports) {
    // This method is used to hook the accessor and method callbacks
    Napi::Function func = DefineClass(env, "Example", {
        InstanceMethod<&Example::GetValue>("GetValue", static_cast<napi_property_attributes>(napi_writable | napi_configurable)),
        InstanceMethod<&Example::SetValue>("SetValue", static_cast<napi_property_attributes>(napi_writable | napi_configurable)),
        StaticMethod<&Example::CreateNewItem>("CreateNewItem", static_cast<napi_property_attributes>(napi_writable | napi_configurable)),
    });

    Napi::FunctionReference* constructor = new Napi::FunctionReference();

    // Create a persistent reference to the class constructor. This will allow
    // a function called on a class prototype and a function
    // called on instance of a class to be distinguished from each other.
    *constructor = Napi::Persistent(func);
    exports.Set("Example", func);

    // Store the constructor as the add-on instance data. This will allow this
    // add-on to support multiple instances of itself running on multiple worker
    // threads, as well as multiple instances of itself running in different
    // contexts on the same thread.
    //
    // By default, the value set on the environment here will be destroyed when
    // the add-on is unloaded using the `delete` operator, but it is also
    // possible to supply a custom deleter.
    env.SetInstanceData<Napi::FunctionReference>(constructor);

    return exports;
}

Example::Example(const Napi::CallbackInfo& info) :
    Napi::ObjectWrap<Example>(info) {
  Napi::Env env = info.Env();
  // ...
[...]

PR #1514, opened 2024-06-06
src: add support for nogc types via `NogcEnv` 

KevinEady:
- Introduce `Napi::NogcEnv` class, taking the place of `Napi::Env` in finalizer callbacks.
- Introduce `void Napi::NogcEnv::AddPostFinalizer()` and overloads.
- Modify tests to support above changes

TODO:
- [ ] docs
- [ ] discuss implementation, namings, etc


KevinEady:
@vmoroz your comment on removing const ref on `std::nullptr_t` has been addressed, and @mhdawson your comment on reference deletion has been addressed, both in [2a689b6](https://github.com/nodejs/node-addon-api/pull/1514/commits/2a689b6bde71b731cb29a2a0b5aec244a2fc59ca)

.github/workflows/ci.yml
```diff
@@ -12,8 +12,12 @@ jobs:
   test:
     timeout-minutes: 30
     strategy:
+      fail-fast: false
       matrix:
-        node-version: [ 18.x, 20.x, 21.x, 22.x ]
+        api_version:
```
vmoroz, line 17: Should we add checks for the standard/experimental to ci-win.yml too?

KevinEady: Good thought... I pretty much copied Gabriel's CI file. I'll comment on his PR to verify, since mine would end up being overwritten when i rebase to master after his is merged.

KevinEady:
Hi team,

Using some SFINAE magic, I was able to determine at compile time how to associate the finalizer -- either directly or with `node_api_post_finalizer`. This would address #1367 as implemented.

This means no changes were needed to any existing tests, and I created a new test in `finalizer_order` to ensure finalizers are being called correctly.

PTAL! @gabrielschulhof @vmoroz @legendecas @mhdawson

test/finalizer_order.js
```diff
@@ -0,0 +1,98 @@
...
+    () => {
+      let test = new binding.finalizer_order.Test(() => { isCallbackCalled = true; });
+      test = null;
+
+      global.gc();
+
+      if (isExperimental) {
+        assert.strictEqual(binding.finalizer_order.Test.isNogcFinalizerCalled, true, 'Expected nogc finalizer to be called [before ticking]');
+        assert.strictEqual(binding.finalizer_order.Test.isGcFinalizerCalled, false, 'Expected gc finalizer to not be called [before ticking]');
+        assert.strictEqual(isCallbackCalled, false, 'Expected callback not be called [before ticking]');
```
gabrielschulhof, line 27:
```suggestion
        assert.strictEqual(isCallbackCalled, false, 'Expected callback to not be called [before ticking]');
```

gabrielschulhof (approved):
I think the Gc vs. NoGc nomenclature is a bit confusing. Unfortunately, we established this in core. I think maybe we should go with `DuringGc` and `OutsideGc`. For env, we'd have `DuringGcEnv` and `Env`, since `Env` is already established, but for the finalizer type names I think we can go with `DuringGc` and `OutsideGc` for clarity.

napi.h
```diff
@@ -2415,6 +2448,7 @@ class ObjectWrap : public InstanceWrap<T>, public Reference<Object> {
       napi_property_attributes attributes = napi_default);
   static Napi::Value OnCalledAsFunction(const Napi::CallbackInfo& callbackInfo);
   virtual void Finalize(Napi::Env env);
+  virtual void Finalize(NogcEnv env);
```
gabrielschulhof, line 2451:
```suggestion
  virtual void Finalize(Napi::NogcEnv env);
```

gabrielschulhof:
I tested this with `Napi::External` and it works great!

```JS
const { makeSync, makeAsync } = require('bindings')('test_sync_fini')

const doAsync = process.argv[2] === 'async'

;(function test(iter = 0) {
  for (let i = 0; i < 1000000; i++) {
    const id = `${iter}:${i}`
    if (doAsync) {
      makeAsync(id)
    } else {
      makeSync(id)
    }
  }
  setTimeout(test, 0, iter + 1)
})()
```

Here's a video of the output. Guess which column is async 🙂 

https://github.com/nodejs/node-addon-api/assets/976081/1eb30993-926c-4233-8495-93953f112ace

gabrielschulhof:
We should rename `AddPostFinalizer` to just `PostFinalizer`.

KevinEady:
@gabrielschulhof ,

>  For env, we'd have DuringGcEnv and Env, since Env is already established, ...

One could argue that being a Node-API wrapper, the naming in this library should be as consistent with Node-API as possible. The (almost) 1-to-1 mapping helps link up with other documentation (eg. referencing the [Node-API](https://nodejs.org/api/n-api.html) docs), or translating between Node-API C code and node-addon-api C++ code.

I do agree though, the name `DuringGcEnv` better describes the restriction ("during GC") on how the environment can be used, but not sure if we should stray from the Node-API names.

---

 > ... but for the finalizer type names I think we can go with `DuringGc` and `OutsideGc` for clarity.

By "finalizer type names" do you mean changing the `Finalizer` in places eg.:
```cpp
template <typename Finalizer>
static External New(napi_env env, T* data, Finalizer finalizeCallback);
```

If so, I don't think we should change this name since the finalizer passed can accept either a `DuringGcEnv` or `Env`. A `Finalizer` is a callback ran either during the GC cycle (if using `DuringGcEnv`) or on a future tick after the GC cycle completes (if using `Env`). This means the concept of "finalization" is dependent on your callback signature, and would be documented as such.

---

> We should rename AddPostFinalizer to just PostFinalizer.

Makes sense, esp. considering what I said previously.

---

While on the subject of documentation: we don't have any specific entry for "finalization", as a finalizer callback is always documented on each individual object or function, eg:

- `Napi::External`:
https://github.com/nodejs/node-addon-api/blob/bf49519a4ce08ee5320327c9a0199cd89d5b87b3/doc/external.md?plain=1#L7
- `Napi::Buffer::New`:
https://github.com/nodejs/node-addon-api/blob/bf49519a4ce08ee5320327c9a0199cd89d5b87b3/doc/buffer.md?plain=1#L73-L75
- and others

I propose we create a new top-level documentation page for finalization describing the above, and can have each use of finalizers in the documentation link to this page.

napi.h
```diff
@@ -312,30 +313,31 @@ using MaybeOrValue = T;
 ///
 /// In the V8 JavaScript engine, a Node-API environment approximately
 /// corresponds to an Isolate.
-class Env {
- private:
-  napi_env _env;
+class NogcEnv {
+ protected:
```
KevinEady, line 317: Should we make this `private` and have `Env` be a friend class? With it being `protected`, the default finalizers (as well as `_env`) would be exposed in any user-defined classes that extend `NogcEnv`.

KevinEady:
I've added some WIP documentation on the top-level finalization page: https://github.com/nodejs/node-addon-api/blob/6591a3609c2b94f22956b72736e06b86586556cf/doc/finalization.md

KevinEady:
@mhdawson @gabrielschulhof as discussed in the 5 July Node-API meeting, I updated the docs to use the concept of a "basic finalizer" as a special categorization of finalizers, and removing the concept of "[a]synchronous". PTAL: https://github.com/nodejs/node-addon-api/blob/7d7b8211552216657494144d0417bf57b690871e/doc/finalization.md 

Once we finalize (hah) this verbiage, I will update any associated type names in the headers + tests, and update the other docs that have any finalization features.

gabrielschulhof:
@KevinEady the doc LGTM 👍

legendecas:
(this is still a draft PR)

mhdawson:
doc LGTM with a few suggestions:

1) I would remove 

NOTE: Optimizations via basic finalizers will only occur if using NAPI_EXPERIMENTAL and the NODE_API_EXPERIMENTAL_NOGC_ENV_OPT_OUT define flag has not been set. Otherwise, the engine will not differentiate between basic and (extended) finalizers.

there is never any promise that optimization will take place so I think we can just remove that.

2) In terms of 

In addition to passing finalizers to Napi::Externals and other Node-API constructs, use Napi::BasicEnv::PostFinalize(Napi::BasicEnv, Finalizer) to schedule a callback to run outside of the garbage collector finalization.

can we explain without promising it will run outside of gc collection?  Maybe something like "to maximize the likelyhood that the finalization not being tied to garbage collection finalization" ...

KevinEady:
@legendecas the overall C++ work/implementation is finished, but the PR still needs some more work (markdown docs, test docs, ...) 

@mhdawson 

> 1. I would remove
> 
> NOTE: Optimizations via basic finalizers will only occur if using NAPI_EXPERIMENTAL and the NODE_API_EXPERIMENTAL_NOGC_ENV_OPT_OUT define flag has not been set. Otherwise, the engine will not differentiate between basic and (extended) finalizers.
> 
> there is never any promise that optimization will take place so I think we can just remove that.

With this, I was trying to convey that using basic finalizers only changes functionality in experimental mode. I could see someone using basic finalizers without experimental mode, running into https://github.com/nodejs/node-addon-api/issues/1213 and being a bit confused. Maybe we can use some other verbiage?

> **NOTE**: Node-API will only differentiate between basic and (extended) finalizers when using `NAPI_EXPERIMENTAL` and the `NODE_API_EXPERIMENTAL_NOGC_ENV_OPT_OUT` define flag has _not_ been set. Otherwise, basic finalizers execute in the same manner as (extended) finalizers.

What do we think of that?

----


> 2. In terms of
> 
> In addition to passing finalizers to Napi::Externals and other Node-API constructs, use Napi::BasicEnv::PostFinalize(Napi::BasicEnv, Finalizer) to schedule a callback to run outside of the garbage collector finalization.
> 
> can we explain without promising it will run outside of gc collection?

Well, I took inspiration from the [Node-API docs for `node_api_post_finalizer`](https://nodejs.org/api/n-api.html#node_api_post_finalizer):

> node_api_post_finalizer helps to work around this limitation by allowing the add-on to defer calls to such Node-APIs to a point in time outside of the GC finalization.

Also, when you say:

> "to maximize the likelyhood that the finalization not being tied to garbage collection finalization"

The API is intended to _guarantee_ the callback does not run inside GC, so saying "maximize the likelihood" is not completely accurate, as it 100% will not run inside GC finalization. 

Since this API strictly deals with callbacks + GC finalization order, I think it's a good thing to keep.

KevinEady:
In Node API meeting 19 July, we discussed:

> 1. I would remove
> 
> NOTE: Optimizations via basic finalizers will only occur if using NAPI_EXPERIMENTAL and the NODE_API_EXPERIMENTAL_NOGC_ENV_OPT_OUT define flag has not been set. Otherwise, the engine will not differentiate between basic and (extended) finalizers.
> 
> there is never any promise that optimization will take place so I think we can just remove that.

Yes, remove this.

----

> 2. In terms of
> 
> In addition to passing finalizers to Napi::Externals and other Node-API constructs, use Napi::BasicEnv::PostFinalize(Napi::BasicEnv, Finalizer) to schedule a callback to run outside of the garbage collector finalization.
> 
> can we explain without promising it will run outside of gc collection?

The name `node_api_post_finalizer` can be made in the context of "running (non-basic) code after (**_post_**) a **_finalizer_**" , so maybe we can change the documentation in core (and also here) if needed, but the current wording is fine.

----

Additionally, we discussed adding an _opt-in_ flag that requires -- at compilation time -- all finalizers are of the basic type, under the realization that:
- if using the C-based Node-API, opting into basic finalizers requires _all_ finalizers to be basic, providing devs a way to guarantee all of their finalizers would be run in the "maybe optimized" way and fail compilation if callbacks do not match the current basic finalizer signature.
- if using this wrapper, (currently) there is no guarantee that all finalizers will be of the basic signature, meaning developers do not have an easy way to forcefully get compilation errors for finalizers that are not basic.

I suggest using a define: `NODE_API_REQUIRE_BASIC_FINALIZERS`

KevinEady:
Hi @mhdawson @vmoroz @legendecas @gabrielschulhof ,

I believe this PR is ready for review now! 

- Finished renaming "synchronous" finalizer to "basic" finalizer
- Updated docs
- Removed semver major label since it's a feature add and nothing breaking
- Changed the PR title to `feat:` to get release-please stuff correct
 
PTAL! Hopefully we can discuss in this week's Node API team meeting.

---

> Additionally, we discussed adding an _opt-in_ flag that requires -- at compilation time -- all finalizers are of the basic type, ...

I will defer this to a different PR, as this will need to update test infrastructure (to test for compilation failures)

KevinEady:
I would like to land https://github.com/nodejs/node-addon-api/pull/1548 and rebase this PR so we can get a Windows + experimental test run.

KevinEady:
The Windows experimental CI passed, eg: [test (experimental, 20.x, x64, windows-2019)](https://github.com/nodejs/node-addon-api/actions/runs/10218072154/job/28273303597#step:7:3698)
```
Testing with Node-API Version '2147483647'.
```

@legendecas @mhdawson @gabrielschulhof @vmoroz PTAL! 👍

mhdawson (approved):
LGTM

napi.h
```diff
@@ -187,6 +187,7 @@ namespace NAPI_CPP_CUSTOM_NAMESPACE {
 #endif
 
 // Forward declarations
+class BasicEnv;
```
legendecas, line 190: Is this necessary?

legendecas (approved):
Good work! Thanks

doc/finalization.md
```diff
@@ -0,0 +1,149 @@
...
+});
+```
+
+## Basic Finalizers
+
+Use of basic finalizers may allow the engine to perform optimizations when
+scheduling or executing the callback. For example, V8 does not allow access to
+the engine heap during garbage collection. Restricting finalizers from accessing
+the engine heap allows the callback to execute during garbage collection,
+providing a chance to free native memory on the current tick.
```
legendecas, line 31: I think it would be good to just say "eagerly" being GC-ed.

```suggestion
providing a chance to free native memory eagerly.
```

KevinEady:
Hi @gabrielschulhof , IIRC you mentioned you had an in-progress review. Any movement on that?

doc/env.md
```diff
@@ -1,8 +1,14 @@
 # Env
```
gabrielschulhof, line 2:
```suggestion
Class `Napi::Env` inherits from class [`Napi::BasicEnv`][].

```

and then add

```
[`Napi::BasicEnv`]: ./basic_env.md
```

to the bottom of the file.

doc/finalization.md
```diff
@@ -0,0 +1,149 @@
+# Finalization
+
+Various node-addon-api methods accept a templated `Finalizer finalizeCallback`
+parameter. This parameter represents a native callback function that runs in
+response to a garbage collection event. A finalizer is considered a _basic_
+finalizer if the callback only utilizes a certain subset of APIs, which may
+provide optimizations, improved execution, or other benefits.
```
gabrielschulhof, line 7:
```suggestion
provide more efficient memory management, optimizations, improved execution, or other benefits.
```

I think we should explicitly point out memory management because that's what caused us to go down this path in the first place.

doc/finalization.md
```diff
@@ -0,0 +1,149 @@
...
+Napi::ArrayBuffer::New(
+    Env(), data, length, [](Napi::BasicEnv /*env*/, void* finalizeData) {
+      delete[] static_cast<uint8_t*>(finalizeData);
+    });
+```
+
+## Scheduling Finalizers
+
+In addition to passing finalizers to `Napi::External`s and other Node-API
+constructs, use `Napi::BasicEnv::PostFinalize(Napi::BasicEnv, Finalizer)` to
```
gabrielschulhof, line 49:
```suggestion
constructs, `Napi::BasicEnv::PostFinalize(Napi::BasicEnv, Finalizer)` can be used to
```

We should use passive voice.

doc/finalization.md
KevinEady: Addressed in [195ec28](https://github.com/nodejs/node-addon-api/pull/1514/commits/195ec28a7b1dcfe571c6df5f8a2e1bb991126ff1).

doc/basic_env.md
```diff
@@ -0,0 +1,201 @@
+# BasicEnv
+
+The opaque data structure containing the environment in which the request is
```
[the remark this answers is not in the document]
KevinEady: Addressed in [195ec28](https://github.com/nodejs/node-addon-api/pull/1514/commits/195ec28a7b1dcfe571c6df5f8a2e1bb991126ff1).

doc/basic_env.md
```diff
@@ -0,0 +1,201 @@
...
+instance of the addon is unloaded. This overload accepts an additional hint to
+be passed to `fini`.
+
+### GetModuleFileName
+
+```cpp
+const char* Napi::Env::GetModuleFileName() const;
+```
+
+Returns a A URL containing the absolute path of the location from which the
```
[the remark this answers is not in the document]
KevinEady: Addressed in [195ec28](https://github.com/nodejs/node-addon-api/pull/1514/commits/195ec28a7b1dcfe571c6df5f8a2e1bb991126ff1).

doc/basic_env.md
```diff
@@ -0,0 +1,201 @@
...
+HintType* hint)`. If not given, the default finalizer will be used, which simply
+uses the `delete` operator to destroy `T*` when the addon instance is unloaded.
+- `[in] data`: A pointer to data that will be associated with the instance of
+the addon for the duration of its lifecycle.
+- `[in] hint`: A pointer to data that will be associated with the instance of
+the addon for the duration of its lifecycle and will be passed as a hint to
+`fini` when the addon instance is unloaded.
+
+Associates a data item stored at `T* data` with the current instance of the
+addon. The item will be passed to the function `fini` which gets called when an
```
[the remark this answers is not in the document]
KevinEady: Addressed in [195ec28](https://github.com/nodejs/node-addon-api/pull/1514/commits/195ec28a7b1dcfe571c6df5f8a2e1bb991126ff1).

doc/basic_env.md
```diff
@@ -0,0 +1,201 @@
...
+instance of the addon is unloaded. This overload accepts an additional hint to
```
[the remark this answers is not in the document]
KevinEady: Addressed in [195ec28](https://github.com/nodejs/node-addon-api/pull/1514/commits/195ec28a7b1dcfe571c6df5f8a2e1bb991126ff1).

doc/env.md
KevinEady: Addressed in [195ec28](https://github.com/nodejs/node-addon-api/pull/1514/commits/195ec28a7b1dcfe571c6df5f8a2e1bb991126ff1).

doc/finalization.md
KevinEady: Addressed in [195ec28](https://github.com/nodejs/node-addon-api/pull/1514/commits/195ec28a7b1dcfe571c6df5f8a2e1bb991126ff1).

doc/finalization.md
KevinEady: Addressed in [195ec28](https://github.com/nodejs/node-addon-api/pull/1514/commits/195ec28a7b1dcfe571c6df5f8a2e1bb991126ff1).

mhdawson:
Looks like all comments have been addressed and CI is green. Landing.

KevinEady:
Hi @legendecas ,

This PR was merged into main, but the `release-please` action failed: https://github.com/nodejs/node-addon-api/actions/runs/10687066491/job/29623803353 

I think it has to do with permissions?

PR #1538, opened 2024-07-12
chore: do not include component name in tag

legendecas:
By default release-please creates a tag with component names: https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#subsequent-versions

There is only one component in the repo so there is no need to include component name in tag name.

Refs: https://github.com/nodejs/node-addon-api/releases/tag/node-addon-api-v8.1.0


mhdawson (approved):
LGTM

vmoroz (approved):

PR #1570, opened 2024-09-03
chore: fix dependabot config

legendecas:
https://github.com/nodejs/node-addon-api/network/updates reports that the config file contains errors:

```
The property '#/updates/0/groups/all/patterns' of type string did not match the following type: array
The property '#/updates/1/groups/all/patterns' of type string did not match the following type: array
The property '#/updates/2/groups/all/patterns' of type string did not match the following type: array
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,8 @@ updates:
       interval: weekly
     groups:
       all:
-        patterns: '*'
+        patterns:
+        - '*'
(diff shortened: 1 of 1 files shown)


legendecas:
I guess this is the reason why we still have so many dependabot PRs https://github.com/nodejs/node-addon-api/pulls?q=is%3Apr+label%3Adependencies+

Issue #1573, opened 2024-09-04
Failures in CI across platforms after nogc env update


mhdawson:
There have been some failures since I landed the PR

https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=rhel8-x64/9278/console

https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=debian10-x64/9278/console

Looks like a warning about an unused parameter

```
../../napi-inl.h:5011:63: error: parameter ‘env’ set but not used [-Werror=unused-but-set-parameter]
 inline void ObjectWrap<T>::FinalizeCallback(node_api_nogc_env env,
```

mhdawson:
@KevinEady is that something you have some cycles to look at ?

mhdawson:
It looks like `env` is unused with the default defines, should just need to adjust the parameter definition, looking at it.

KevinEady:
> It looks like `env` is unused with the default defines, should just need to adjust the parameter definition, looking at it.

It's not as simple because of the `constexpr` checks: we only need to use `env` if the base class defines certain methods. 

The test fails with:
```
In file included from ../../napi.h:3257,
                 from ../addon_data.cc:3:
../../napi-inl.h: In instantiation of ‘static void Napi::ObjectWrap<T>::FinalizeCallback(node_api_nogc_env, void*, void*) [with T = Addon::VerboseIndicator; node_api_nogc_env = napi_env__*]’:
```
... and this is because [`Addon::VerboseIndicator`](https://github.com/nodejs/node-addon-api/blob/600fe6c114a127ea11fc7909e08bf623dfadea95/test/addon_data.cc#L18) does not have a `Finalize`. Since there's no `VerboseIndicator::Finalize`, the `ObjectWrap::FinalizeCallback` doesn't use the `env` parameter.

Looking at the previous implementation: the `ObjectWrap` base class defines an empty `Finalize()` to call with `env`:
https://github.com/nodejs/node-addon-api/blob/bf49519a4ce08ee5320327c9a0199cd89d5b87b3/napi-inl.h#L4924-L4932
https://github.com/nodejs/node-addon-api/blob/bf49519a4ce08ee5320327c9a0199cd89d5b87b3/napi-inl.h#L4837-L4838

Should I do something similar...? Seems odd to need to do this just to get rid of the warning... but I could imagine some projects using the same warnings-as-error configuration for this `unused-but-set-parameter`.

KevinEady:
[...]

PR #1574, opened 2024-09-04
fix: call base basic finalizer if none defined

KevinEady:
Call no-op `ObjectWrap<T>::Finalize` if user does not provide a `Finalizer` method.

Fixes: #1573

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -5023,6 +5023,11 @@ inline void ObjectWrap<T>::FinalizeCallback(node_api_nogc_env env,
 #endif
 
     instance->Finalize(Napi::BasicEnv(env));
+  } else {
+    // The child class may override Finalize but _not_ with a BasicEnv. Calling
+    // via `instance->Finalize()` would then give a compile error. Explicitly
+    // use the ObjectWrap<T>'s Finalize, which has BasicEnv override.
+    instance->ObjectWrap<T>::Finalize(Napi::BasicEnv(env));
   }
 
   // If class overrides the (extended) finalizer, either schedule it or
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -5011,6 +5011,10 @@ template <typename T>
 inline void ObjectWrap<T>::FinalizeCallback(node_api_nogc_env env,
                                             void* data,
                                             void* /*hint*/) {
+  // If the child class does not override _any_ Finalize() method, `env` will be
+  // unused because of the constexpr guards. Explicitly reference it here to
+  // bypass compiler warnings.
+  (void)env;
   T* instance = static_cast<T*>(data);
 
   // Prevent ~ObjectWrap from calling napi_remove_wrap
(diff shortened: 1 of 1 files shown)


napi-inl.h
```diff
@@ -5023,6 +5023,11 @@ inline void ObjectWrap<T>::FinalizeCallback(node_api_nogc_env env,
 #endif
 
     instance->Finalize(Napi::BasicEnv(env));
+  } else {
```
mhdawson, line 5026: I see how this fixes the existing, not used error, but does it also fix running some finalization that should have run before but would not have, or now call the method just to fix the not used error?

KevinEady: > but does it also fix running some finalization that should have run before but would not have

This is **not** the case. If the user defined their own `MyClass::Finalize(BasicEnv)`, the `constexpr (details::HasBasicFinalizer<T>::value)` would be true, and then call `instance->Finalize(Napi::BasicEnv(env))`.

> ... now call the method just to fix the not used error

This **is** the case, so it is adding an inefficiency (perhaps the compiler optimizes it out)

mhdawson: My questions is to make sure the additional invocation is desired, if so good, sounds like we found a potential bug as well. If not I think we can just add a `(void) env;` early in the function to mark it as used to address the warning.

KevinEady:
I've started a single Jenkins v22 run to see if the fix works

| Version | Job | Status |
|---:|---|:---:|
| v22.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/9288/ | ... |

napi-inl.h
NickNaso: I'm trying the same solution you proposed adding `(void) env;` on my local machine

NickNaso: it worked well on my local machine.

NickNaso:
| Version | Job | Status |
|---:|---|:---:|
| v22.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/9289/ | ... |

NickNaso:
The CI is green now I try to launch it for all LTS and `main`

CI:
| Version | Job | Status |
|---:|---|:---:|
| main | https://ci.nodejs.org/job/node-test-node-addon-api-new/9291/ | ... |
| v22.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/9282/ | ... |
| v20.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/9293/ | ... |
| v18.x | https://ci.nodejs.org/job/node-test-node-addon-api-new/9294/ | ... |

NickNaso (approved):

mhdawson (approved):
LGTM

mhdawson:
Looks like CIs are good. @NickNaso thanks for kicking those off and checking the solution. Going to land.

mhdawson:
Will check the regular node-addon ci tomorrow to confirm all locks good. @KevinEady, @NickNaso thanks for the quick work to get this addressed.

Issue #1584, opened 2024-09-30
Upgrading from v8.1.0 to v8.2.0 breaks compilation


ddolcimascolo:
Hi all,

In a Renovate MR upgrading node-addon-api to v8.2.0 (from v8.1.0) my build fails with lots of compilation errors in `napi.h`  :

```
In file included from ../src/supersecret.cc:3:                                                                                                                                                          
/data/dev/supersecret/node_modules/node-addon-api/napi.h:317:3: error: ‘node_api_nogc_env’ does not name a type                                                                                         
  317 |   node_api_nogc_env _env;                                                                                                                                                                       
      |   ^~~~~~~~~~~~~~~~~                                                                                                                                                                             
/data/dev/supersecret/node_modules/node-addon-api/napi.h:325:29: error: expected ‘)’ before ‘env’                                                                                                       
  325 |   BasicEnv(node_api_nogc_env env);                                                                                                                                                              
      |           ~                 ^~~~                                                                                                                                                                
      |                             )
/data/dev/supersecret/node_modules/node-addon-api/napi.h:326:12: error: expected type-specifier before ‘node_api_nogc_env’
  326 |   operator node_api_nogc_env() const;
      |            ^~~~~~~~~~~~~~~~~
/data/dev/supersecret/node_modules/node-addon-api/napi.h: In member function ‘bool Napi::BasicEnv::operator==(const Napi::BasicEnv&) const’:
/data/dev/supersecret/node_modules/node-addon-api/napi.h:335:12: error: ‘_env’ was not declared in this scope
  335 |     return _env == other._env;
      |            ^~~~
/data/dev/supersecret/node_modules/node-addon-api/napi.h:335:26: error: ‘const class Napi::BasicEnv’ has no member named ‘_env’
  335 |     return _env == other._env;
      |                          ^~~~
/data/dev/supersecret/node_modules/node-addon-api/napi.h: In member function ‘bool Napi::BasicEnv::operator==(std::nullptr_t) const’:
/data/dev/supersecret/node_modules/node-addon-api/napi.h:338:12: error: ‘_env’ was not declared in this scope
  338 |     return _env == nullptr;
      |            ^~~~
/data/dev/supersecret/node_modules/node-addon-api/napi.h: At global scope:
/data/dev/supersecret/node_modules/node-addon-api/napi.h:2467:32: error: ‘node_api_nogc_env’ has not been declared
[...]

PR #1585, opened 2024-10-04
fix: missing node_api_nogc_env definition

legendecas:
Fixes: https://github.com/nodejs/node-addon-api/issues/1584

Add GitHub Actions for `node-api-headers` compatibility test, covering old Node.js versions. `node-api-headers` strips experimental apis so only test with stable NAPI_VERSION.

As there was no feature detection defs for `node_api_basic_finalize` or `node_api_basic_env`, use `NODE_API_EXPERIMENTAL_HAS_POST_FINALIZER` for feature detection since it was introduced about the time where `node_api_basic_env`/`node_api_nogc_env` was introduced.

### DIFF
--- a/.github/workflows/node-api-headers.yml
+++ b/.github/workflows/node-api-headers.yml
@@ -1,0 +1,71 @@
+name: Node.js CI with node-api-headers
+
+on: [push, pull_request]
+
+env:
+  PYTHON_VERSION: '3.11'
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        api_version:
+          - '9'
+        node-version:
+          - 22.x
+        node-api-headers-version:
+          - '1.1.0'
+          - '1.2.0'
+          - '1.3.0'
+        os:
+          - ubuntu-latest
+        compiler:
+          - gcc
+          - clang
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Harden Runner
+      uses: step-security/harden-runner@91182cccc01eb5e619899d80e4e971d6181294a7 # v2.10.1
+      with:
+        egress-policy: audit
+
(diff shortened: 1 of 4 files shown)


.github/workflows/node-api-headers.yml
```diff
@@ -0,0 +1,70 @@
...
+        node-version:
+          - 22.x
+        node-api-headers-version:
+          - '1.1.0'
+          - '1.2.0'
+          - '1.3.0'
+        os:
+          - ubuntu-latest
+        compiler:
+          - gcc
```
NickNaso, line 28: `clang` is missing.

legendecas: Thnaks, added!

napi.h
```diff
@@ -314,16 +322,20 @@ using MaybeOrValue = T;
...
 #if NAPI_VERSION > 5
   template <typename T>
   static void DefaultFini(Env, T* data);
   template <typename DataType, typename HintType>
   static void DefaultFiniWithHint(Env, DataType* data, HintType* hint);
 #endif  // NAPI_VERSION > 5
  public:
-  BasicEnv(node_api_nogc_env env);
+  BasicEnv(node_addon_api_basic_env env);
+#ifdef NODE_API_EXPERIMENTAL_HAS_POST_FINALIZER
```
KevinEady, line 334: In the node-addon-api meeting on 4 Oct, we discussed that it should be possible just to have a operator to `node_addon_api_basic_env` and `node_addon_api_basic_finalize`, as the compiler will allow conversion from `BasicEnv` -> `node_addon_api_basic_env` -> `node_api_nogc_env` or `napi_env` depending on options allowed.

legendecas: Updated, ptal again, thanks

NickNaso (approved):

mhdawson (approved):
LGTM

KevinEady (approved):
LGTM

PR #1586, opened 2024-10-04
doc: amend outdated doc about multiple environments

legendecas:
Since the first stable version of node-api, multiple environments are supported.


mhdawson (approved):
LGTM

Issue #1555, opened 2024-08-12
std::exception catching once more


kamil321:
Hi there. I might be beating a dead horse here (#601, #1225), but I have to try. I've got some bindings with thousands of functions and a lot of them like to throw exceptions. I was pretty shocked to find out that node just shuts down without any message when there's anything wrong.
And that's of course because of the callback wrappers catching Napi::Error and nothing more.

I - kind of - understand that touching unknown exceptions sounds bad, but shutting down everything with neither a message nor a possibility to catch on the node end seems much, much worse. Leaving the choice to the developers is probably a good call, but not if they have to write the same try-catch 82394723 times. 

I ended up modifying the header just like in this comment (e.what from std:exc, simple message from unknowns): https://github.com/nodejs/node-addon-api/issues/601#issuecomment-554614021 and some inner peace returned. At least until the next version upgrade.

Is this the perfect solution and should find its way into the api? No idea. But it could just lay there, hidden behind a global. Or maybe Napi could provide a way to inject a generic exception handler?

Thanks in advance for reconsidering.

legendecas:
In today's Node-API meeting, we discussed about loosing the type constraint on the default catch clause and avoid leaking the c++ exceptions to Node.js core because exceptions are only supported in node-addon-api.

We may add a handler/configuration/macro to allow setting custom arbitrary exception handler so that if poeple still like the addon to crash if exception is unhandled can be still achievable: https://github.com/nodejs/node-addon-api/blob/main/napi-inl.h#L94-L128.

We also agreed that it will not be considered as a breaking change if the default to changed to catch all exceptions and throw as JS exceptions.

legendecas:
@kamil321 would you like to working on it?

kamil321:
@legendecas thank you, that's a great news. But about working on it - heart says 'sure', brain screams 'nooo'. See, I'm mostly a TS dev and this is my first dip into cpp after like 7 years or so. 

Even in my local patch I've encountered a pretty serious wall: to construct a Napi:Error in a potential handler, we apparently need some napi_env. So I added it as a param to both WrapCallback and WrapVoidCallback and tried to pass it in every place. It worked in most cases, but sometimes I got a random nullptr instead of env, or some exception related to unwrapping the callback. If I spent another month on it, then maybe I could fix it, but I guess someone familiar with Napi internals can do it in 10 minutes.

[...]

PR #1593, opened 2024-10-18
feat: allow catching all exceptions

KevinEady:
Add new compile-time define (`NODE_ADDON_API_CPP_EXCEPTIONS_ALL`)  to enable catching all exceptions. This must be used _in conjunction with_ `NODE_ADDON_API_CPP_EXCEPTIONS`.

Rename the `NAPI_` prefixed exception preprocessor definitions (`NAPI_CPP_EXCEPTIONS` and `NAPI_DISABLE_CPP_EXCEPTIONS`) to be prefixed with our standard `NODE_ADDON_API_` prefix in a backwards-compatible way.

Fixes: #1555

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -150,8 +150,9 @@ inline void WrapVoidCallback(napi_env, Callable callback) {
   }
 #endif  // NODE_ADDON_API_CPP_EXCEPTIONS_ALL
 #else
-  // When C++ exceptions are disabled, errors are immediately thrown as JS
-  // exceptions, so there is no need to catch and rethrow them here.
+  // When C++ exceptions are disabled, there is no need to catch and rethrow C++
+  // exceptions. JS errors should be thrown with
+  // `Error::ThrowAsJavaScriptException`.
   callback();
 #endif  // NODE_ADDON_API_CPP_EXCEPTIONS
 }
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -79,19 +79,33 @@ inline napi_status AttachData(napi_env env,
 // For use in JS to C++ callback wrappers to catch any Napi::Error exceptions
 // and rethrow them as JavaScript exceptions before returning from the callback.
 template <typename Callable>
-inline napi_value WrapCallback(Callable callback) {
-#ifdef NAPI_CPP_EXCEPTIONS
+#ifdef NODE_ADDON_API_CPP_EXCEPTIONS_ALL
+inline napi_value WrapCallback(napi_env env, Callable callback) {
+#else
+inline napi_value WrapCallback(napi_env, Callable callback) {
+#endif
+#ifdef NODE_ADDON_API_CPP_EXCEPTIONS
   try {
     return callback();
   } catch (const Error& e) {
     e.ThrowAsJavaScriptException();
     return nullptr;
   }
-#else   // NAPI_CPP_EXCEPTIONS
+#ifdef NODE_ADDON_API_CPP_EXCEPTIONS_ALL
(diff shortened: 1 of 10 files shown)


KevinEady:
I'd like to discuss the naming of the new define `NAPI_CPP_EXCEPTIONS_ALL` ... to go with convention it should be `NODE_ADDON_API_CPP_EXCEPTIONS_ALL`.

We can use `NODE_ADDON_API_CPP_EXCEPTIONS_ALL` and introduce `NODE_ADDON_API_CPP_EXCEPTIONS` as a (backwards-compatible) replacement for `NAPI_CPP_EXCEPTIONS_ALL` ?

KevinEady:
We discuss in the 18 Oct Node-API meeting:
- Use the `NODE_ADDON_API_` prefix on the defines
- Should the "catch all" be the new default?
  - What about some best practices doc? We also mentioned this with the requiring basic finalizers option

napi-inl.h
```diff
@@ -78,43 +78,61 @@ inline napi_status AttachData(napi_env env,
 // For use in JS to C++ callback wrappers to catch any Napi::Error exceptions
 // and rethrow them as JavaScript exceptions before returning from the callback.
 template <typename Callable>
-inline napi_value WrapCallback(Callable callback) {
-#ifdef NAPI_CPP_EXCEPTIONS
+inline napi_value WrapCallback(napi_env env, Callable callback) {
```
NickNaso, line 81: Why are adding the `env` parameter?

NickNaso (commented):
Is this branch in sync with the `main`?

KevinEady:
The CIs are failing either because (1) unused parameters, or (2) experimental mode basic_env type mismatches.

Regarding 2: Catching all exceptions will construct a new`Napi::Error` to wrap the native exception. This cannot be done on a `basic_env`:

https://github.com/nodejs/node-addon-api/blob/98aae3343c3af36b4befd6b67c4cb19ba49b8d20/napi-inl.h#L209-L227

This `Wrapper` is used when eg. constructing a new `Napi::External` for the finalizer. We can't pass the basic env to `WrapVoidCallback` since it needs a regular env to wrap the error.

This means that a combination of `NAPI_HAS_POST_FINALIZER` and `NODE_ADDON_API_CPP_EXCEPTIONS_ALL` cannot happen, since `Napi::Error`-wrapping cannot occur. 

Should we make it a compile-time `#error` if these two options are used in conjunction?

napi-inl.h
```diff
@@ -112,10 +126,40 @@ inline void WrapVoidCallback(Callable callback) {
...
+  catch (const std::exception& e) {
+    Napi::Error::New(env, e.what()).ThrowAsJavaScriptException();
+  } catch (...) {
+    Napi::Error::New(env, "A native exception was thrown")
+        .ThrowAsJavaScriptException();
+  }
+#endif  // NODE_ADDON_API_CPP_EXCEPTIONS_ALL
+#else
+  // When C++ exceptions are disabled, errors are immediately thrown as JS
+  // exceptions, so there is no need to catch and rethrow them here.
```
legendecas, line 154: I think the term "thrown" should be clarified here. Technically, the JS errors are not "thrown" in C++ natively.

```suggestion
  // When C++ exceptions are disabled, there is no need to catch and rethrow C++ exceptions. JS errors should be thrown with `Error::ThrowAsJavaScriptException`.
```

legendecas (commented):
Looks good to me! A minor suggestion...

legendecas (approved):

mhdawson (approved):
LGTM

Issue #1602, opened 2024-10-25
ObjectWrap leak introduced in 8.2.0


arsnyder16:
When using ObjectWrap that does not override a finalizer a memory leak is occurring. 

I traced this back to #1514

Which introduced the following [code](
https://github.com/nodejs/node-addon-api/blob/ae2c23e7460d91ab5b1844d04372f42801bfa48e/napi-inl.h#L5048) which seems to trigger the leak. Removing this set eliminated the leak, but its not clear to me how this affects a basic or extended finalizer. 

```c++
// Prevent ~ObjectWrap from calling napi_remove_wrap
instance->_ref = nullptr;
```

arsnyder16:
Repro steps 
```sh
CFLAGS=-fsanitize=address CXXFLAGS=-fsanitize=address npm run pretest
LD_PRELOAD=`clang -print-file-name=libclang_rt.asan-x86_64.so` NODE_API_BUILD_CONFIG=Release node --expose-gc test/objectwrap.js
```

you will see the very first leak reported comes from a ObjectWrap<> construction
```
Direct leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x7fdf194265fd in operator new(unsigned long) (/usr/lib/llvm-19/lib/clang/19/lib/linux/libclang_rt.asan-x86_64.so+0x10e5fd) (BuildId: 897c70dd2d121c287b9879aab2db7e64324572a9)
    #1 0x000000c5dd9e in v8impl::ReferenceWithFinalizer::New(napi_env__*, v8::Local<v8::Value>, unsigned int, v8impl::ReferenceOwnership, void (*)(napi_env__*, void*, void*), void*, void*) (/root/.nvm/versions/node/v20.18.0/bin/node+0xc5dd9e) (BuildId: fbe09eeb702823fc4c9cc18fd5d44c32225e2ab6)
    #2 0x000000c6632f in napi_wrap (/root/.nvm/versions/node/v20.18.0/bin/node+0xc6632f) (BuildId: fbe09eeb702823fc4c9cc18fd5d44c32225e2ab6)
    #3 0x7fdeeaaa946c in Napi::cstm::ObjectWrap<Test>::ObjectWrap(Napi::cstm::CallbackInfo const&) (/root/node-addon-api/test/build/Release/binding_custom_namespace.node+0x15646c) (BuildId: 68f9f6c62e70d2753bc17857e6be2f35924283c2)
[...]

PR #1607, opened 2024-10-31
fix: missing napi_delete_reference on ObjectWrap ref

legendecas:
According to https://nodejs.org/api/n-api.html#napi_wrap, the `napi_reference` returned from `napi_wrap` should be deleted with `napi_delete_reference`. In node-addon-api, for an `ObjectWrap`, this is done with `~Reference` destructor when the `_ref` is not `nullptr`.

To prevent `~ObjectWrap` from creating a handle during finalization, add a new flag to guard on it, instead of nullify `_ref`.

Submitted https://github.com/nodejs/node/pull/55620 to allow `napi_delete_reference` been immediately invoked in an `node_api_basic_finalizer`.

Fixes https://github.com/nodejs/node-addon-api/issues/1602

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -3341,6 +3341,8 @@ template <typename T>
 inline Reference<T>::~Reference() {
   if (_ref != nullptr) {
     if (!_suppressDestruct) {
+      // TODO(legendecas): napi_delete_reference should be invoked immediately.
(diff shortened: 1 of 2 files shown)


mhdawson (approved):
LGTM

PR #1620, opened 2024-11-24
chore: enable tokenless Codecov uploads

targos:
Refs: https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token


legendecas:
Looks like codecov is working given the above comment.

Montiwa11:
> Looks like codecov is working given the above comment.

Release

Issue #1651, opened 2025-04-12
Add a normal method to convert PropertyLValue into Value


neko-para:
Currently, a PropertyLValue can only be converted into Value via the type convertion.

```cpp
// cast into var
Napi::Value val = obj["key"];
val.ToString();

// explicit cast
Napi::Value(obj["key"]).ToString();
```

It‘s fairly common to directly use it as a value, but the methods above are quite annoying. The first one needs to name the value, and the second one needs to move the cursor back. Thus I'd suggest adding a helper function behaves like the converter.

> demo usage

```cpp
obj["key"].value().ToString();
```

> possible impl

```cpp
Value value() const { return *this; }
```

mhdawson:
We discussed this in the Node-api team meeting today, it seemed reasonable to us. Would you like to submit a pull request?

mhdawson:
@neko-para maybe in https://github.com/nodejs/node-addon-api/blob/b347dbe35c85296bb983f25b27fa365a347bb41e/test/basic_types/value.cc

PR #1655, opened 2025-04-25
feat: add sugar method for PropertyLValue (#1651)

neko-para:

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1578,6 +1578,11 @@ inline Object::PropertyLValue<Key>& Object::PropertyLValue<Key>::operator=(
 }
 
 template <typename Key>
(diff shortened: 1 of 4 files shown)


mhdawson:
@neko-para thanks for the PR, could you add a test as well?

neko-para:
Actually i'm not sure where should i add the tests. Maybe i shall add an extra entry?

neko-para:
> @neko-para just to confirm its not possible to use the existing operator:
> ```
>     /// Converts an L-value to a value.
>     operator Value() const;
> ```
> 
> instead of adding the AsValue method ?
> 

The cast operator should always work. Adding the method is generally only for convenience, as explicit calling cast operator is not as intuitive as calling method.

mhdawson (approved):
LGTM

Issue #1667, opened 2025-06-13
Add then and catch methods to Promise class


mhdawson:
See discussion in parent issue.

PR #1668, opened 2025-06-14
Add then and catch methods to Promise class

miguelmarcondesf:
Issue: [1667](https://github.com/nodejs/node-addon-api/issues/1667)

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -2813,6 +2813,28 @@ inline void Promise::CheckCast(napi_env env, napi_value value) {
 
 inline Promise::Promise(napi_env env, napi_value value) : Object(env, value) {}
 
+inline MaybeOrValue<Promise> Promise::Then(napi_value onFulfilled) const {
+  EscapableHandleScope scope(_env);
+#ifdef NODE_ADDON_API_ENABLE_MAYBE
+  MaybeOrValue<Value> thenMethodMaybe = Get("then");
+  Function thenMethod = thenMethodMaybe.Unwrap().As<Function>();
+#else
+  Function thenMethod = Get("then").As<Function>();
+#endif
+  MaybeOrValue<Value> result = thenMethod.Call(*this, {onFulfilled});
+#ifdef NODE_ADDON_API_ENABLE_MAYBE
+  if (result.IsJust()) {
+    return Just(scope.Escape(result.Unwrap()).As<Promise>());
+  }
+  return Nothing<Promise>();
+#else
+  if (scope.Env().IsExceptionPending()) {
+    return Promise();
+  }
+  return scope.Escape(result).As<Promise>();
+#endif
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 // Buffer<T> class
 ////////////////////////////////////////////////////////////////////////////////
--- a/napi.h
+++ b/napi.h
@@ -1574,7 +1574,10 @@ class Promise : public Object {
 
   static void CheckCast(napi_env env, napi_value value);
 
+  Promise();
   Promise(napi_env env, napi_value value);
+
+  MaybeOrValue<Promise> Then(napi_value onFulfilled) const;
 };
 
 template <typename T>
--- a/test/promise.cc
+++ b/test/promise.cc
@@ -1,4 +1,5 @@
 #include "napi.h"
+#include "test_helper.h"
 
 using namespace Napi;
 
@@ -23,6 +24,22 @@ Value PromiseReturnsCorrectEnv(const CallbackInfo& info) {
   return Boolean::New(info.Env(), deferred.Env() == info.Env());
 }
 
+Value ThenMethodOnFulfilled(const CallbackInfo& info) {
+  auto deferred = Promise::Deferred::New(info.Env());
+  Function onFulfilled = info[0].As<Function>();
+
+  Promise resultPromise = MaybeUnwrap(deferred.Promise().Then(onFulfilled));
+
+  bool isPromise = resultPromise.IsPromise();
+  deferred.Resolve(Number::New(info.Env(), 42));
+
+  Object result = Object::New(info.Env());
+  result["isPromise"] = Boolean::New(info.Env(), isPromise);
+  result["promise"] = resultPromise;
+
+  return result;
+}
+
 Object InitPromise(Env env) {
   Object exports = Object::New(env);
 
@@ -31,6 +48,7 @@ Object InitPromise(Env env) {
   exports["rejectPromise"] = Function::New(env, RejectPromise);
   exports["promiseReturnsCorrectEnv"] =
       Function::New(env, PromiseReturnsCorrectEnv);
+  exports["ThenMethodOnFulfilled"] = Function::New(env, ThenMethodOnFulfilled);
 
   return exports;
 }
--- a/test/promise.js
+++ b/test/promise.js
@@ -17,4 +17,10 @@ async function test (binding) {
   rejecting.then(common.mustNotCall()).catch(common.mustCall());
 
   assert(binding.promise.promiseReturnsCorrectEnv());
+
+  const onFulfilled = (value) => value * 2;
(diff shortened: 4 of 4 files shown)


napi.h
```diff
@@ -1574,7 +1574,16 @@ class Promise : public Object {
 
   static void CheckCast(napi_env env, napi_value value);
 
+  Promise();
```
legendecas, line 1577: Is this necessary?

napi-inl.h
```diff
@@ -2813,6 +2813,84 @@ inline void Promise::CheckCast(napi_env env, napi_value value) {
...
+  MaybeOrValue<Value> result = thenMethod.As<Function>().Call(*this, {onFulfilled, onRejected});
+  if (result.IsJust()) {
+    return Just(scope.Escape(result.Unwrap()).As<Promise>());
+  }
+  return Nothing<Promise>();
+#else
+  Function thenMethod = Get("then").As<Function>();
+  MaybeOrValue<Value> result = thenMethod.Call(*this, {onFulfilled, onRejected});
+  if (scope.Env().IsExceptionPending()) {
+    return Promise();
```
legendecas, line 2854: To avoid adding a new empty constructor:

```suggestion
  return result;
```

miguelmarcondesf: Hey, if I do that and remove the constructor, it starts raising the missing default constructor for the `Nothing<Promise>()` because the `Maybe<T>` class has a private default constructor for the `_value` member.
Do you think it makes sense to make some changes to the `Maybe` class to handles types without a default constructor?

legendecas: yeah, you are right. Might not worth it.

legendecas (approved):

doc/promises.md
```diff
@@ -75,5 +75,47 @@ Rejects the Promise object held by the `Napi::Promise::Deferred` object.
 
 * `[in] value`: The Node-API primitive value with which to reject the `Napi::Promise`.
 
+## Promise Methods
+
+### Then
```
KevinEady, line 80: We typically create a section for each overload of a function (eg. see `Function` has a bunch of `New`: https://github.com/nodejs/node-addon-api/blob/b594d548d16569ea84827b3ad6c7604b2e0c27d9/doc/function.md)  ... so I think these are missing:

```cpp
Napi::Promise Napi::Promise::Then(const Function& onFulfilled, const Function& onRejected) const;

Napi::Promise Napi::Promise::Then(napi_value onFulfilled) const;
Napi::Promise Napi::Promise::Then(const Function& onFulfilled) const;

Napi::Promise Napi::Promise::Catch(const Function& onRejected) const;
```

The overloads for `Then()` and `Catch()` that are only differing by `napi_value` vs `const Function&` can be grouped together, like we do for `Napi::Symbol::New`: https://github.com/nodejs/node-addon-api/blob/0b70d4de26861c2395ba262fbeba926cbc20f2e0/doc/symbol.md#new

Additionally, the documented return value should be `Napi::Promise` and not `<Promise>`: we do not worry about the `Maybe` unwraps in documentation, eg. all the [`Function::Call` docs](https://github.com/nodejs/node-addon-api/blob/0b70d4de26861c2395ba262fbeba926cbc20f2e0/doc/function.md#call) just return an `Napi::Value`, even though they do [return `Maybe`s](https://github.com/nodejs/node-addon-api/blob/b594d548d16569ea84827b3ad6c7604b2e0c27d9/napi.h#L1525-L1537).

napi-inl.h
```diff
@@ -2811,8 +2811,95 @@ inline void Promise::CheckCast(napi_env env, napi_value value) {
...
 
+inline Promise::Promise() : Object() {}
+
 inline Promise::Promise(napi_env env, napi_value value) : Object(env, value) {}
 
+inline MaybeOrValue<Promise> Promise::Then(napi_value onFulfilled) const {
+  EscapableHandleScope scope(_env);
+#ifdef NODE_ADDON_API_ENABLE_MAYBE
+  Value thenMethod;
+  if (!Get("then").UnwrapTo(&thenMethod)) {
```
KevinEady, line 2822: Open question: using `Get("then")` returns the `then` property from this object. Do we need to concern ourselves with the possibility of the `then` being shadowed, and not being the original `then` from the `Promise`? If so, we should use `Promise.prototype.then` ... similar to how we are suggested (eg. [eslint rule](https://eslint.org/docs/latest/rules/no-prototype-builtins)) to use `Object.prototype.hasOwnProperty.call(foo, bar)` instead of `foo.hasOwnProperty(bar)`.

On the flipside, the Promise spec does specifically state how `then` should behave on a Promise object (even if it's shadowed): https://promisesaplus.com/#the-then-method

I think `Get("then")` is okay, just wanted to bring this up as a point of discussion.

doc/promises.md
```diff
@@ -75,5 +75,47 @@ Rejects the Promise object held by the `Napi::Promise::Deferred` object.
...
+**Returns:** A new `<Promise>` that resolves or rejects based on the handler's result.
+
+**Example:**
+```cpp
+// Single fulfillment handler
+Promise newPromise = existingPromise.Then(fulfillmentHandler);
+
+// Both fulfillment and rejection handlers
+Promise chainedPromise = existingPromise.Then(onFulfilled, onRejected);
+```
```
KevinEady, line 101: We typically put examples in their own section at the end of the document, eg:
- https://github.com/nodejs/node-addon-api/blob/0b70d4de26861c2395ba262fbeba926cbc20f2e0/doc/async_worker.md#example
- https://github.com/nodejs/node-addon-api/blob/0b70d4de26861c2395ba262fbeba926cbc20f2e0/doc/handle_scope.md#example
- https://github.com/nodejs/node-addon-api/blob/0b70d4de26861c2395ba262fbeba926cbc20f2e0/doc/threadsafe_function.md#example

As it stands, I don't think this example really provides that much value? If you end up documenting all the `Then` and `Catch` overloads, then this is just giving us information that is already present in the documentation (ie. that `Then` returns an `Napi::Promise` and takes either one or two arguments). I would say remove this, or add a more concrete example. Up to you 😃

doc/promises.md
```diff
@@ -75,5 +75,47 @@ Rejects the Promise object held by the `Napi::Promise::Deferred` object.
...
+
+**Parameters:**
+* `[in] onRejected`: A function to be called when the promise is rejected
+
+**Returns:** A new `<Promise>` that handles rejection cases.
+
+**Example:**
+```cpp
+Promise handledPromise = existingPromise.Catch(rejectionHandler);
+```
```
KevinEady, line 119: Same as the other example's note: I don't think this provides that much value and can be removed.

KevinEady (changes requested):
Mostly some changes on the docs... but the implementation looks good! Thanks for the contribution 👍

doc/promises.md
miguelmarcondesf: @KevinEady Thank you very much for all the clarifications, I'll do that!

doc/promises.md
miguelmarcondesf: It makes sense, I'll remove those examples

napi-inl.h
miguelmarcondesf: This is a very valid point, given what you brought up, I'm also okay to keep it that way in light of how `then` should behave, even if it was shadowed.

doc/promises.md
```diff
@@ -75,5 +75,56 @@ Rejects the Promise object held by the `Napi::Promise::Deferred` object.
...
+
+### Then
+
+```cpp
+Napi::Promise Napi::Promise::Then(napi_value onFulfilled, napi_value onRejected) const;
+Napi::Promise Napi::Promise::Then(const Function& onFulfilled,
+                                  const Function& onRejected) const;
+```
+
+Attaches fulfillment and/or rejection handlers to the promise and returns a new promise.
```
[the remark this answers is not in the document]
miguelmarcondesf: You are totally right, thanks

mhdawson (approved):
LGTM

PR #1670, opened 2025-06-19
feat: add ConstructorTraitTag to allow constructor return overridding

### LINKED ISSUE #1661: Wrapping Event in a C++ class so that every where you pass an event it works.
in JavaScript I can do this

```js
class MyEvent extends Event {
  constructor(type, msg) {
    super(type);
    this.message = msg;
  }
}
const target = new EventTarget();
const e = await new Promise(resolve => {
  target.addEventListener('custom', resolve);
  target.dispatchEvent(new MyEvent('custom', 'hello'));
});
assert.ok(e.message === 'hello');
```

Is it possible to have `MyEvent` be a C++ class?

In my current attempt, when I call `target.dispatchEvent` I get

```
node:internal/event_target:220
      throw new ERR_INVALID_THIS('Event');
            ^

TypeError [ERR_INVALID_THIS]: Value of "this" must be of type Event
    at get type [as type] (node:internal/event_target:220:13)
    at EventTarget.dispatchEvent (node:internal/event_target:755:40)
```

I tried setting the prototype chain in C++ and this passes

```js
assert.ok(new MyEvent('custom', 'hello') instanceof Event);
```

but of course it's not actually an `Event`, it's a non-event who's prototype chain contains `Event` which is what node is complaining about.

legendecas:
Fixes: https://github.com/nodejs/node-addon-api/issues/1661

- [ ] Add docs.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -4630,9 +4630,11 @@ inline napi_value InstanceWrap<T>::WrappedMethod(
 ////////////////////////////////////////////////////////////////////////////////
 
 template <typename T>
-inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo) {
+template <typename ConstructorTraitTag>
+inline ObjectWrap<T>::ObjectWrap(const Napi::CallbackInfo& callbackInfo,
+                                 ConstructorTraitTag) {
   napi_env env = callbackInfo.Env();
-  napi_value wrapper = callbackInfo.This();
+  napi_value wrapper = ConstructorTraitTag::GetThis(callbackInfo);
   napi_status status;
   napi_ref ref;
   T* instance = static_cast<T*>(this);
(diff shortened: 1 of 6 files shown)


greggman:
Thank you for working on this!

It's not clear from the comments if this is working yet. It sounds like maybe not?

I would suggest adding a few more tests

* Make a C++ class that inherits from `Event`, then pass it to a native event target

  ```js
  const myEvent = new CPPClassThatInheritsFromEvent('nameOfEvent');
  const target = new EventTarget();
  const e = await new Promise(resolve => {
    target.addEventListener('nameOfEvent', resolve, { once: true });
    target.dispatchEvent(myEvent);
  });
  assert(e === myEvent);
  ```

* Make a C++ class that inherits from `EventTarget`. Maybe sure you can use EventTarget's functionality

  ```js
  const target = new CPPClassThatInheritsFromEventTarget();
  const event = new Event('foo');
  const e = await new Promise(resolve => {
    target.addEventListener('foo', resolve, { once: true });
    target.dispatchEvent(event);
  });
  assert(e === event);
  ```

KevinEady:
> This does not work for `InstanceMethod` because there is a branch check enabled at https://github.com/nodejs/node/blob/main/src/js_native_api_v8.cc#L1028.

Hi @legendecas , is there anything we can do to address this branch check? Is it absolutely needed?

PR #1688, opened 2025-10-28
feat: add SharedArrayBuffer

KevinEady:
- Add support for `Napi::SharedArrayBuffer`
- Add `Value::IsSharedArrayBuffer()`
- Update docs and tests

### DIFF
--- a/test/value_type_cast.cc
+++ b/test/value_type_cast.cc
@@ -27,6 +27,11 @@ namespace {
 #define V(Type)                                                                \
   void TypeCast##Type(const CallbackInfo& info) { USE(info[0].As<Type>()); }
 TYPE_CAST_TYPES(V)
+
+#ifdef NODE_API_EXPERIMENTAL_HAS_SHAREDARRAYBUFFER
+V(SharedArrayBuffer)
+#endif
+
 #undef V
 
 void TypeCastBuffer(const CallbackInfo& info) {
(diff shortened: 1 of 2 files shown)


napi.h
```diff
@@ -1202,8 +1205,30 @@ class Object::iterator {
 };
 #endif  // NODE_ADDON_API_CPP_EXCEPTIONS
 
+class ArrayBufferLike : public Object {
```
legendecas: Usually we use the name `XxxLike` when it describes a shape of a protocol. For ArrayBuffer and SharedArrayBuffer, this is not the case. The `Napi::ArrayBuffer` and `Napi::SharedArrayBuffer` do not work with a JavaScript mocked ArrayBuffer-like objects.

I think a better name can be `ArrayBufferBase`. Or just duplicate these two methods as in JavaScript, `ArrayBuffer` and `SharedArrayBuffer` have no common ancestry.

KevinEady: Addressed in https://github.com/nodejs/node-addon-api/pull/1688/commits/f861d36e5aba8b3d84f78f08a3399fea1c84128e

KevinEady: I went with the approach of:

> Or just duplicate these two methods as in JavaScript, ArrayBuffer and SharedArrayBuffer have no common ancestry.

doc/shared_array_buffer.md
```diff
@@ -0,0 +1,65 @@
+# SharedArrayBuffer
```
NickNaso: You should add this as link to the `README` under section

API Documentation -> Data Types

legendecas (commented):
Would you mind adding a test case in https://github.com/nodejs/node-addon-api/blob/main/test/value_type_cast.cc as well? Or we can add it as a follow-up.

KevinEady:
@legendecas, I added the value_type_cast test! PTAL

PR #1692, opened 2025-11-14
Silence a legitimate sanitizer warning that's enabled by default in Android NDK 29

matthargett:
Silence a legitimate vfptr sanitizer warning that is on by default in Android NDK 29's clang. Stripping RTTI from a vfptr is usually a no-no for debugging and field-supportability reasons, but it's intrinsic to how ObjectWrap works right now.

This was found whilst porting BabylonNative (which implements NAPI for add-ons) to Android XR, and was integrated (and tested) there as a local patch:
https://github.com/BabylonJS/JsRuntimeHost/pull/118/files#diff-a9d53b78150c809d9bf63a90260b4f823918bc6e6d9c3e02eb68d855bf9deda1R4499

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -21,6 +21,12 @@
 #include <type_traits>
 #include <utility>
 
+#if defined(__clang__) || defined(__GNUC__)
+#define NAPI_NO_SANITIZE_VPTR __attribute__((no_sanitize("vptr")))
+#else
+#define NAPI_NO_SANITIZE_VPTR
+#endif
+
 namespace Napi {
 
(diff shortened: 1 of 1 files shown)


napi-inl.h
```diff
@@ -21,6 +21,12 @@
 #include <type_traits>
 #include <utility>
 
+#if defined(__clang__) || defined(__GNUC__)
+#define NAPI_NO_SANITIZE_VPTR __attribute__((no_sanitize("vptr")))
```
KevinEady: I believe the macro def should be `NODE_ADDON_API_NO_SANITIZE_VPTR`

The `NAPI_` / `NODE_API_` prefix is meant for core.

PR #1706, opened 2026-01-30
Add String::New overload for string_view

### LINKED ISSUE #1703: Add `String::New` for `std::string_view`
It would be nice to have a string "constructor" which takes a `string_view`

Could be as simple as:

```cpp
Napi::String::New(napi_env env, std::string_view value)
{
    return Napi::String::New(env, value.data(), value.size());
}
```

cacharle:
Fixes #1703


KevinEady (changes requested):
Guards for c++17 can be removed.

cacharle:
Should be good now

Thanks for finishing #1704 for me btw, life got in the way

PR #1735, opened 2026-06-05
feat: add Object::New overload utilizing node_api_create_object_with_properties

### LINKED ISSUE #1689: Porting create_object_with_properties
We have merged [node-api: Add node_api_create_object_with_properties method](https://github.com/nodejs/node/pull/59953) and now it would be good to have it here as well

I'd like to work on that too, so I'll open a PR as soon as possible.

KevinEady:
Add `Object::New` overload utilizing `node_api_create_object_with_properties`

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1635,6 +1635,31 @@ inline Object Object::New(napi_env env) {
   return Object(env, value);
 }
 
+#ifdef NODE_API_EXPERIMENTAL_HAS_CREATE_OBJECT_WITH_PROPERTIES
+inline Object Object::New(napi_env env,
+                          napi_value prototypeOrNull,
+                          std::vector<napi_value>& propertyNames,
+                          std::vector<napi_value>& propertyValues) {
+  if (propertyNames.size() != propertyValues.size()) {
+    Napi::Error::New(env, "Mismatch in size of property names and values")
+        .ThrowAsJavaScriptException();
+    return Object();
+  }
+
+  napi_value value;
+  napi_status status =
+      node_api_create_object_with_properties(env,
+                                             prototypeOrNull,
+                                             propertyNames.data(),
+                                             propertyValues.data(),
+                                             propertyNames.size(),
+                                             &value);
+
+  NAPI_THROW_IF_FAILED(env, status, Object());
+  return Object(env, value);
+}
+#endif
+
 inline void Object::CheckCast(napi_env env, napi_value value) {
   NAPI_CHECK(value != nullptr, "Object::CheckCast", "empty value");
 
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1635,6 +1635,31 @@ inline Object Object::New(napi_env env) {
   return Object(env, value);
 }
 
+#ifdef NODE_API_EXPERIMENTAL_HAS_CREATE_OBJECT_WITH_PROPERTIES
+inline Object Object::New(napi_env env,
+                          napi_value prototypeOrNull,
+                          std::vector<napi_value>& propertyNames,
+                          std::vector<napi_value>& propertyValues) {
+  if (propertyNames.size() != propertyValues.size()) {
+    NAPI_THROW(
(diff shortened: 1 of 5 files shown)


legendecas:
The CI failure is https://github.com/nodejs/node-gyp/issues/3327

napi-inl.h
```diff
@@ -1635,6 +1635,31 @@ inline Object Object::New(napi_env env) {
   return Object(env, value);
 }
 
+#ifdef NODE_API_EXPERIMENTAL_HAS_CREATE_OBJECT_WITH_PROPERTIES
+inline Object Object::New(napi_env env,
+                          napi_value prototypeOrNull,
+                          std::vector<napi_value>& propertyNames,
+                          std::vector<napi_value>& propertyValues) {
+  if (propertyNames.size() != propertyValues.size()) {
+    Napi::Error::New(env, "Mismatch in size of property names and values")
```
legendecas: This should prefer `NAPI_THROW` to throw as a c++ exception when it is enabled:

```suggestion
  NAPI_THROW(
      Napi::Error::New(env, "Mismatch in size of property names and values"),
      Object());
```

napi.h
```diff
@@ -903,6 +903,16 @@ class Object : public TypeTaggable {
   static Object New(napi_env env  ///< Node-API environment
   );
 
+#ifdef NODE_API_EXPERIMENTAL_HAS_CREATE_OBJECT_WITH_PROPERTIES
+  /// Creates a new Object value with the given property names and values.
+  static Object New(
+      napi_env env,                ///< Node-API environment
+      napi_value prototypeOrNull,  ///< Prototype (Object) or null / empty Value
+      std::vector<napi_value>& propertyNames,  ///< Property names
+      std::vector<napi_value>& propertyValues  ///< Property values
```
legendecas, line 912: This could take const vectors:

```suggestion
      const std::vector<napi_value>& propertyNames,  ///< Property names
      const std::vector<napi_value>& propertyValues  ///< Property values
```

KevinEady: Hi @legendecas ,

Thanks for your patience :)

This cannot be `const` because using the `.data()` would then return `const napi_value*`, but `node_api_create_object_with_properties` does not accept `const`-qualified pointers: 

```
../../../napi-inl.h:1651:7: error: no matching function for call to 'node_api_create_object_with_properties'
      node_api_create_object_with_properties(env,
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/kevineady/Library/Caches/node-gyp/24.14.0/include/node/js_native_api.h:56:1: note: candidate function not viable: 3rd argument ('const value_type *' (aka 'napi_value__ *const *')) would lose const qualifier
node_api_create_object_with_properties(napi_env env,
```

```cpp
node_api_create_object_with_properties(napi_env env,
                                       napi_value prototype_or_null,
                                       napi_value* property_names,
                                       napi_value* property_values,
                                       size_t property_count,
                                       napi_value* result);
```

Might have been an oversight of the Node-API for `node_api_create_object_with_properties`...?

Eg. `Function::New` can accept an `const std::vector<napi_value>& args`, but that is because the underlying Node-API `napi_new_instance` accepts a `const napi_value* argv`.

napi-inl.h
KevinEady: Addressed in push from `23b62a9` to `be1b480` ([diff](https://github.com/nodejs/node-addon-api/compare/23b62a942a71afcca3182b8d93c906eecd041e46..be1b480799b777862e63ae6456f7593d058b27fe))

napi.h
legendecas: You are right, I think this can be a follow-up. `node_api_create_object_with_properties` is actually read-only on `property_names` and `property_values`, so it should be safe to add the `const` qualifiers.

In the interim, I think we could also use `const_cast` because this does not write into the pointee so this does not incur any runtime behavior change. But we could do it after updating `node_api_create_object_with_properties`.

legendecas (approved):

PR #1738, opened 2026-07-13
doc: add inline comment hints on finalizer signatures

legendecas:
Add inline comment hints on finalizer signatures.

When C++20 is the baseline, we should add C++ concepts to hint about the templated parameters.

### DIFF
--- a/napi.h
+++ b/napi.h
@@ -422,12 +422,16 @@ class BasicEnv {
 #endif  // NAPI_VERSION > 8
 
 #ifdef NODE_API_EXPERIMENTAL_HAS_POST_FINALIZER
+  // FinalizerType must implement `void operator()(Env env)`.
   template <typename FinalizerType>
   inline void PostFinalizer(FinalizerType finalizeCallback) const;
 
+  // FinalizerType must implement `void operator()(Env env, T* data)`.
   template <typename FinalizerType, typename T>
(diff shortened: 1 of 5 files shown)


napi.h
```diff
@@ -422,12 +422,16 @@ class BasicEnv {
 #endif  // NAPI_VERSION > 8
 
 #ifdef NODE_API_EXPERIMENTAL_HAS_POST_FINALIZER
+  // FinalizerType must implement `void operator()(Env env)`.
```
alexweej, line 425: Have we considered using C++ Concepts for this instead of these human-only annotations?

alexweej: (Though clearly this doesn't need to block the PR...)

legendecas: it will be a semver-major to bump minimum c++ version to c++20. Right now node-addon-api still supports c++17 as in node v18/v20 https://github.com/nodejs/node-addon-api/blob/13c28f6673776e4015cd8d675e92b143177bb816/package.json#L478

KevinEady, line 425: Should the Markdown documentation files also be updated to reflect the implementation signature for the callbacks? 🤔

legendecas: @KevinEady updated md docs

KevinEady (approved):

PR #1742, opened 2026-08-07
fix: resolve Symbol::For overload ambiguity

### LINKED ISSUE #1741: Ambiguity on `Symbol::For` overload between `std::string` and `std::string_view` convertible
https://github.com/nodejs/node-addon-api/pull/1722 introduced overloads on `Symbol::For` with `std::string_view`, but this also introduced an ambiguity when a type is convertible to both `std::string` and `std::string_view`.

An alternative path forward is that we should remove `std::string` overload when introducing `std::string_view` support when applicable.

umuoy1:
`Symbol::For` an ambiguous for types that can convert to both `std::string` and `std::string_view`.

Instead of removing the existing `std::string` overload, this keep it so string-only wrappers continue to work and adds a constrained forwarding overload for the ambiguous case. Calls that already have a unique best match keep using the existing overloads.

Fixes #1741

### DIFF
--- a/napi.h
+++ b/napi.h
@@ -800,10 +800,10 @@ struct string_convertible_probe {
 };
 
 template <typename T, typename = void>
-struct has_unambiguous_symbol_for_overload : std::false_type {};
+struct has_unambiguous_string_convertible_overload : std::false_type {};
 
 template <typename T>
-struct has_unambiguous_symbol_for_overload<
+struct has_unambiguous_string_convertible_overload<
     T,
     std::void_t<decltype(string_convertible_probe::select(std::declval<T>()))>>
     : std::true_type {};
@@ -814,11 +814,11 @@ struct has_unambiguous_symbol_for_overload<
 // Exclude nullptr because it matches the pointer overloads equally well and
 // cannot safely initialize a std::string_view.
 template <typename T>
-using enable_if_ambiguous_symbol_for_t =
+using enable_if_ambiguous_string_convertible_t =
     std::enable_if_t<!std::is_null_pointer_v<std::decay_t<T>> &&
                          std::is_convertible_v<T, const std::string&> &&
                          std::is_convertible_v<T, std::string_view> &&
-                         !has_unambiguous_symbol_for_overload<T>::value,
+                         !has_unambiguous_string_convertible_overload<T>::value,
                      int>;
 
 }  // namespace details
@@ -870,7 +870,8 @@ class Symbol : public Name {
 
   // Resolve otherwise ambiguous string-like arguments through the
   // std::string_view overload
-  template <typename T, details::enable_if_ambiguous_symbol_for_t<T&&> = 0>
+  template <typename T,
+            details::enable_if_ambiguous_string_convertible_t<T&&> = 0>
   static MaybeOrValue<Symbol> For(napi_env env, T&& description);
 
   // Create a symbol in the global registry, C style string (null terminated)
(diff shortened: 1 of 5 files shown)


test/symbol.cc
```diff
@@ -5,6 +5,21 @@
...
+class StringLike {
+ public:
+  explicit StringLike(const std::string& value) : _value(value) {}
+
+  operator std::string() const { return _value; }
+  operator std::string_view() const { return _value; }
+
+ private:
+  std::string _value;
+};
```
legendecas, line 56: Can we also add a variant testing a type that only supports converting to `std::string`?

```c++
struct StringLike {
  operator std::string() const;
};
```

If this fails, we might need to add a type constraint like for string-like types:

```c++
template <
    typename T,
    std::enable_if_t<
        std::is_convertible_v<T, std::string> &&
        std::is_convertible_v<T, std::string_view> &&
        !std::is_same_v<std::decay_t<T>, std::string> &&
        !std::is_same_v<std::decay_t<T>, std::string_view>,
        int> = 0>
static MaybeOrValue<Symbol> For(napi_env env, T&& description) {
  return For(env, std::string_view(std::forward<T>(description)));
}
```

umuoy1: Good point, thanks! I'll add a variant that only converts to `std::string` and update the overload handling accordingly.

napi.h
```diff
@@ -786,6 +788,41 @@ class String : public Name {
       const;  ///< Converts a String value to a UTF-16 encoded C++ string.
 };
 
+namespace details {
+
+// This overload set must mirror the non-template Symbol::For overloads.
+struct symbol_for_overload_probe {
```
legendecas: This could be a general helper for APIs taking a string/string_view, like in `Napi::String::New` (it can be a follow-up, not in this particular PR):

```suggestion
struct string_convertible_probe {
```

umuoy1: Added , thanks

napi.h
```diff
@@ -786,6 +788,41 @@ class String : public Name {
...
+struct string_convertible_probe {
+  static void select(const std::string&);
+  static void select(std::string_view);
+  static void select(const char*);
+  static void select(String);
+  static void select(napi_value);
+};
+
+template <typename T, typename = void>
+struct has_unambiguous_symbol_for_overload : std::false_type {};
```
legendecas:
```suggestion
struct has_unambiguous_string_convertible_overload : std::false_type {};
```

napi.h
```diff
@@ -786,6 +788,41 @@ class String : public Name {
...
+
+template <typename T>
+struct has_unambiguous_symbol_for_overload<
```
legendecas:
```suggestion
struct has_unambiguous_string_convertible_overload<
```

napi.h
```diff
@@ -786,6 +788,41 @@ class String : public Name {
...
+    std::void_t<decltype(string_convertible_probe::select(std::declval<T>()))>>
+    : std::true_type {};
+
+// Enable the template overload only for string-like arguments that have no
+// unique best match among the non-template Symbol::For overloads.
+//
+// Exclude nullptr because it matches the pointer overloads equally well and
+// cannot safely initialize a std::string_view.
+template <typename T>
+using enable_if_ambiguous_symbol_for_t =
```
legendecas:
```suggestion
using enable_if_ambiguous_string_convertible_t =
```

legendecas (approved):

PR #1755, opened 2026-08-28
fix: allow const vectors with Object::New overload utilizing node_api_create_object_with_properties

KevinEady:
Use `const_cast` to remove `const`-qualifier from the pointer returned by the `vector`s' `data()` method.

### DIFF
--- a/napi-inl.h
+++ b/napi-inl.h
@@ -1644,8 +1644,8 @@ inline Object Object::New(napi_env env) {
 #ifdef NODE_API_EXPERIMENTAL_HAS_CREATE_OBJECT_WITH_PROPERTIES
 inline Object Object::New(napi_env env,
                           napi_value prototypeOrNull,
-                          std::vector<napi_value>& propertyNames,
-                          std::vector<napi_value>& propertyValues) {
+                          const std::vector<napi_value>& propertyNames,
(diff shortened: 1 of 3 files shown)


KevinEady:
Tested locally against https://github.com/nodejs/node/pull/65621

napi-inl.h
```diff
@@ -1644,22 +1644,22 @@ inline Object Object::New(napi_env env) {
...
 
   napi_value value;
-  napi_status status =
-      node_api_create_object_with_properties(env,
-                                             prototypeOrNull,
-                                             propertyNames.data(),
-                                             propertyValues.data(),
-                                             propertyNames.size(),
-                                             &value);
+  napi_status status = node_api_create_object_with_properties(
```
legendecas, line 1658: Could we add a TODO to remove the const_cast when https://github.com/nodejs/node/pull/65621 is fully backported?

KevinEady: Addressed in push from cefce1a to 12be15f ([diff](https://github.com/nodejs/node-addon-api/compare/cefce1adc526c84935aec19dffca702a4f88905d..12be15fb169136209043e044c3699a47016dacb2))

legendecas (approved):
Thanks! Let's wait a bit until https://github.com/nodejs/node/pull/65621 lands.

